PR #59, opened 2014-12-23
Few minor fixes

DanSimon:
* added a convenient `withHeader` method to `HttpRequest`
* added some new constructors to `AsyncServiceClient` that mirror those when creating clients in the DSL
* added very basic debugging feature to stream parsers that when enabled, will copy all of the data they parse and output raw data for each parsed item.  Obviously defaults to off.  This should probably be refined a bit but it already has proven to be pretty useful.


DanSimon:
I pulled it out because it was too hacky.  There's other better ways to accomplish it.

PR #62, opened 2015-01-05
Streaming buffers

DanSimon:
Initial merge of all the fancy new streaming stuff:

* New Controller Layer separates the semantics of encoding/decoding messages from the request/response cycle of services.  This opens the door for streams, multiplexed services, and many other situations that don't fit into the relatively strict service paradigm.
* New Pipes, Sources, and Sinks for creating streams of data.


nsauro:
Big PR..I've been working off of this branch for the http streaming stuff, so I have already implicitly approved :)

PR #63, opened 2015-01-06
adding support for http chunked transfer

### LINKED ISSUE #58: Http chunked transfer encoding responses not supported
Right now we don't support it at all and the only way to avoid it is to use http 1.0.

We don't need streaming for this, just modify the parser to handle the chunks.

DanSimon:
Fixes #58 

This adds (non-streaming) chunked transfer encoding support both for requests and responses.  So while this won't work for infinite streams (which is intentional), it will work for ordinary responses that happen to be chunk encoded.

There's still some quirks dealing with content-length, but they're unrelated to this fix so it'll be addressed in another PR

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
@@ -16,11 +16,13 @@ object HttpRequestParser {
   def apply(size: DataSize = DefaultMaxSize) = maxSize(size, httpRequest)
 
   protected def httpRequest: Parser[HttpRequest] = httpHead |> {head => 
-    if (head.contentLength > 0) {
-      bytes(head.contentLength) >> {body => HttpRequest(head, Some(body))}
-    } else {
-      const(HttpRequest(head, None))
-    }
+    head.singleHeader(HttpHeaders.TransferEncoding) match { 
+      case None | Some("identity") => head.contentLength match {
+        case Some(0) | None => const(HttpRequest(head, None))
+        case Some(n) => bytes(n) >> {body => HttpRequest(head, Some(body))}
+      }
+      case Some(other)  => chunkedBody >> {body => HttpRequest(head, Some(body))}
+    } 
   }
 
   protected def httpHead = firstline ~ headers >> {case method ~ path ~ version ~ headers => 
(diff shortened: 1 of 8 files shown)


colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
```diff
@@ -16,11 +16,13 @@ object HttpRequestParser {
   def apply(size: DataSize = DefaultMaxSize) = maxSize(size, httpRequest)
 
   protected def httpRequest: Parser[HttpRequest] = httpHead |> {head => 
-    if (head.contentLength > 0) {
-      bytes(head.contentLength) >> {body => HttpRequest(head, Some(body))}
-    } else {
-      const(HttpRequest(head, None))
-    }
+    head.singleHeader(HttpHeaders.TransferEncoding) match { 
+      case None | Some("identity") => head.contentLength match {
```
nsauro: Not that I have a strong feeling about it, but do we want to handle a case where both content length and transfer encoding are present in the header, and possibly emit a failure?

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
```diff
@@ -19,12 +19,15 @@ object HttpResponseParser {
   def apply(size: DataSize = DefaultMaxSize) = maxSize(size, response)
 
   def response: Parser[HttpResponse] = firstLine ~ headers |> {case (version, code) ~ headers =>
-    val clength = headers.collectFirst{case (c,v) if (c == HttpHeaders.ContentLength) => v.toInt}.getOrElse(0)
-    if (clength > 0) {
-      bytes(clength) >> {body => HttpResponse(version, code, body, headers)}
-    } else {
-      const(HttpResponse(version, code, ByteString(), headers))
-    }
+    val clength = headers.collectFirst{case (c,v) if (c == HttpHeaders.ContentLength) => v.toInt}
```
nsauro: can't we just use :  headers.collectFirst{case (HttpHeaders.ContentLength,v) => v.toInt}

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
```diff
@@ -19,12 +19,15 @@ object HttpResponseParser {
...
+    val encoding = headers.collectFirst{case (c,v) if (c == HttpHeaders.TransferEncoding) => v.toLowerCase}
```
nsauro: Same remark as with the request parser with regards to content length and chunk encoded transfer headers both being present.

nsauro:
other than my comments which are pretty small, looks great.

colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
DanSimon: I checked the [http spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4) and it says to ignore content length in that situation.

PR #64, opened 2015-01-07
Removing Response and Completion types

DanSimon:
Both the `Response` and `Completion` types are removed.  This should dramatically cleanup service code, especially in places where business logic is split across multiple classes and all of the response wrapping got in the way.

* Services must now return `Callback[Output]` and implicit conversions must be manually pulled in.  There is also no built-in implicit from converting from `Future[T]` to `Callback[T]`, which should help eliminate confusion that arises about when code is running inside a future vs a callback.
* A new `TagDecorator[I,O]` typeclass replaces the `withTags` functionality of `Completion`.  Basically this can be used to attach tags to the built-in metrics of services for each request, for example adding the http status code.

Some minor things are currently broken, but I figured this PR was getting large enough so I will address the in separate issues:
* There's no replacement for the `PostWriteAction` that was part of `Completion`.  This was used, for example, by http responses to automatically close the connection when using Http 1.0.  The replacement will be a `postWrite` method along with a `gracefulDisconnect` for services.
* Some basic TagDecorators should be created for existing codecs, especially http.


DanSimon:
rebasing not working, gonna squash merge into another branch and re-open

PR #65, opened 2015-01-08
Http streaming

nsauro:
preview of http streaming..rebasing right now.


nsauro:
This version is ready for review

colossus-tests/src/test/scala/colossus/controller/PipeSpec.scala
```diff
@@ -84,7 +84,7 @@ class PipeSpec extends WordSpec with MustMatchers {
       pipe.push(1) must equal(Success(PushResult.Ok))
       val r = pipe.push(1) 
       var triggered = false
-      r mustBe an[Success[PushResult.Full]]
+      r mustBe an[Success[_]]
```
DanSimon: These tests needs to be further fixed, simply checking if its a Success is not enough

colossus-tests/src/test/scala/colossus/controller/InputControllerSpec.scala
```diff
@@ -17,7 +14,9 @@ class InputControllerSpec extends ColossusSpec {
   "Input Controller" must {
     "decode a stream message" in {
       val expected = ByteString("Hello world!")
+      println(s"size: ${expected.size.toString}")
```
DanSimon: printlns

DanSimon:
Looks good for now, lot of open questions though.  It would be awesome if we could figure out a way not to have two different codecs, but at least right now I don't see it.

PR #75, opened 2015-01-12
Wrap Redis key and value with ByteString

fehmicansaglam:
Examples were not compiling but I am not sure this is the best way to fix it.


DanSimon:
Looks good, thanks for catching that.

PR #78, opened 2015-01-12
first round of streaming in http chunked encodings

nsauro:
just getting this ready, still needs some work

### DIFF
--- a/colossus-tests/src/test/scala/colossus/controller/PipeCombinatorSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/controller/PipeCombinatorSpec.scala
@@ -1,0 +1,86 @@
+package colossus.controller
+
+import org.scalatest.{TryValues, OptionValues, MustMatchers, WordSpec}
+
+import scala.util.{Failure, Success}
+
+class PipeCombinatorSpec extends WordSpec with MustMatchers with OptionValues with TryValues{
+
+  "PipeCombinator" must {
+
+    "combine 2 pipes with a map function" in {
+
+      import ReflexiveCopiers._
+
+      val writeSink: Pipe[String, String] = new InfinitePipe[String](100)
+      val readSource: Pipe[Int, Int] = new InfinitePipe[Int](100)
+
+      val map: (String) => Seq[Int] = (s: String) => s.split(":").map(_.toInt).toSeq
+
+      val combined: Pipe[String, Int] = writeSink.join(readSource)(map)
+      combined.push("1:2:3")
+      combined.push("4:5")
+      combined.complete()
+
+      var callbackExecuted = false
+
+      combined.fold(Seq[Int]())((el, ac) => ac :+ el).execute {
+        case Success(x) => callbackExecuted = true; x must equal(Seq(1, 2, 3, 4, 5))
+        case Failure(t) => throw t
+      }
+
+      callbackExecuted must be (true)
+    }
+
+    "completing a pipe will complete both sides" in {
+      import ReflexiveCopiers._
+
+      val writeSink: Pipe[String, String] = new InfinitePipe[String](100)
+      val readSource: Pipe[Int, Int] = new InfinitePipe[Int](100)
+
+      val map: (String) => Seq[Int] = (s: String) => s.split(":").map(_.toInt).toSeq
+
+      val combined: Pipe[String, Int] = writeSink.join(readSource)(map)
+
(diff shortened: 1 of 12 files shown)


colossus-tests/src/test/scala/colossus/controller/PipeCombinatorSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+
+      val map: (String) => Seq[Int] = (s: String) => s.split(":").map(_.toInt).toSeq
+
+      val combined: Pipe[String, Int] = writeSink.join(readSource)(map)
+      combined.push("1:2:3")
+      combined.push("4:5")
+      combined.complete()
+
+      combined.fold(Seq[Int]())((el, ac) => ac :+ el).execute {
+        case Success(x) => x must equal(Seq(1, 2, 3, 4, 5))
```
DanSimon: It's possible this test is still not actually working, you should create a variable outside the callback and set its value inside the execute closure, then check the value after calling execute.  Right now if the execute closure throws an exception, it may actually be getting suppressed (like what would happen if you did it inside a future)

colossus-tests/src/test/scala/colossus/protocols/http/client/StreamingHttpResponseParserSpec.scala
```diff
@@ -35,15 +35,63 @@ class StreamingHttpResponseParserSpec extends WordSpec with MustMatchers with Tr
...
+
+      val body = "3\r\nfoo\r\ne\r\n123456789abcde\r\n0\r\n\r\n"
+
+      val p = HttpClientCodec.streaming()
+      val parsed: DecodedResult[StreamingHttpResponse] = p.decode(DataBuffer(ByteString(res))).get
+
+      parsed match {
+        case DecodedResult.Streamed(x, s) => {
+          s.push(DataBuffer(ByteString(body)))
+          x.stream.pullCB().execute(x => ByteString(x.success.value.value.takeAll) must equal (ByteString("foo123456789abcde")))
```
DanSimon: same thing here, better to do the assertion outside the callback execution

colossus/src/main/scala/colossus/controller/Pipe.scala
```diff
@@ -22,10 +22,27 @@ object Copier {
...
 }
 
+//hrm....
+object ReflexiveCopiers{
+
+  trait Reflect[T] {
+    def copy(t : T) : T = t
+  }
+
+  implicit object StringReflect extends Copier[String] with Reflect[String]
```
DanSimon: I'm pretty sure you can make a fully generic identity copier by having a type parameterized method and putting a default value on the implicit parameter that calls it.  I did that with `TagDecorator` and it seems to work well

colossus/src/main/scala/colossus/controller/PipeCombinator.scala
```diff
@@ -0,0 +1,93 @@
...
+        }
+      }
+
+      //can be triggered by either a source or sink, this will immediately cause
+      override def terminate(reason: Throwable) {
+        terminatePipe(input, reason)
+        terminatePipe(output, reason)
+      }
+
+      private def terminatePipe(p : Pipe[_, _], t : Throwable) {
```
DanSimon: probably just add type params to the method instead of having wildcard parameters.

colossus/src/main/scala/colossus/protocols/http/HttpClientCodec.scala
```diff
@@ -32,13 +32,33 @@ object HttpClientCodec {
...
+      }
+
+      private def createChunkedReponse(headers : StreamingHttpResponseHeaderStub) : ResponseAndSink = {
+        val combinator = StreamingHttpChunkPipe(new InfinitePipe[DataBuffer](streamingQueueSize), new InfinitePipe[DataBuffer](streamingQueueSize)) //uergh...this terrifies me
+        val res = StreamingHttpResponse(headers.version, headers.code, headers.headers, combinator)
+        ResponseAndSink(res, combinator)
+      }
+
+      private def createNonChunkedReponse(headers : StreamingHttpResponseHeaderStub) : ResponseAndSink = {
+        val sink = new InfinitePipe[DataBuffer](streamingQueueSize)
```
DanSimon: Think we already talked about this, but it should be a FiniteBytePipe

Issue #84, opened 2015-01-16
CallBack matcher for tests


nsauro:
Would be nice to have a Callback Scalatest matcher, to make testing Callback's a bit easier.  I have begun working on an initial version with the following specs:

class CallbackEvaluateTo[T](f : T => Any) extends Matcher[Callback[T]]

Failure cases are if the callback never gets executed, or if the 'f' function, or the callback itself throw

passing cases are if the callback successfullly executes, and no exceptions are recorded.  Any exceptions are caught, and the messages bubbled up via the MatchResult.

Issue #81, opened 2015-01-16
FiniteBytePipe of length 0 doesn't work


nsauro:
It never enters the "complete" state, and therefore, if you have a pipe of length 0, it will always return a Success(Done), but it's internal InfiniteBytePipe is never closed, since that only happens when the FiniteBytePipe's counter is incremented.

A quick solution is to special case 0, i.e: in the constructor of FiniteBytePipe, if length is 0, automatically complete the internal pipe.

Another solution is to always check the internal's pipe completion phase when returning the Success(Done) 

Another solution is to introduce a NilBytePipe..meh

Additionally, FiniteBytePipe shouldn't allow negative numbers.

Issue #82, opened 2015-01-16
Pipe's semantic for throwing exception when pushing on a finished pipe


nsauro:
I don't think these are exceptional cases, and I would consider them part of the normal workflow, especially as it is expected that the pipe will be controlled/closed from outside of the controller.

We should just have additional PushResult states for this, instead of exceptions.

Easy to update, but we probably just want to think this through a bit before doing it.

nsauro:
We should also update the behavior of a Controller when receiving a PipeClosedException, as that should close the client, since it will now be in a bad state.

DanSimon:
This was fixed in #125

PR #85, opened 2015-01-16
A few things in this PR:

### LINKED ISSUE #83: StreamedResponseBuilder for HTTP is broken
Oops.  I broke it.  I am also fixing it, and will have a PR out soon.  The issue is that it is inserting an extra "/r/n" between the HTTP Header and the body.  Good times were had by none.

nsauro:
- Fixed a bug with the HttpServerCodec's StreamingHttpResponseDatReaderBuilder(phew)..this was inserting an extra '/r/n' which was breaking all responses
- Tweaked Controller to reset the 'currentSink' when a PipeClosed/Terminated exception is thrown.
- Fixed a bug with FiniteBytePipe, which was not correctly closing its internal pipe when a size of 0 was given(this was killing incoming streaming http responses with no bodies)
- added a required on FixedBytePipe to disallow negative byte sizes
- Codec: removed type alias for 'Decoded', was just making things unreadable
- StreamingHttpResponse has a new convenience function 'fromValue' which allows the creation of a StreamingResponse from any ByteStringLike
- added a Callback Scalatest Matcher for ease of testing Callbacks
- added more tests surrounding the fixes and additions of the above items
- fixed bug with streaming more than one response out

This has a few things in it, but the big one is the streaming http response fix.
This pr fixes #83, #81, #84, #89


colossus-testkit/src/test/scala/colossus/CallbackMatchersSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import scala.util.Try
+
+class CallbackMatchersSpec extends WordSpec with MustMatchers{
+
+  import CallbackMatchers._
+
+  "A CallbackMatcher" should {
+
+    "fail to match if a Callback never executes" in {
```
nsauro: @DanSimon ..I think this is a proper test case for testing when a callback never fires....can you verify?

colossus/src/main/scala/colossus/controller/Controller.scala
```diff
@@ -297,8 +300,11 @@ extends ConnectionHandler {
           }
           currentTrigger = Some(trigger)
         }
-        case Failure(reason) => {
-          //todo: what to do here?
+        //TODO: are the following 2 cases really exceptions..seems like they are part of the normal control flow from a controller's POV
```
nsauro: So..I'm not sure how convinced I am of this code.  If the pipe is closed/terminated, we should def reset our sink, as we don't want to push to it..but..there might be more needed here.  Def something to think through.  We might want to have some kind of pluggable behavior here..where if a pipe is closed/terminated while the controller is processing..maybe the controller can take some different actions: ignore, die, etc.  

As I was writing a test for this behavior..is when I started to doubt how "complete" just resetting the sink is.

colossus-testkit/src/main/scala/colossus-testkit/CallbackMatchers.scala
```diff
@@ -0,0 +1,40 @@
...
+package testkit
+
+import colossus.service.Callback
+import org.scalatest.matchers.{MatchResult, Matcher}
+
+import scala.util.{Failure, Success}
+
+trait CallbackMatchers {
+
+  class CallbackEvaluateTo[T](f : T => Any) extends Matcher[Callback[T]] {
```
nsauro: I plan to make a negative matcher, which will match on Failures.  This will be to assert exceptions, etc.

colossus/src/main/scala/colossus/protocols/http/HttpServerCodec.scala
```diff
@@ -88,7 +88,6 @@ object StreamedResponseBuilder extends HttpResponseDataReaderBuilder {
     builder.sizeHint(100)
     appendHeaderBytes(response, builder)
     builder append NEWLINE
-    builder append NEWLINE
```
nsauro: This was the heartbreaker...broke streamed http..:).  The tests I wrote in the StreamingHttpresonseParserSpec verify that this is actually working now

colossus-tests/src/test/scala/colossus/protocols/http/client/StreamingHttpResponseParserSpec.scala
```diff
@@ -97,6 +99,45 @@ class StreamingHttpResponseParserSpec extends WordSpec with MustMatchers with Tr
...
+      validateStreamingEncoding(ByteString(""))
+    }
+
+    "decode a response that was encoded by colossus with a body" in {
+      validateStreamingEncoding(ByteString("body bytes!"))
+    }
+
   }
 
+  private def validateStreamingEncoding(body : ByteString){
```
nsauro: This test just tests that when we encode a streamed response, we can successfully decode it, which should verify that it is a valid http response.

colossus/src/main/scala/colossus/controller/Controller.scala
DanSimon: I'm pretty sure a PipeTerminated exception needs to close the connection, since it means the controller is streaming data into something, and that thing is dead, so the controller has no way to know what to do with the incoming data.  That exception should only be thrown when the situation is basically "Everybody panic!!"  

This does though add evidence that the PipeClosedException should not really be an exception, and just a normal return value.

colossus-testkit/src/test/scala/colossus/CallbackMatchersSpec.scala
DanSimon: I think that should do it

colossus/src/main/scala/colossus/controller/Controller.scala
nsauro: There is an issue #82 to address those, and that stuff can go into another branch I think.

nsauro:
@DanSimon I will merge this if you approve it.

PR #90, opened 2015-01-23
fixed bug in controller's push function, which would not stream more tha...

### LINKED ISSUE #89: Streaming responses is kinda jacked
Looks like we have an issue with streaming out responses....the issue being that we are not correctly removing the "currentStream" option when we are done streaming out the response.  This is causing every subsequent request to fail to respond, as it will never begin to write the response.

nsauro:
...n one response out

fixes #89


nsauro:
I'm not a smart person.  I'm going to kill this PR, and take the commit and put it into the existing PR #85 which has a buncho other stuff in it.

PR #95, opened 2015-01-28
First usable implementation of Mongo Protocol

fehmicansaglam:
I have included an example for MongoDB backed Redis server implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+.*
+!.gitignore
+
 *.class
 *.log
 *.swp
--- a/colossus-examples/src/main/scala/colossus-examples/Main.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/Main.scala
@@ -39,5 +39,6 @@ object Main extends App {
   //chat server using the controller layer
   val chatServer = ChatExample.start(9005)
 
+  val redisMongoServer = MongoBackedRedisExample.start(9006)
 
 }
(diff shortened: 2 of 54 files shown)


nsauro:
Cool, I will start leafing through this.  It might take me a bit as I have some other work that I have to get done beforehand :).  Dan should be back today or tomorrow, so once he gets back into the swing of things, we should start speeding up a bit.

DanSimon:
So all of this looks pretty well-written, however I think for now it would be better if you published this yourself as a standalone jar.  The reason is we won't be using this anywhere ourselves, and I think at this point it would be too much for us to try to maintain as we continue to change the core framework.  We're probably going to be doing the same with other protocols as well.  We might go back on this and want to bring it into the framework eventually, but I think for now this makes the most sense.

Right now we don't have any stable release of 0.6, but we'll be pushing out a milestone release soon (0.6.0-M1) which you can target for your builds.  You can use Sonatype for publishing, which is what we use.  When you have a jar published, we'll be more than happy to link to the project in our docs.

fehmicansaglam:
I see. Can you release a snapshot from the master branch which is automatically updated via Travis? Thus I can make sure that Mongo library is up to date and release a stable version just after Colossus stable version is released.

DanSimon:
Hmm I'll take a look at getting a snapshot automatically published, it should be doable.  At some point we'll also start publishing nightly builds, which might be a little easier to develop against.

fehmicansaglam:
@DanSimon I have created the project and released a snapshot: https://github.com/fehmicansaglam/colossus-extensions

DanSimon:
Great, I just added a link to the repo in our readme, so I'll be closing this PR.  

We'll have to see how this format works out.  The only suggestion I might have going forward is to publish the jar as "colossus-mongo" or something, and put any other independent extensions in other jars, so users can bring in only what they need.

We're also working on setting up a snapshot jar that will be rebuilt every time there's a commit to master.  I hope to get that working in the next few days.

PR #96, opened 2015-01-28
Adding more verbosity on metrics logging

### LINKED ISSUE #94: add a MetricsSender for both Logging and TSDB
I like metrics, logs and tsdb.  Why cannot I not have all 3?!  The simple fix here is to just make a new sender which does both.   A slightly better way of doing it would be to have a tiny DSL for a decorator.

nsauro:
It is controlled now by akka / slf4j log levels.  At the debug level it will print out detailed info of each metric that is being logged which i found super useful for debugging.

Fixes #94


reeve:
:+1:

PR #97, opened 2015-01-28
Close socket on actor stop

benblack86:
Fix a bug where the socket wasn't being properly closed when actor restarted due to socket exception.


DanSimon:
:+1:

reeve:
Nice catch Ben. :100:

nsauro:
There's no emoji for awesomesauce..but picture one here (     )

PR #99, opened 2015-01-29
Memcached protocol tests for confirming that the generated code matches spec. 

codingjester:
I decided to feel my way around scala by updating and adding some features to the Memcached protocol as I saw it was lacking for some projects I was playing around with.

I get the feeling that I need to work on a proper way to do key filtering (I kinda just one lined it), but it works. I can add more tests if you guys feel that anything is sorely lacking. 

Rip my scala apart, it's not very good :)

-------
Nick updates:

Improved upon what JB started, and added the following to memcached:
- implemented the following commands:
 ADD,APPEND,DELETE,GET,PREPEND,REPLACE,SET,INCR,DECR,TOUCH
- multi get
- 'flags' on writes are now supported for reads and writes
- cas is still not supported yet
- refactored ByteString generating code to use size hints to avoid array resizing

In addition to this, there are now IT tests for redis/memcached are now part of the SBT build.  Updated the travis.yml file to run redis/memcached, and then run all tests using a custom 'test-all' task.

@dsimon @nsauro


colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -10,39 +10,149 @@ import java.util.zip.{Deflater, Inflater}
...
+ * than 30 days in seconds, the server will consider it to be real Unix time value rather than an offset from current time. 
+ * 
+ */
+
 
 object MemcacheCommand {
   val RN = ByteString("\r\n")
+  val SP = ByteString(" ")
+
+  private def FormatCommand(command: ByteString, key: ByteString, data: Option[ByteString] = None, ttl: Option[ByteString] = None, flags: Option[ByteString] = None): ByteString = {
```
DanSimon: methods should be camelCase, so this should be `formatCommand`

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -10,39 +10,149 @@ import java.util.zip.{Deflater, Inflater}
...
+    def bytes(c: Compressor) = FormatCommand(ByteString("delete"), key)
   }
   object Delete {
     def apply(key: String): Delete = Delete(ByteString(key))
   }
+  object Incr {
+    def apply(key: String, value: Integer): Incr = Incr(ByteString(key), ByteString(s"${value}"))
+  }
+  case class Incr(key: ByteString, value: ByteString) extends MemcacheCommand {
+    def bytes(c: Compressor) = ByteString("incr ") ++ key.filter(_ > ' ').slice(0,250) ++ ByteString(" ") ++ value ++ RN
```
DanSimon: is this doing the same thing as the formatCommand method?  In any case, looks like there's some duplicated code in the `bytes` methods for these case classes.

codingjester: So Incr/Decr actually have a slightly different format that everything that uses FormatCommand, I can work on figuring out a way for them to go through the same interface though. Basically, Incr/Decr don't have an extra \r\n before the data you want to increment by.

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -10,39 +10,149 @@ import java.util.zip.{Deflater, Inflater}
...
+    def apply(key: String, value: String, ttl: Integer = 0): Set = Set(ByteString(key), ByteString(value), ttl)
+  }
+  object Add {
+    def apply(key: String, value: String, ttl: Integer = 0): Add = Add(ByteString(key), ByteString(value), ttl)
+  }
+  case class Add(key: ByteString,  value: ByteString, ttl: Integer) extends MemcacheCommand {
+  
+    def bytes(compressor: Compressor) = {
+      val data = compressor.compress(value)
+      FormatCommand(ByteString("add"), key, Some(value), Some(ByteString(s"${ttl}")), Some(ByteString(s"0")))
```
DanSimon: so it looks to me like every time you call FormatCommand, you're converting everything into ByteStrings.  It might make more sense to have the method simply take strings/integers as params and do the conversions itself.  Also it might be a good idea to put all the strings into constants inside an object.  Neither of these are very important here since all the code is self-contained, but they're things we try to do in general.

nsauro: The one thing I've noticed with all these classes is that they all have 2 constructors, one constructor using Strings, Ints,  and other "human readable" types, and another using ByteStrings. 

This leaves us with a pattern of :

case class Command(k : ByteString, v : Bytestring, t : ByteString)

and then a corresponding object for more "human readable parameters" :
object  Command {

def apply(k : String, v : String, t : Int) = Command(ByteString(k).......)
}

The tests seem to indicate(maybe?) that the users of this API will always speak in terms of Strings, Ints, etc..not necessarily in ByteStrings.  The ByteStrings themselves seemed to only be used internally, in order to generate the DataBuffers

My suggestion would be to clearly separate these concerns out, and only have the human readable parameter types in the case class, and have the ByteString conversions only happen internally.

This would allow for the removal of some of the unnecessary object declarations, which would streamline the code a bit, and also maybe clean up the API a bit?

Thoughts?  Am I totally missing the mark on how these are going to be used?

DanSimon: That makes sense to me, I think the only place `ByteString` should be exposed as a public parameter type here is when you're reading/writing raw data from/to memcached.

nsauro:
bump.  Made some changes

nsauro:
PR updated with all the integration test stuff.  Build is failing b/c it looks like typesafe.artifactoryonline.com is hosed.

Going to update the PR summary

nsauro:
bump.  final.  Ready for review.  Give it hell

nsauro:
build is failing right now b/c typesafe.artifactoryonline is 404'ing

colossus-tests/src/it/scala/colossus/MemcacheITSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+
+import scala.concurrent.duration._
+/*
+Please be aware when running this test, that if you run it on a machine with a memcached server
+running on 11211, it will begin to interact with it.  Be mindful if you are running this on a server
+which has data you care about.
+This test runs by hitting the REST endpoints exposed by teh TestMemcachedServer, which just proxies directly to
+a Memcached client, which communicates with memcached
+ */
+class MemcacheITSpec extends HttpServiceSpec{
```
DanSimon: So these tests seem pretty good, but I am kind of confused why we need to go through the process of starting up a http server to test a memcache client.  Since we're just testing the protocol and not any of the client connection semantics, we should be able to test with just an `AsyncServiceClient` and directly check the replies from Memcached.  Not a huge deal though.

nsauro:
another bump.  scrapped the http layer in the INT tests

DanSimon:
:+1:

PR #100, opened 2015-01-29
Raise default max http parser size to 10MB as a temporary measure pendin...

zgagnon:
...g codec configuration ability.


DanSimon:
LGTM, ignore the travis failures, unrelated to this.

Issue #103, opened 2015-01-30
Cleanup ServiceClient binding logic


DanSimon:
So we've got some weirdness with how connection handlers are created and bound to workers.

Originally the constructor for `ServiceClient` required a `WorkerRef` as a parameter.  This meant when you wanted to open a connection from outside the IOSystem, you'd give the worker manager a closure of type `WorkerRef => ConnectionHandler`, which the worker itself would execute.

This was changed when we broke out the logic for binding worker items, such that the worker was no longer a constructor parameter, and the user would create the connection handler themselves before sending it to the worker manager.  Then the worker would bind the connection handler.

I think we should revert that for a few reasons:

1.  It makes it more difficult for connection handlers to setup metrics, since they don't have the worker (which contains the `MetricCollection`) when constructed.  Right now we get around this by making everything lazy, but that is kinda gross

2.  We now have the issue where if someone manually creates a service client, it is not bound to anything and won't connect.  Instead we have an alternative constructor that does take in a worker as a parameter and simply calls bind as soon as the connection handler is created.  This means the default constructor for ServiceClient should never be used, which is dumb.

3.  Originally I was thinking that the connection handler should not get a reference to a worker until it is bound to the worker, but this seems unnecessary, since connection handlers deterministically known when they are bound.

4.  Users directly opening connections from outside the IOSystem doesn't really happen anymore.  The only place it actually happens at all is with `AsyncServiceClient` but that's behind a pretty clean DSL.

So the work would just be to put the worker back in as a constructor param, and switch a few things to take closures again.  But I think it needs a little more investigation to make sure I'm not forgetting anything else.

Issue #98, opened 2015-01-29
Don't expose Connection binding/unbinding to users


DanSimon:
Users (people writing connection handlers) should not have to worry about, or have access to, the bind and unbind methods.  It may be as easy as just making the methods private

We should investigate if this is actually doable and implement if possible.

PR #107, opened 2015-01-30
Cleaning up binding logic

DanSimon:
This fixes both #98 and #103.  This mostly reverts some changes that were made a while ago to address problems that no longer exist.

* worker is once again a constructor parameter on service client.  This addresses some of the points in #103 
* The `bind`/`unbind` methods on `WorkerItem` are now private.  Users should never need to worry about manually binding items


colossus-tests/src/test/scala/colossus/core/ConnectionHandlerSpec.scala
```diff
@@ -92,7 +92,7 @@ class ConnectionHandlerSpec extends ColossusSpec {
       }
       withIOSystem{ implicit io =>
         //obvious this will fail to connect, but we don't care here
-        io ! IOCommand.BindAndConnectWorkerItem(new InetSocketAddress("localhost", TEST_PORT), new MyHandler)
+        io ! IOCommand.BindAndConnectWorkerItem(new InetSocketAddress("localhost", TEST_PORT), _ => new MyHandler)
```
nsauro: Bit of a drive by comment, as I'm about to run out the door, and Naomi is on the tear, but, here is a slightly half baked thought:

object BindAndConnectWorkerItem {
    implicit def liftItemToFunc(item : WorkerItem) : (WorkerRef => WorkerItem) = _ => item
  }

Then, in these places where we are really just using the function to wrap an instance, we can just import that implicit and the code doesn't change here, it will remain 
io ! IOCommand.BindAndConnectWorkerItem(new InetSocketAddress("localhost", TEST_PORT), new MyHandler)

Thoughts?

nsauro:
Shouldn't disconnect/gracefulDisconnect call unbind?  If not, they will stick around in The Worker.

DanSimon:
I think there might be a few bugs in that regard.  It also looks like if a client connection gives up attempting to reconnect, it is never unbound from the worker.

I think I will open a separate issue on that, because I think we might want to rethink a little about how that works.

DanSimon:
Ok, I think I have a solution for the unbinding issue.  I need to open an intermediary branch for a pre-requisite, but I will include the end result into this PR.

DanSimon:
Wow this PR got all crazy, gonna rebase and re-open

PR #108, opened 2015-02-02
graceful disconnect allows input stream to finish, fixed pipe bug

### LINKED ISSUE #102: ServiceClient's gracefulDisconnect only takes current writes into consideration
Right now i can start a read from a remote source, and then use gracefulDisconnect to schedule a disconnect for when the read is complete.  Graceful disconnect currently only allows for the current writing operation to finish, and immediately disconnects any read.

This can probably be (easier) to implement with Dan's idea of using a state machine inside of the Controller.

DanSimon:
This fixes #102

### DIFF
--- a/colossus/src/main/scala/colossus/controller/Controller.scala
+++ b/colossus/src/main/scala/colossus/controller/Controller.scala
@@ -190,16 +190,23 @@ extends ConnectionHandler {
    * pending/ongoing write operations to complete before disconnecting
    */
   def gracefulDisconnect() {
-    endpoint.foreach{_.disableReads()}
+    if (currentSink.isEmpty) {
+      endpoint.foreach{_.disableReads()}
+    }
     disconnecting = true
     //maybe wait instead for the message to finish? probably not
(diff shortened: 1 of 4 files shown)


nsauro:
:+1:

colossus/src/main/scala/colossus/controller/Controller.scala
```diff
@@ -190,16 +190,23 @@ extends ConnectionHandler {
...
     //maybe wait instead for the message to finish? probably not
-    currentSink.foreach{_.terminate(new DisconnectingException("Connection is closing"))}
-    currentTrigger.foreach{_.cancel()}
     checkGracefulDisconnect()
   }
 
   private def checkGracefulDisconnect() {
-    if (disconnecting && waitingToSend.size == 0 && currentlyWriting.isEmpty && currentStream.isEmpty) {
+    if (
+      disconnecting &&
```
nsauro: lol.  Yea..we need to do something about all these states :)

Issue #109, opened 2015-02-02
Multiple bugs with unbinding client connections


DanSimon:
So currently, when a client connection handler is unregistered for a connection, it is not automatically unbound from the worker.  This is because in the situation where a client connection is disrupted and it is attempting to reconnect, we do not want to unbind it from the worker, since it may miss external messages in the process (very important with AsyncServiceClient).

However, right now there are at least 2 situations where client connections are not unbound when they should be:

1.  When the user manually disconnects
2.  When a client connection gives up trying to reconnect

We should at the very least patch those bugs, or even better figure out if there's a more sane way to handle this.  Maybe possibly move this logic into the worker.

DanSimon:
So what I plan to do is move all of the unbinding logic into the `unregisterConnection` method inside the Worker.  The Worker will look at both the disconnect cause (to see if it's an error cause) and the connection (to see if it's a server or client) and take action accordingly.

The basic logic is server connections are always unbound automatically, and client connections are only unbound on non-error disconnects.  The one remaining issue is when clients fail to connect/reconnect and do not retry.  Only in that case do they need to actually tell the worker to unbind the handler.

I will be roping this into #107

Issue #42, opened 2014-12-10
AsyncServiceClient infinite retry loop


nsauro:
If an AsyncServiceClient fails to connect,  a race condition can arise where the underlying Worker will send the AsyncServiceClient's proxy actor a "MessageDeliveryFailed", which in turn the Proxy will send back to the Worker.

DanSimon:
Quick update to this issue.  I believe point 1 in my previous comment has since been fixed and was due to the fact that clients could not receive messages while reconnecting, which was solved by separating the connection logic from the worker binding logic.

The second point is still an issue.

PR #113, opened 2015-02-04
Private binding2

DanSimon:
This PR fixes a whole bunch of stuff:

* #103 : A `WorkerRef`was previously a constructor param on `ServiceClient`.  It was removed for reasons that are no longer an issue, so now it's back.  This allows `ServiceClients` to have the worker available in their constructor.
* #98  : just made the `bind` and `unbind` methods on `WorkerItem` private.  Users never need to call these manually
* #109 : The biggest issue here is in multiple situations, connection handlers were not being unbound when their associated connection closed.  This was mostly because for client handlers, the default behavior was to NOT unbind them, and a `AutoUnbindHandler` mixin existed for handlers to extend.  This has now been inverted such that the default behavior is to automatically unbind handlers, except for those that mixin the new `ManualUnbindHandler`.  For example, `ServiceClient` mixes this in, so it must unbind itself when it stops trying to reconnect.  Also, this only happens when the `DisconnectCause` is an error cause, otherwise the handler is automatically unbound always.
* #42 : Finally!  Changed the logic of `AsyncServiceClient` so it is aware when its associated handler is unbound.

There's a few small issues still, possibly a bug in `ServiceClient` regarding pending requests, but this PR is large enough.


colossus-tests/src/test/scala/colossus/core/ConnectionHandlerSpec.scala
```diff
@@ -66,14 +67,20 @@ class ConnectionHandlerSpec extends ColossusSpec {
       val probe = TestProbe()
       class MyHandler extends BasicSyncHandler {
         override def onUnbind() {
+          println("UNBINDING!!!!!!!")
```
nsauro: we can probably remove this println.  On the other hand..we REALLY need to start getting our logging situation sorted out.

colossus-tests/src/test/scala/colossus/core/WorkerItemSpec.scala
```diff
@@ -77,7 +77,7 @@ class WorkerItemSpec extends ColossusSpec {
           }
         }
         val item = new MyItem
-        io ! IOCommand.BindWorkerItem(item)
+        io ! IOCommand.BindWorkerItem(_ => item)
```
nsauro: So, I take it you didn't like the implicit lifting idea?

colossus-tests/src/test/scala/colossus/core/ConnectionHandlerSpec.scala
```diff
@@ -111,34 +131,74 @@ class ConnectionHandlerSpec extends ColossusSpec {
...
+        probe.expectMsg(500.milliseconds, "UNBOUND")
+      }
+
+    }
+
+    "automatically unbind on failure to connect" in {
+      val probe = TestProbe()
+      withIOSystem{ implicit io =>
+        io ! IOCommand.BindAndConnectWorkerItem(
+          new InetSocketAddress("localhost", TEST_PORT), _ => new MyHandler(probe.ref, true, true)
```
nsauro: I"m missing something here.  What is causing the connection failure here?  This looks like the above test, sans the 'end(server)'.

colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
```diff
@@ -180,7 +179,7 @@ class ServerSpec extends ColossusSpec {
         }
       }
 
-      "open up spot when connection closes" in {
+      "open up spot when connection closes" taggedAs(org.scalatest.Tag("test")) in {
```
nsauro: What's this Tag used for?

colossus/src/main/scala/colossus/core/Worker.scala
```diff
@@ -470,11 +396,18 @@ private[colossus] class Worker(config: WorkerConfig) extends Actor with ActorMet
           con.handleConnected()
         } catch {
           case c: java.net.ConnectException => {
+            //TODO consider rolling in connection failure as a disconnect cause
```
nsauro: These 2 catches are pretty much identical, sans the log msg.  Should we combine them?

nsauro:
Looks good.  One outstanding question I have is around how the disconnect behavior should work with regards to streaming.  We can look into implementing in a separate PR, but we have  several places in Controller where we still have TODO's around pulling and pushing data errors.  Should we just bubble up those exceptions so that it gets linked into this Disconnect logic?  Def something to discuss.

colossus-tests/src/test/scala/colossus/core/WorkerItemSpec.scala
DanSimon: I think it's kind of overkill, especially since this is more of an internal thing and usually should not be used by users.

colossus-tests/src/test/scala/colossus/core/ConnectionHandlerSpec.scala
DanSimon: No server is started

colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
DanSimon: it was for testing just that test, I can remove it

DanSimon:
Yeah I was looking at some of those and I think in most of those cases the only solution is to kill the connection.  In any case this stuff deals more with how a disconnect happens as opposed to when it should happen.

colossus/src/main/scala/colossus/core/Worker.scala
DanSimon: Yeah I can eliminate the duplicate code, but I do want to maintain that behavior, since `ConnectException` is normal but anything else in that situation would be something really unexpected and should be noted.

PR #117, opened 2015-02-09
Stream content length

DanSimon:
This just adds in the content-length header when creating a stream from a static `HttpResponse`.


nsauro:
groovy

PR #118, opened 2015-02-09
metric systems get a unique id to avoid tick conflicts, system metrics optional

DanSimon:
Killing two birds with one stone:

* Since the `Tick` messages used to collect metrics from various sources is broadcast on the event bus, if there are multiple metric systems, they end up reacting to ticks from the wrong systems.  Now MetricSystems have a unique id, which is included in the `Tick` message and checked by anyone processing it.
* You can now configure whether a metric system collects system-level metrics.


nsauro:
ship it.  looks good.  Once this is merged I'll build everything locally and see how it works in the POC I'm working on for the stats service.

Issue #115, opened 2015-02-05
Clean up streaming error handling logic


DanSimon:
Currently there's a bunch of places in the controller where we haven't fully decided how to handle errors.  We should solidify the expected behavior and patch things up

PR #122, opened 2015-02-12
Controller cleanup

### LINKED ISSUE #119: Clean up controller logic
Right now the Controller layer is basically one big class with a bunch of different states.  I think we can clean this up a lot to help ensure we're properly handling all possible states and also make the code easier to follow.

DanSimon:
Fixes #119 and probably #115 as well.

Well, this basically turned into almost a total rewrite of the Controller class.  We now have 2 main traits, one for input and one for output, which are both pulled into `Controller`.  In addition, each sub-controller has been rewritten as a state machine, which I think helps make the code easier to read as well as eliminate some of the nonsensical states we were hitting previously.

There are still a few areas where we have some iffy or undefined logic, but this change should make it easier to identify and address those places.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/service/ServiceClientSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/service/ServiceClientSpec.scala
@@ -194,7 +194,7 @@ class ServiceClientSpec extends ColossusSpec {
       response must equal(Some(reply))
     }
 
-    "fail pending requests on disconnect with failFast" in {
+    "fail pending requests on disconnect with failFast" taggedAs(org.scalatest.Tag("test")) in {
       val command1 = Command(CMD_GET, "123456789012345678901234567890")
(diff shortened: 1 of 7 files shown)


colossus-tests/src/test/scala/colossus/service/ServiceClientSpec.scala
```diff
@@ -194,7 +194,7 @@ class ServiceClientSpec extends ColossusSpec {
       response must equal(Some(reply))
     }
 
-    "fail pending requests on disconnect with failFast" in {
+    "fail pending requests on disconnect with failFast" taggedAs(org.scalatest.Tag("test")) in {
```
nsauro: This tag 'test' is back.  Intentional?

nsauro:
I went over this locally, gh is not the best for large PRs.  It looks good..it does make things easier to follow, so I think going forward this will help a lot.

PR #125, opened 2015-02-17
Removing internal buffering from pipes

### LINKED ISSUE #123: Remove internal buffering from Infinite Pipes
There's a bug where the internal buffering of pipes can result in lost data.  If the consumer of a pipe is not pulling fast enough, the pipe will begin buffering items internally.  For `DataBuffers`, the buffer must be copied.  However if not all the data in the databuffer is for the consumer, the remaining bytes are dropped and lost forever.

This would occur, for example, if a http stream receives two chunked-transfer responses one after another.  It's likely a single `DataBuffer` will contain bytes for both the last chunk of the first response and part of the header of the second.  If the consumer of the first response is too slow and the pipe copies that `DataBuffer`, the consumer will read the last chunk and drop the first bytes of the next response.

The most immediate solution is to totally remove this internal buffering.  Thus a pipe never has to copy the buffers, and instead the controller will hold onto a copy itself and immediately disable reads from the event loop.  This is actually probably the better way to handle backpressure anyway, as it totally avoids any buffer copying that could fill up the heap.

DanSimon:
This fixes #123 and addresses a serious bug in how data is pushed into streams.  Previously, when the controller pushed a `DataBuffer` into a stream, and the receiving side of the pipe could not immediately process the data, the pipe would copy the buffer and queue it, while the controller would see that as a success and thus toss the `DataBuffer`.  However, it's possible that the `DataBuffer` contained both the end of that stream as well as the beginning of the next stream.  Since the receiver would just ignore the extra data, the data would be lost forever.

This PR almost completely changes the semantics of pipes such that they are more of a synchronization mechanism now than an intermediary buffer.  The removal of any internal buffering totally eliminates the above problem as well as another issue dealing with the unnecessary copying of `DataBuffers` and worrying about queue sizes.  Now any backpressure is immediately escalated to the controller.

* Pipes are now implemented as a state machine.
* Since there is no internal buffering, pushing a value into a pipe only succeeds if the puller has already made a request.  The return `PushResult` of a push operation then depends on whether the puller's callback calls `pull` again or `complete`.  Therefore, with every push operation, the pusher will know both the success of the push and whether the pipe is able to receive more data based only on the return value.

I think there are two more things that need to be done.

* FiniteBytePipe still does some `DataBuffer` copying, and it does some crazy logic to avoid doing it when it doesn't have to.  If we supply a way to "slice" a `DataBuffer`, all that will be eliminated.
* We might want to change the signature of the pull callback, such that it's return value is an ADT rather than `Unit`, with values like `Pull`, `PullWith(new_callback)`, or `Complete`.  This would totally break the ability to turn the pull method into returning a callback, but I think this is not important.  It's more likely users will either want to loop over all the items being pulled, or use a callback just to fold the stream into a single value, which will still be doable.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/controller/PipeSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/controller/PipeSpec.scala
@@ -8,6 +8,8 @@ import akka.util.ByteString
 import core.DataBuffer
 import scala.util.{Success, Failure}
 
+import PushResult._
+
 class PipeSpec extends WordSpec with MustMatchers with CallbackMatchers {
 
   implicit object IntCopier extends Copier[Int] {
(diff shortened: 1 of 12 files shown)


colossus/src/main/scala/colossus/controller/Pipe.scala
```diff
@@ -240,102 +243,98 @@ class DualSource[T](a: Source[T], b: Source[T]) extends Source[T] {
...
+/**
+ * This is a special exception that Input/Output controllers look for when
+ * error handling pipes.  In most cases they will log the error that terminated
+ * the pipe, but for this one exception, the failure will be silent.  This is
+ * basically for situations where a certain amount of data is expected but for
+ * some reason the receiver decides to cancel for some business-logic reason.
+ */
+class PipeCancelledException extends Exception("Pipe Cancelled") with PipeException
+
+class InfinitePipe[T] extends Pipe[T, T] {
```
nsauro: def diggin the state machine based impl.  cleans up stuff

colossus-tests/src/test/scala/colossus/controller/PipeSpec.scala
```diff
@@ -221,18 +183,21 @@ class PipeSpec extends WordSpec with MustMatchers with CallbackMatchers {
 
     }
 
-    "properly calculates bytes remaining" in {
+    "properly calculates bytes remaining" taggedAs(org.scalatest.Tag("test")) in {
```
nsauro: the tag has returned

nsauro:
pretty big PR..but I think it looks ok..nothing stood out to me.  I did notice that you changed the StreamingHttpResponse.fromStatic..was that for an unrelated bug?

nsauro:
:100:

DanSimon:
Yeah that's because you can no longer push to a pipe right after creating it.  Easier just to create it as a Source.one.  I almost think we might want to change the name `Pipe` to something else, since its semantics are quite different from e.g. a unix pipe, but for now whatever.

Issue #114, opened 2015-02-05
Use last write time when checking if a connection is idle


DanSimon:
Currently, only the last time data was read is used to determine if a connection is idle.  This will incorrectly result in connections that are only writing data to appear as idle.  We need to also take into account the last time data is written.

PR #126, opened 2015-02-17
Idle timeout fixes

### LINKED ISSUE #88: ClientConnections can never be timed out
We should probably make ClientConnections timeoutable.  Right now, ClienConnections never time out, so you have a potential connection leak if an application is constantly creating new connections due to streaming or such.

DanSimon:
Fixes #88 and #114 

* The new `maxIdleTime` on `ClientConnectionHandler` gives client connections a way to specify a max timeout.  This works differently from server connections, since not only are all server connections configured the same way, their idle timeout can change based on the server's number of open connections.  Client connections are largely independent of each other, so it makes the most sense to put the dependency here.  
* Connections now keep track of the last time they successfully wrote data on the socket, and use this value to determine if it is idle.

### DIFF
--- a/colossus/src/main/scala/colossus/core/Connection.scala
+++ b/colossus/src/main/scala/colossus/core/Connection.scala
@@ -106,6 +106,11 @@ private[core] abstract class Connection(val id: Long, val key: SelectionKey, _ch
   def isTimedOut(time: Long): Boolean
   def unbindHandlerOnClose: Boolean
 
+  def timeIdle(currentTime: Long) = currentTime - math.max(lastTimeDataReceived, lastTimeDataWritten)
+
+  protected def isTimedOut(maxIdleTime: Duration, currentTime: Long): Boolean = {
+    maxIdleTime != Duration.Inf && timeIdle(currentTime) > maxIdleTime.toMillis
+  }
 
   protected val channel = _channel
   val worker = sender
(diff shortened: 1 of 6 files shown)


colossus/src/main/scala/colossus/core/Connection.scala
```diff
@@ -179,7 +184,7 @@ private[core] class ServerConnection(id: Long, key: SelectionKey, channel: Socke
     super.close(cause)
   }
 
-  def isTimedOut(time: Long) = server.maxIdleTime != Duration.Inf && time - lastTimeDataReceived > server.maxIdleTime.toMillis
+  def isTimedOut(time: Long) = server.maxIdleTime != Duration.Inf && time - math.max(lastTimeDataReceived, lastTimeDataWritten) > server.maxIdleTime.toMillis
```
nsauro: The 'isTimedOut' function is identical in both ServerConnection and ClienConnection.  The only difference being of where the 'maxIdleTime' comes from.  Can we make a function on Connection:

isTimedOut(maxIdleTime : Duration, time : Long) =  maxIdleTime != Duration.Inf && time - math.max(lastTimeDataReceived, lastTimeDataWritten) > maxIdleTime.toMillis

This would just move this code into a common location

colossus/src/main/scala/colossus/core/Connection.scala
```diff
@@ -74,6 +74,11 @@ trait WriteEndpoint {
...
    */
   def enableReads()
+
+  /**
+   * Returns a timestamp, in milliseconds of when the last time data was written on the connection
+   */
+  def lastTimeDataWritten: Long
 }
 
 private[core] abstract class Connection(val id: Long, val key: SelectionKey, _channel: SocketChannel, val handler: ConnectionHandler)(implicit val sender: ActorRef)
```
nsauro: Grr..I cannot comment on the line directly, but info function's way of calculating timeIdle needs to change.  It is only accounting for lastTimeDataReceived.  Do we want to also include lastTimeDataWritten?

nsauro:
other than those 2 small things..looks good :100:

DanSimon:
Gonna close this and merge --squash manually

PR #127, opened 2015-02-17
Removing the Actor API

### LINKED ISSUE #105: Consider eliminating Actor-based API
Currently in the core layer we have a fully actor-based API that wraps a connection handler.  It is very similar to the Akka I/O API.  The stream example in the examples project uses it.

Aside from the example, nobody has ever used it, nobody has even expressed interest in it, and I think we should kill it.  It was created in the first few days of Colossus' development and I think is now clearly outside the scope of what we want the framework to do.  The functionality in the stream example can now be done with the Controller layer.

Thoughts?

DanSimon:
Fixes #105 , which we decided was a good idea.

This actor-based API is basically a clone of the Akka I/O API, except not quite as good.  It mostly worked, except nobody every used it and nowadays it definitely is outside the scope of what we want the framework to include out-of-the-box.

A handful of pieces are used in other areas so they have been relocated to more appropriate files.


nsauro:
:100:

PR #128, opened 2015-02-17
Removing shared interface from ServiceClient, LoadbalancingClient

### LINKED ISSUE #104: Consider eliminating shared interface on service clients
Currently `ServiceClient` and `LoadBalancingClient` have a `shared` method which returns a new interface that is thread safe.  This was created back when we had more implicit conversions between `Callback` and `Future`, but now that we have made that conversion much more explicit, having this shared interface actually doesn't make much sense.

Whereas before, if the flow of request processing moved into a `Future`, the shared interface could be used to chain a new future directly to it.  Now it makes more sense to just wrap the future in `Callback.fromFuture` and turn it into a `Callback`, and proceed from normal there.

Needs some more thought/discussion if there are any legitimate use cases for the shared interface.

DanSimon:
This fixes #104 .  It is, perhaps unsurprisingly, not a lot of code.  

This was developed back when we thought there'd be more fluidity between Callbacks and Futures, such that you would want to directly use a client within a Future.  This is no longer the case, and instead any Futures that arise in the execution of a request should be converted into a Callback before any mapping, flatmapping.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/service/LoadBalancingClientSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/service/LoadBalancingClientSpec.scala
@@ -27,15 +27,8 @@ class LoadBalancingClientSpec extends ColossusSpec with MockitoSugar{
       requestTimeout = 1.second
     )
     val r = customReturn.getOrElse(Success(address.getPort))
-    val m = mock[SharedClient[String, Int]]
-    //future.fromTry not in scala 2.10
-    when(m.send("hey")).thenReturn(r match {
-      case Success(r) => Future.successful(r)
-      case Failure(err) => Future.failed(err)
-    })
     val c = mock[ServiceClient[String, Int]]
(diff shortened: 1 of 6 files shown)


colossus-tests/src/test/scala/colossus/service/LoadBalancingClientSpec.scala
```diff
@@ -27,15 +27,8 @@ class LoadBalancingClientSpec extends ColossusSpec with MockitoSugar{
       requestTimeout = 1.second
     )
     val r = customReturn.getOrElse(Success(address.getPort))
-    val m = mock[SharedClient[String, Int]]
```
nsauro: Perhaps change the name of the file from LoadBalancingSpec to something else?

nsauro:
disregard the comment about changing the name of the spec

nsauro:
+1

PR #134, opened 2015-02-26
Small tweaks to metrics including scaladoc

zgagnon:
I've added scaladoc for some of the metrics api and also expanded the LoggerSender to allow a custom format.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
@@ -2,17 +2,28 @@ package colossus.metrics
 
 import akka.actor._
 import akka.agent.Agent
-import akka.pattern.ask
+import colossus.metrics.MetricAddress.Root
 
-import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
-import akka.util.Timeout
-import MetricAddress.Root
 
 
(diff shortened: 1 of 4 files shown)


colossus-metrics/src/main/scala/colossus/metrics/senders/LoggerSender.scala
```diff
@@ -22,11 +23,18 @@ class LoggerSenderActor extends Actor with ActorLogging with  MetricsLogger {
   override def postStop() {
     log.info("shutting down dummy sender")
   }
+}
+
+class LoggerSender(val formatter:Formatter) extends MetricSender {
+  override def name: String = "logger"
```
DanSimon: These shouldn't have to override

colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
```diff
@@ -2,17 +2,28 @@ package colossus.metrics
...
-import akka.util.Timeout
-import MetricAddress.Root
 
 
 case class MetricSystemId(id: Long)
 
-
+/**
+ * The MetricSystem is a set of actors which handle the background operations of dealing with metrics. In most cases,
+ * you only want to thave one MetricSystem per application.
```
nsauro: thave = have

nsauro:
:+1:

Issue #133, opened 2015-02-25
add idle / timeout metric rates


nsauro:
Just nice to have some kind of stats around these.

Issue #132, opened 2015-02-25
race condition with Server registration


nsauro:
If a worker dies, it will be resuscitated, and a server will try to register itself.  However, there is a case where the server is not yet registered with the reborng Worker,and is sending connections to it, who then rejects them.

PR #136, opened 2015-02-26
Http refactoring, chunked transfer support

DanSimon:
There's still work to be done, but this PR is getting pretty big and basically everything pre-existing works.

A whole bunch of changes are made here to better work with chunked transfer encoding in streaming http.  In particular we're beginning to address the issue that sometimes you want the body of a chunked request to be "dechunked" into just the raw data while other times (such as in a proxy) you want to leave the body totally untouched.  As the proxy situation is more relevant to our immediate needs, I focused on that use case for now, though getting the dechunking streams to work should not be much more work.

* Moved a whole bunch of stuff around to better organize things.  In particular there's now a proper `HttpResponseHead` similar to the `HttpHead` (soon to be renamed) for `HttpRequest`.  Also we had mixed up transfer-encodings with content-encodings, so those are now properly separated.
* There's now more of a separation between the Http and Streaming http codecs.  There's probably room to merge some stuff together, but right now it's much easier to iterate if we keep everything separated for now.
* `DataBuffer` has a new `peek` method which allows you to read from the buffer inside a closure, and afterwards the buffer is reset to the state before the closure began.  This is primarily needed for the next point...
* By default now, a streaming HttpClient will not dechunk the body.  This is enabled with the new `ChunkPassThroughPipe` which is able to parse the chunk headers to keep track of when the stream ends, but doesn't actually consume any data.

things that still need to be done (I'll move these into one or more issues) :

* Implement the pipe that that will dechunk the http body.
* Possibly indicate -- either as a new field in `HttpResponse` or possibly as a wrapper ADT around `DataBuffer` -- whether the attached `Source` is being dechunked or not.
* Probably move the `Transfer-Encoding`, `Content-Length`, and `Content-Encoding` headers into some kind of first-class value on `HttpResponseHead`.  This will make it easier (and probably faster) to figure out exactly what the body of the Response looks like.
* Generalize all the above to work for `HttpRequest` as well.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/protocols/http/HttpPipeSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/protocols/http/HttpPipeSpec.scala
@@ -45,7 +45,7 @@ class HttpPipeSuite extends WordSpec with MustMatchers{
 
     }
 
-    "terminate the pipe on parsing exception" ignore {
+    "terminate the pipe on parsing exception" in {
       val p = new ChunkPassThroughPipe
       val c = foldTester(p)
       p.push(DataBuffer(badChunkedBody)) mustBe a[PushResult.Error]
--- a/colossus-tests/src/test/scala/colossus/protocols/http/HttpPipeSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/protocols/http/HttpPipeSpec.scala
@@ -1,0 +1,72 @@
+package colossus
+package protocols.http
+
+import core._
+import controller.Pipe
+import controller.PushResult
+import testkit.PipeFoldTester
+
+import org.scalatest._
+
+import akka.util.ByteString
+
+import service.Callback
+import scala.util.{Try, Success, Failure}
+
+
+
+
+class HttpPipeSuite extends WordSpec with MustMatchers{
+    
+  val chunkedBody = ByteString("3\r\nfoo\r\nb\r\n12345678901\r\n0\r\n\r\n")
+  val badChunkedBody = ByteString("3\r\nfoo\r\n1WHATTT\r\n12345678901\r\n0\r\n\r\n")
+
+
+  def foldTester[A](p: Pipe[A, DataBuffer]): PipeFoldTester[ByteString] = new PipeFoldTester(p.fold(ByteString())(PipeFoldTester.byteFolder))
+
+  "ChunkPassthroughPipe" must {
+
+    "leave the data untouched" in {
+      val p = new ChunkPassThroughPipe
+      val c = foldTester(p)
+      p.push(DataBuffer(chunkedBody)) must equal(PushResult.Complete)
+      c.expect(chunkedBody)
+    }
+
+    "properly detect the end of the stream" in {
+      val p = new ChunkPassThroughPipe
+      val c = foldTester(p)
+
+      val extra = ByteString("ADSFASDFASDF")
+      val data = DataBuffer(chunkedBody ++ extra)
+      p.push(data) must equal(PushResult.Complete)
+      c.expect(chunkedBody)
+      data.remaining must equal(extra.size)
+
+    }
+
+    "terminate the pipe on parsing exception" ignore {
+      val p = new ChunkPassThroughPipe
+      val c = foldTester(p)
+      p.push(DataBuffer(badChunkedBody)) mustBe a[PushResult.Error]
+      c.expectFailure()
+    }
+
+
+  }
+
+  "ChunkDecodingPipe" must {
+
+    "Process a chunked body" in {
+      val expected = ByteString("foo12345678901")
+      val p = new ChunkDecodingPipe
+      val c = foldTester(p)
+      p.push(DataBuffer(chunkedBody)) must equal(PushResult.Complete)
+      c.expect(expected)
+    }
+
+
+  }
+
+
+}
--- a/colossus-tests/src/test/scala/colossus/protocols/http/client/StreamingHttpResponseParserSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/protocols/http/client/StreamingHttpResponseParserSpec.scala
@@ -5,11 +5,13 @@ import colossus.controller._
 import colossus.core.{DataBuffer, DataStream, DataReader}
 import colossus.protocols.http._
 import colossus.service.{Callback, DecodedResult}
-import colossus.testkit.CallbackMatchers
+import colossus.testkit.{PipeFoldTester, CallbackMatchers}
 import org.scalatest.{WordSpec, MustMatchers, TryValues, OptionValues}
 import scala.util.{Try, Success, Failure}
 
 class StreamingHttpResponseParserSpec extends WordSpec with MustMatchers with TryValues with OptionValues with CallbackMatchers {
+
+
 
 
   "StreamingHttpResponseParser" must {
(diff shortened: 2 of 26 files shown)


colossus/src/main/scala/colossus/protocols/http/package.scala
```diff
@@ -72,4 +80,13 @@ package object http {
       worker = worker
     )
 
+    */
+  implicit object ByteStringLikeString extends ByteStringLike[String] {
```
nsauro: Can we keep these colocated with the ByteStringLike trait?

colossus/src/main/scala/colossus/protocols/http/HttpServerCodec.scala
```diff
@@ -6,13 +6,14 @@ import akka.util.ByteStringBuilder
 import colossus.controller.{DualSource, Source}
```
nsauro: we have some unused imports here.

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
```diff
@@ -3,68 +3,231 @@ package protocols.http
...
-  val DefaultQueueSize : Int = 100
-
-  //not totally in love with this extra object allocation, but feel its more straightforward than extractors
-  case class ParsedHeaders(headers : Seq[(String, String)]) extends HttpHeaderUtils
-
-  def static(size: DataSize = DefaultMaxSize) : HttpResponseParser[HttpResponse] = {
-    new HttpResponseParser[HttpResponse](() => maxSize(size, staticResponse))
-  }
+  //TODO: eliminate duplicate code
+  //TODO: Dechunk on static
```
nsauro: Is this comment still valid?  It looks like you are already dechunking static responses.

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
```diff
@@ -3,68 +3,231 @@ package protocols.http
...
-    val encoding = pHeaders.getEncoding
-    encoding match {
-      case TransferEncoding.Identity => clength match {
-        case Some(0) | None => const(HttpResponse(version, code, headers, ByteString()))
-        case Some(n) => bytes(n) >> {body => HttpResponse(version, code, headers, body)}
+  def streamBody(dechunk: Boolean): Parser[DecodedResult[StreamingHttpResponse]] = head >> {parsedHead =>
+    parsedHead.transferEncoding match {
+      case TransferEncoding.Identity => parsedHead.contentLength match {
+        case Some(0) | None => DecodedResult.Static(StreamingHttpResponse(parsedHead, None))
+        case Some(n) => streamingResponse(parsedHead, Some(n), false)
```
nsauro: If you want to avoid another allocation, you could just a @ Some(n) => streamingResponse(...a..)

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
```diff
@@ -3,68 +3,231 @@ package protocols.http
...
-case class StreamingHttpResponseHeaderStub(version : HttpVersion, code : HttpCode, headers : Seq[(String, String)]) extends HttpResponseHeader
\ No newline at end of file
+/** A pipe designed to accept a chunked http body.
+ *
+ * The pipe will scan the chunk headers and auto-close itself when it hits the
+ * end of the stream (reading a 0 length chunk).  It also strips all chunk
+ * headers from the data, so the output stream is the raw data
+ *  
+ */
+class ChunkDecodingPipe extends InfinitePipe[DataBuffer] {
```
nsauro: Silly request:  can you move this near the ChunkEncodingPipe?  Just feels right :)

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -1,77 +1,151 @@
...
+import akka.util.{ByteString, ByteStringBuilder}
+import colossus.controller.Source
 import colossus.core.DataBuffer
+import core._
+import service.Callback
+import controller._
+import HttpParse._
+
+
+case class HttpResponseHead(version : HttpVersion, code : HttpCode, headers : Vector[(String, String)] = Vector()) extends HttpHeaderUtils {
```
nsauro: Not a comment for this PR, but in general, I've been increasingly noticing our headers aren't a case class, but a tuple(which is essentially a case class).  SHould we just make it a case class and have nice names on it?

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -1,77 +1,151 @@
...
+  def head: HttpResponseHead
+
+  /**
+   * Resolves the body into a bytestring.  This operation completes immediately
+   * for static responses.  For streaming responses, this only completes when
+   * all data has been received.  Be aware if the response is an infinite
+   * stream of data (using chunked transfer encoding), this will never
+   * complete.
+   */
+  def resolveBody(): Option[Callback[ByteString]] = this match {
```
nsauro: Indifferent remark: Any reason why this is just not abstract with each subclass providing their own impl?

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -1,77 +1,151 @@
...
+
+  def withHeader(key: String, value: String) = copy(head = head.withHeader(key,value))
+
 }
 
-case class StreamingHttpResponse(version : HttpVersion, code : HttpCode, headers : Seq[(String, String)] = Nil, stream : Source[DataBuffer]) extends HttpResponseHeader {
-  def withHeader(key: String, value: String) = copy(headers = headers :+ (key, value))
-  
+/**
+  * Converter typeclass for bytestrings.  Default implementations are in package.php
```
nsauro: God, I really hope that isn't the case :)

colossus/src/main/scala/colossus/protocols/http/package.scala
DanSimon: I think these should go here because otherwise you have to start importing them just to create responses with ByteStrings.  Either that or we should rework this so no imports are needed to create responses from bytestrings.

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -63,19 +63,28 @@ trait HttpHeaderUtils {
...
+  //headers are being parsed.  Benchmark before changing
+
+  /** Returns the value of the content-length header, if it exists.
+   * 
+   * Be aware that lacking this header may or may not be a valid request,
+   * depending if the "transfer-encoding" header is set to "chunked"
+   */
+  val contentLength: Option[Int] = headers.collectFirst{case (HttpHeaders.ContentLength, l) => l.toInt}
+
+  val transferEncoding : TransferEncoding = singleHeader(HttpHeaders.TransferEncoding).flatMap{TransferEncoding.unapply(_)}.getOrElse(TransferEncoding.Identity)
```
nsauro: I'm pretty sure you can turn TransferEncoding.unapply(_) into simply TransferEncoding.unapply.

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
DanSimon: yeah I'm thinking of doing that at the same time we move some of the special headers into separate vals.

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
DanSimon: That's what i get for multitasking :hamster:

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -113,25 +122,40 @@ object TransferEncoding {
...
+  private val all = Seq(Identity, Chunked)
+  def unapply(str : String) : Option[TransferEncoding] = {
+    all.find(_.value == str.toLowerCase)
+  }
+}
+
+sealed trait ContentEncoding {
+  def value: String
+}
+object ContentEncoding {
```
nsauro: ContentEncoding...are we planning on writing pipes/implementations for handling Accept Encoding on http requests?  Would be cool...but for a later date :)

colossus/src/main/scala/colossus/core/Worker.scala
```diff
@@ -189,6 +189,7 @@ private[colossus] class Worker(config: WorkerConfig) extends Actor with ActorMet
       }
       if (timedOut.size > 0) {
         log.debug(s"Terminated ${timedOut.size} idle connections")
+        //TODO: Tick Metric
```
nsauro: Both of these comments are for those issues we found when testing the 304 service, right?

colossus/src/main/scala/colossus/protocols/http/Header.scala
DanSimon: Yeah, I started to actually add more functionality for this, but backed off since it was less trivial than I first thought.

colossus-tests/src/test/scala/colossus/protocols/http/client/StreamingHttpResponseParserSpec.scala
```diff
@@ -109,21 +138,23 @@ class StreamingHttpResponseParserSpec extends WordSpec with MustMatchers with Tr
       }
 
     }
-    "decode a response that was encoded by colossus with no body" in {
+    "decode a response that was encoded by colossus with no body" ignore {
       validateEncodedStreamingHttpResponse(ByteString(""))
     }
 
-    "decode a response that was encoded by colossus with a body" in {
```
nsauro: Are these tests no longer valid?

colossus/src/main/scala/colossus/core/Worker.scala
DanSimon: Yeah  #132 and #133 is for these I believe.

colossus-tests/src/test/scala/colossus/protocols/http/client/StreamingHttpResponseParserSpec.scala
nsauro: We have several ignored tests in this spec..if they are no good, we should just delete them.  They look like they would be useful however.

DanSimon: At least one of them depends on #137 .  Once these PR's get merged in I'm going to make a pass and try to address all of the ignored tests.

colossus-tests/src/test/scala/colossus/protocols/http/HttpPipeSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+      val p = new ChunkPassThroughPipe
+      val c = foldTester(p)
+      p.push(DataBuffer(chunkedBody)) must equal(PushResult.Complete)
+      c.expect(chunkedBody)
+    }
+
+    "properly detect the end of the stream" in {
+      val p = new ChunkPassThroughPipe
+      val c = foldTester(p)
+      p.push(DataBuffer(chunkedBody ++ ByteString("ADSFASDFASDF"))) must equal(PushResult.Complete)
```
nsauro: Is there anyway to assert that the ADSF string is till in the buffer, and has not been dropped?  I know you were concerned about that case.

nsauro:
Phew..that took a good amount of time up.  All in all, looks pretty good..I didn't see anything glaring or incorrect.  The http classes are looking cleaned up..which, having a clean base will allow us to build on it some more(first class headers, etc)
:100:

PR #137, opened 2015-02-28
fixing bug in Callbacks where exception gets thrown too early

DanSimon:
Ran into this in another branch.  This is a pretty obscure, but serious issue where if an exception is thrown inside a Callback, and then the callback is flatmapped and then recovered, the exception was not properly propagated through the flatmap into the recover, but instead was thrown.  This was doubly weird since the stack trace on the Exception was totally wrong.

Now if in a `flatMap` operation, a `Failure` is passed into the mapping, the user-passed closure is skipped and the `Failure` is immediately pushed into the next phase of execution.  In other words, if you have `Callback[A].flatMap{a => Callback[B]}.recover{...}` and the first Callback throws an exception during execution, the user's `a => Callback[B]` is properly skipped and the recover method is invoked.


nsauro:
+1

PR #138, opened 2015-03-02
new binary parsers for short, int, long

### LINKED ISSUE #129: Binary Integer parser
right now we have an integer parser that parses the string representation with a terminal byte.  What we need is a byte parser than can turn the bytes into an n-byte integer, where n <=8.  It also needs to handle both big and little endian values.

DanSimon:
Fixes #129.  3 new parsers for parsing big-endian shorts, ints and longs.  These are different from `intUntil`, which parses an ASCII string as an integer.


nsauro:
:1

PR #139, opened 2015-03-02
wrapping clientConnect in try catch

### LINKED ISSUE #131: Worker clientConnect can throw
This will kill workers.  We should wrap this in a try/catch and handle appropriately.

DanSimon:
Fixes #131 .  

We might want to eventually add a `reason` parameter to `connectionFailed`, but for now that's out of scope.


nsauro:
:+1:   Do we want metrics on this?

DanSimon:
Not sure really.  The service layer is already tracking failed connections for clients, so basically in every situation we currently care about we do have metrics on this.  Might be worth thinking about moving that metric out of the service layer and into the core layer, but we don't really have any use cases yet that would directly benefit from it.

PR #141, opened 2015-03-04
fixing a bug where calling gracefulDisconnect in a callback threw an exception

DanSimon:
Basically what is happening is the client is trying to disconnect twice, leading to an invalid state.  It's possible we should change the state logic as well, but this fixes the immediate issue.


nsauro:
:100:

PR #142, opened 2015-03-04
workers now properly reject connections when server not registered

### LINKED ISSUE #132: race condition with Server registration

DanSimon:
Fixes #132 .  If a worker receives a connection from an unregistered server (which can happen if the worker was just restarted), the worker will then refuse the connection and send it back to the server.  The server will attempt to do this 3 times before giving up and killing the connection.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
@@ -1,7 +1,7 @@
 package colossus
+package core
 
 import testkit._
-import core._
 import service.AsyncServiceClient
 
 import akka.actor._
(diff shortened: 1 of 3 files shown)


colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
```diff
@@ -308,6 +308,26 @@ class ServerSpec extends ColossusSpec {
         alive() must equal(0)
 
       }
+
+      "attempt to re-register connection if refused by worker" taggedAs(org.scalatest.Tag("test")) in {
```
nsauro: is this tag being auto generated somehow?

DanSimon: No I throw it in when I want to just run one test but I keep forgetting to remove it afterwards :ghost:

colossus/src/main/scala/colossus/core/Server.scala
```diff
@@ -390,9 +397,19 @@ object ServerStatus {
  *
  */
 object Server {
+
+  val MaxConnectionRegisterAttempts = 3
```
nsauro: I'm ok w/ this being hardcoded for now..I can't imagine a case right now where this needs to be configurable, since this is really just controlling a hopefully rare case of Worker dying.

nsauro:
:100:

PR #143, opened 2015-03-05
added metric for timedout idle connections

### LINKED ISSUE #133: add idle / timeout metric rates

nsauro:
fixes #133


DanSimon:
I wonder if it makes more sense to have this metric in the server and not the worker.  For instance, we could have the server include the disconnect cause as a tag in the "closed" metric.  Then it might also make sense to add `IdleTimeout` as a new `DisconnectCause`, or something like that.

nsauro:
How would we handle timed out idle client connections?

nsauro:
We could just have both I suppose.  On timeout, we can check whether or not it was a client/server connection, and in the event of a server connection, send a message to a server, similar to how ConnectionClosed works now.  For client, we could just tick this in Worker metric and add SOME kind of identifying tag.

nsauro:
updated.

DanSimon:
:+1: 

For the record, it turns out we're smarter than we realized and Colossus already does almost all of this :tada:

PR #145, opened 2015-03-06
http servers now auto-disconnect if the version is 1.0

DanSimon:
This utilizes the ability for codecs to provide their own server handlers.  Now any `Http` or `StreamingHttp` service will properly kill connections using http 1.0


nsauro:
+1

PR #146, opened 2015-03-09
made ClientConfig available on an AsynsServiceClient

### LINKED ISSUE #45: cannot get ClientConfig from AsyncServiceClient
Just need a function that returns the ClientConfig.  Should be straightforward.

nsauro:
fixes #45


DanSimon:
:+1:

PR #148, opened 2015-03-12
Add support for decoding messages that depend on the connection closing

DanSimon:
This largely only applies for http responses that don't include a content-length, but instead close the connection as a means for signalling that the response is complete.

This requires a new `endOfStream` method in all of the parser combinators, along with a new `bytesUntilEOS` parser that basically aggregates all received data until the method is called.  This is then wired into `InputController` such that the method is called when the connection is closed and any returned message is processed.

Also support is added for responses that do not allow for a body, which also generally do not include a content-length header.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
@@ -22,24 +22,34 @@ object ResponseResult {
 object HttpResponseParser  {
   val DefaultMaxSize: DataSize = 10.MB
 
+  def static(maxResponseSize: DataSize = DefaultMaxSize): Parser[DecodedResult.Static[HttpResponse]] = maxSize(maxResponseSize, staticBody(true))
+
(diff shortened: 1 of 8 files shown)


colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
```diff
@@ -22,24 +22,34 @@ object ResponseResult {
 object HttpResponseParser  {
   val DefaultMaxSize: DataSize = 10.MB
 
+  def static(maxResponseSize: DataSize = DefaultMaxSize): Parser[DecodedResult.Static[HttpResponse]] = maxSize(maxResponseSize, staticBody(true))
+
+  def stream(dechunkBody: Boolean): Parser[DecodedResult[StreamingHttpResponse]] = streamBody(dechunkBody)
+
+
   //TODO: eliminate duplicate code
   //TODO: Dechunk on static
```
nsauro: I'm pretty sure this comment is invalid.  I think we are already dechunking static responses.

nsauro:
Looks good.

Issue #37, opened 2014-12-02
Fix ServiceClient request timeouts


DanSimon:
Right now service clients don't fully handle timing out requests.  Part of the issue is that it's not very clear how we should react to sent requests that have timed out.  Most likely the only solution will be to kill the connection, but it's not clear if we should treat that the same way we treat unexpected disconnects.

PR #149, opened 2015-03-12
Implement timeout checking for queued messages in controller

reeve:
Couple of significant API changes:
1) Add timeout field to ControllerConfig
2) Add optional createdMillis field to OutputController.push

I'm going to do the in-flight timeout checks in another PR.

@DanSimon @nsauro


nsauro:
There is no ticket for this?

reeve:
#37 

Sorry forgot to link

colossus/src/main/scala/colossus/controller/Controller.scala
```diff
@@ -16,7 +18,8 @@ trait StreamMessage {
 }
 
 case class ControllerConfig(
-  outputBufferSize: Int
+  outputBufferSize: Int,
```
nsauro: We should really doc this class.  It's not immediately obvious what these values do.

colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -110,7 +112,7 @@ extends Controller[I,O](codec, ControllerConfig(50)) {
       requests.hit(tags = tags)
       latency.add(tags = tags, value = (System.currentTimeMillis - done.creationTime).toInt)
       concurrentRequests.decrement()
-      push(comp) {
+      push(comp, done.creationTime) {
```
nsauro: Blegh..annoying that I just can't comment on unaffected lines in a PR.  This comment is not directed to this line.  It is for the comment above this function which is slightly out of date.  The comment states: 

/**
   * This is the only function that actually writes to the channel.  It will
   * write any queued responses until it hits an incomplete promise (or the
   * write buffer fills up)
   */
Which isn't wholly true anymore..this function actually will just enqueue the finished requests/responses.  When they get written is up to the controller

colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -89,11 +88,14 @@ extends Controller[I,O](codec, ControllerConfig(50)) {
   private val requestBuffer = collection.mutable.Queue[SyncPromise]()
   private var numRequests = 0
 
-  def idleCheck(period: Duration) {
+  override def idleCheck(period: Duration) {
+    super.idleCheck(period)
```
nsauro: Let's see if I can say this intelligently, or if this is even an intelligent comment:

  I've never been a fan of having code that get's structured in a way that if a function is overridden and the super isn't invoked..Bad Things Happen.  If the super's function absolutely needs to run(as in this case), I'm more a fan of making it so that you cannot override that function, but provide a hook which gets called at the end of the function.  That way children classes can override that function, add their own functionality and not accidentally fubar stuff.

Not saying we should change this..but def something I think we should strive to avoid if at all possible.

colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -295,7 +297,8 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize)) with
   }
 
 
-  def idleCheck(period: Duration) {
-    //TODO: timeout pending requests
+  override def idleCheck(period: Duration) {
```
nsauro: Can't we just remove this?

colossus/src/main/scala/colossus/controller/OutputController.scala
```diff
@@ -285,4 +291,13 @@ trait OutputController[Input, Output] extends MasterController[Input, Output] {
     }
   }
 
+  def idleCheck(period: Duration): Unit = {
```
nsauro: Possibly a totally inane comment, but figured I'd let it rip.

I just want to make sure we didn't introduce a race condition. Before, the only function that was mutating the 'waitingToSend' queue was 'checkQueue'.  The scenario I was dreaming up was where both checkQueue and idleCheck were invoked around the same time, and this function says "yep..the head is timed out", and before it pops, checkQueue rushes in and pops the queue.  Then this pops the queue.  Oops.  We would have just popped a non timed out message.

However, after tracking down the invocations of checkeQueue(and its callers callers), it seems like they are all bound by the Worker, meaning that they are called only when the Worker initiates the call chain after it receives a message.  Since that is threadsafe, we are safe here as well.  

I think that covers it, but @DanSimon there are no other places where these functions get invoked, right?

nsauro:
:+1: Most of my comments were rambles or unrelated.  I think my only real valid comment here is that I would prefer the commits squashed into a single commit, just to keep the history clean.

colossus/src/main/scala/colossus/controller/OutputController.scala
reeve: I can certainly sync the access (I don't think there's a better way of making it threadsafe) but my understanding was it wasn't needed here.

colossus/src/main/scala/colossus/service/ServiceClient.scala
reeve: Yup.

colossus/src/main/scala/colossus/service/ServiceServer.scala
reeve: I totally agree, in this case it's a result of the idleCheck being declared on ConnectionHandler which is inherited all down the stack (and the fact that the ServiceClient/Controller hierarchy is inheritance based). Scala makes it extra fruity by allowing multiple inheritance so the end result depends on declaration order. To do what you suggest I think we'd need to repeat that hook pattern in a few places, it could get really messy.

Would be interested in @DanSimon 's views on this :)

reeve:
@nsauro maybe you can show me the best way to squish after the fact tomorrow :)

colossus/src/main/scala/colossus/controller/OutputController.scala
DanSimon: Yeah these methods are only invoked by the worker and are thus threadsafe.  In fact, there is no code in this class or any other class inheriting `ConnectionHandler` that is invoked anywhere outside the worker.

colossus/src/main/scala/colossus/service/ServiceServer.scala
DanSimon: Yeah I realize that "hoping" the parent method gets invoked leaves some room for error, but I think starting to put hooks everywhere might end up being just as error prone.  At least we're only doing it in a few places and it's something that's easily testable.

colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -294,8 +296,4 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize)) with
     }
   }
 
-
-  def idleCheck(period: Duration) {
```
DanSimon: So there's still the issue that we're not going to be timing out any sent requests, as that check is definitely up to service layer.  However I don't think we've fully fleshed out what we want the behavior to be when that happens, so we can push that off to another issue.

DanSimon:
:+1: 

The easiest way to squash is to `git rebase -i master`, and except for the first commit change all of the "pick" to "fix" or just "f".  Then do a `git push -f`.  Travis will detect it and rerun so you'll know if everything worked out.  Alternatively you can `git merge --squash` into a new branch, but I only do that if the rebase has merge conflicts.

Also I noticed this PR is off of master.  I think we had originally planned this to go in 0.6.1, but we can just roll it into 0.6.0-RC3 which will be going out friday.

PR #150, opened 2015-03-13
Timeout requests in flight

reeve:
Added an idleCheck to ServiceClient which looks at the head of the sent buffer, and if that request has expired, kills the connection. 

@DanSimon can you confirm this is the right way to close the connection? I found a number of possibilities but this seemed best

@nsauro PR to master as Dan wants to put this stuff on 0.6.0 RC3

### DIFF
--- a/colossus/src/main/scala/colossus/service/ServiceClient.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceClient.scala
@@ -306,7 +306,7 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
 
     if (sentBuffer.size > 0 && sentBuffer.front.isTimedOut(System.currentTimeMillis)) {
       // the oldest sent message has expired with no response - shutdown the connection
-
+      disconnect()
     }
   }
 }
--- a/colossus/src/main/scala/colossus/service/ServiceClient.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceClient.scala
@@ -123,8 +123,12 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
   private val latency = worker.metrics.getOrAdd(Histogram(name / "latency", periods = List(60.seconds), sampleRate = 0.10, percentiles = List(0.75,0.99)))
   lazy val log = Logging(worker.system.actorSystem, s"client:$address")
 
+  private val responseTimeoutMillis: Long = config.requestTimeout.toMillis
+
   case class SourcedRequest(message: I, handler: ResponseHandler) {
     val start: Long = System.currentTimeMillis
+
+    def isTimedOut(now: Long) = now > (start + responseTimeoutMillis)
   }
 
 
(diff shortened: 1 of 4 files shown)


colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -296,4 +300,13 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
     }
   }
 
+
+  override def idleCheck(period: Duration) {
+    super.idleCheck(period)
+
+    if (sentBuffer.size > 0 && sentBuffer.front.isTimedOut(System.currentTimeMillis)) {
+      // the oldest sent message has expired with no response - shutdown the connection
+      disconnect()
```
DanSimon: I think we need to think a bit more about if this is the right thing to do.  In particular, do we want the client to attempt to reconnect?  I think we do want to reconnect, so then we don't want to just call `disconnect()` because that treats it like the user is consciously deciding to close the connection and no reconnect is attempted.

Instead we might want to handle this more like an error, in which case we should have the client basically go through the same logic as if `connectionLost` is called.

Also at the very least we should be more vocal about this happening.  Just calling disconnect will close the connection silently without any indication as to why.

reeve: Interesting re: reconnecting - when I was testing I had to explicitly disable reconnection otherwise it did seem to do it. 
The communication aspect is valid - how about I fail that timed out request (and any others) with a timeout error and then close?

DanSimon: That first thing might be a bug.  Calling disconnect() should always be a "final" operation as it is intended to only be used when the user actually wants to shut things down.  I will check on this and maybe write some more tests.

So if we have this go through the same process that `connectionLost` goes though, the client will deal with all pending and in-flight requests correctly, since what happens depends on whether `failFast` is enabled or not.

DanSimon: Ok there is a bug with `connectionStatus` returning the wrong value, but the client isn't actually trying to reconnect.

I'm just going to push the fix right into this branch because it's really minor and I also made a couple minor edits to your test.

reeve:
Looks great - thanks @DanSimon!

DanSimon:
So this definitely needs some more thought.  In particular, when the timeout occurs, we essentially want to treat this as if the connection was lost, so that the client properly cancels outstanding requests and possibly retries to connect.  Right now calling disconnect() won't do any of that, and manually calling `connectionLost`is also probably not the way to go.

I'm going to bump this to 0.6.1 so we have some more time to iron this out.

DanSimon:
Well after a month of deliberation and procrastination, I finally figured out how to solve the outstanding issue with this PR.  Basically I created a new `Kill` message that the client can send to the worker to force a connection disruption.  Even though the client is the one actually closing the connection, it reacts to it exactly as it would if the connection was just severed, including metrics and logs.

PR #151, opened 2015-03-13
setting ServiceServer output timeout to infinite

DanSimon:
I just realized for servers we never want to timeout anything in the output buffer, since we cannot just drop a response if the request has timed out, otherwise the client will end up getting the wrong responses for requests.


nsauro:
ha..that's a nasty lil bug. :+1:

DanSimon:
Yeah, I'm almost thinking we might want to totally remove the buffer from `OutputController`, since I think it is very application specific how timeouts are handled, and anything pushing into the output controller will have to deal with the buffer being full anyway, so the buffer doesn't allow us to ignore backpressure.  So it might be better to just remove the extra complexity.  Not sure on that though, certainly out of scope for this PR.

PR #159, opened 2015-03-23
EventCollectors now registering with MetricDatabase

nsauro:


colossus-metrics/src/main/scala/colossus/metrics/MetricClock.scala
```diff
@@ -35,6 +55,33 @@ class MetricDatabase(systemId: MetricSystemId, namespace: MetricAddress, snapsho
...
+      }else{
+        log.debug(s"Registered EventCollector $ref")
+        collectors.add(ref)
+      }
+    }
+    case Terminated(child) => {
+      log.warning(s"oh no!  We lost an EventCollector $child. Removing from registered collectors.")
+      collectors.remove(child)
+      //weird condition here:  we get a tock of data, it is collected.  The actor is terminated, we remove it from our collectors
+      //Now, when we report percentage complete, we will have x reported and x-1 collectors, giving us > 100%!
```
DanSimon: Probably the way to fix this is to only remove actors from the hashset when the tick occurs, but I think that's a pretty minor edge case.

DanSimon:
:+1:  feel free to merge once travis finishes up.

PR #160, opened 2015-03-23
Simplify locality

### LINKED ISSUE #153: Simply the locality stuff, remove .shared on event collectors
Right now every event collector (rate, histogram, etc) has a `.shared`method to get a shared version of itself.  This is no longer needed since it was originally intended to be used when jumping back and forth between callbacks and futures, which we've determined is not a great idea.

DanSimon:
Fixes #153 

`Locality` no longer has as type-param.  It used to need it in order to support the ability to turn a `E with LocaliLocality[E]` into a `E with SharedLocality[E]`, so Locality itself needed to know the type of event collector it was working with.  Now that we've removed the need for this, it can become just a regular trait.  Creation of shared collectors is now handled by the `Generator` typeclass.

This also allowed me to cleanup a little how `SharedCollection` is created, since previously all local collectors needed access to the shared collector actor, since even though they never used it themselves they needed the reference when using the `.shared` method.  Now that this method has been removed, the whole process has become more streamlined.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
@@ -15,21 +15,17 @@ trait MetricEvent {
 
 case class MetricSystemConfig(tickPeriod: FiniteDuration = 1.second)
 
-sealed trait Locality[+T] {
-  def shared : T with SharedLocality[T]
-}
-trait SharedLocality[+T] extends Locality[T] {self: T =>
-  def shared : T with SharedLocality[T] = self
-}
-trait LocalLocality[+T] extends Locality[T] with MetricProducer {
+sealed trait Locality
+trait SharedLocality extends Locality
+trait LocalLocality  extends Locality with MetricProducer {
   def event: PartialFunction[MetricEvent, Unit]
 }
   
 
 trait EventLocality[E <: EventCollector]
 object EventLocality {
-  type Shared[E <: EventCollector] = E with SharedLocality[E]
-  type Local[E <: EventCollector] = E with LocalLocality[E]
+  type Shared[E <: EventCollector] = E with SharedLocality
+  type Local[E <: EventCollector] = E with LocalLocality
 }
 import EventLocality._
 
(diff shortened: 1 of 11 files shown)


colossus-metrics/src/main/scala/colossus/metrics/collectors/Rate.scala
```diff
@@ -21,16 +21,13 @@ object Rate {
 
   def apply(address: MetricAddress, periods: List[FiniteDuration] = List(1.second, 60.seconds)): RateParams = RateParams(address, periods)
 
-  //this appears to be obselete now, AFAICT shared rates will only be created by called .shared on a local rate
-  implicit object SharedRate extends Generator[SharedLocality,Rate, RateParams] {
-    def apply(params: RateParams)(implicit collector: ActorRef): Shared[Rate] = {
+  implicit object LocalRate extends Generator[Rate, RateParams] {
```
nsauro: Is LocalRate a valid name for this Generator?  Seems like it generates both local and shared.  Should it just be RateGenerator, or something along those line?

colossus-metrics/src/main/scala/colossus/metrics/collectors/Counter.scala
```diff
@@ -20,8 +20,9 @@ object Counter {
 
   def apply(address: MetricAddress) = CounterParams(address)
 
-  implicit object LocalCounterGenerator extends Generator[LocalLocality, Counter, CounterParams] {
-    def apply(params: CounterParams)(implicit collector: ActorRef) = new LocalCounter(params, collector)
+  implicit object LocalCounterGenerator extends Generator[Counter, CounterParams] {
```
nsauro: Same with this..seems like the name should match the other generators, and just be CounterGenerator

colossus-metrics/src/main/scala/colossus/metrics/collectors/Rate.scala
DanSimon: ah yeah it should, that's what the other have.  That was more remnants from an earlier era of this stuff

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -197,11 +192,16 @@ object LocalCollection {
  * events just as function calls), this should not be used for very-high
  * frequency events.
  */
-class SharedCollection(val local: LocalCollection) extends Collection[SharedLocality] {
+private[colossus] class SharedCollection(val local: LocalCollection, val collector: ActorRef) extends Collection[SharedLocality] {
```
nsauro: Is this private modifier here just to prevent external application code from invoking this constructor?

nsauro:
looks good to me. :+1:

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
DanSimon: yeah, outsiders should never use that constructor.

PR #165, opened 2015-03-25
Metrics global intervals

nsauro:


DanSimon:
Looks like there's some timing issues in the new tests, travis seems to be failing with a fairly high probability.

DanSimon:
So it looks like there's a pretty big component this doesn't do yet, which might actually be ok until I get my branch merged in, but we need the event collectors that track multiple intervals (rate, PeriodicHistogram) to initialize with the global intervals instead of user-supplied ones, and then when the Ticks come in, they only report the metrics for the correct interval.  Right now A Periodic Histogram is going to end up reporting both its 1 second and 60 second aggregation in all intervals, so we're just duplicating our data.

Also in those cases the event collectors are only expecting one global tick, not one tick per interval, so I think its going to require some fairly significant changes to those classes.

PR #167, opened 2015-03-27
Metrics monoidization

DanSimon:
The values in a `MetricMap` now implement a `MetricValue`trait, with a defined `+` operation.  This means a whole bunch of things:

A lot of automatic aggregation can now take place when collecting metrics from event collectors.  Previously, if two collectors were recording the same metric (such as request rate in Colossus recorded by each event loop), the aggregator had no way to know how to properly aggregate multiple values for the same metrics.  So previously we were just adding worker-id tags to every metric to avoid any collisions, forcing users to manually squash these metrics in filters.  Now since the aggregation method is encoded in the value, we need not be afraid of collisions.

In particular, when writing filters to report metrics, whereas previously a filter *had* to look like:

```
MetricFilter("SELECT /my-service/latency WHERE period=60 GROUP BY mytag1, mytag2 MAX")
```
it can now become
```
MetricFilter("/my-service/latency")
```
requiring no special selection or aggregation when you simply want to report metrics with the tags intact.

Furthermore, when aggregating histogram percentiles (and the soon-to-once-again-exist rate moving averages), the aggregation can actually generate the correct weighted average, rather then selecting the max or unweighted average (although that is still available).

I also removed a bunch of unused code.  Also, once all the outstanding branches are merged we'll do a code-cleanup pass.


colossus-metrics/src/main/scala/colossus/metrics/Aggregation.scala
```diff
@@ -38,20 +38,23 @@ object TagSelector {
...
+  def aggregate(values: Seq[MetricValue]): RawMetricValue
 }
 object AggregationType {
   case object Sum extends AggregationType {
-    def aggregate(values: Seq[Long]) = values.sum
+    def aggregate(values: Seq[MetricValue]): RawMetricValue = values.foldLeft(0L){case (build: RawMetricValue, next: MetricValue) => build + next.value}
   }
   case object Average extends AggregationType {
-    def aggregate(values: Seq[Long]) = values.sum / values.size
+    def aggregate(values: Seq[MetricValue]) = values.foldLeft(0L){case (build, next) => build + next.value} / values.size
```
nsauro: Are we sure we won't get  divide by zero exceptions here?  Should we have an empty check on this list?

colossus-metrics/src/main/scala/colossus/metrics/Aggregation.scala
```diff
@@ -38,20 +38,23 @@ object TagSelector {
...
   }
   case object Max extends AggregationType {
-    def aggregate(values: Seq[Long]) = values.max
+    def aggregate(values: Seq[MetricValue]) = values.foldLeft(0L){case (max, next) => if (max > next.value) max else next.value}
```
nsauro: For both this and Min...if we are iterating over a set of negative values, wouldn't starting at 0 cause some issues?
Should we use reduceLeft instead, and use an empty check to return a default value?

colossus-metrics/src/main/scala/colossus/metrics/Aggregation.scala
```diff
@@ -38,20 +38,23 @@ object TagSelector {
...
   }
   case object Min extends AggregationType {
-    def aggregate(values: Seq[Long]) = values.min
+    def aggregate(values: Seq[MetricValue]) = values.foldLeft(Long.MaxValue){case (min, next) => if (min < next.value) min else next.value}
+  }
+  case object Natural extends AggregationType {
+    def aggregate(values: Seq[MetricValue]) = values.reduce{_ + _}.value
```
nsauro: FWIW:  Intellij just chirped that we could use a 'sum' function here. <shrug>

colossus-metrics/src/main/scala/colossus/metrics/Aggregation.scala
DanSimon: Yeah there should be a check, though apparently it's never actually happened because it's been like that since forever and I've never seen an exception thrown.

colossus-metrics/src/main/scala/colossus/metrics/Aggregation.scala
DanSimon: hmmm, yes on both

colossus-metrics/src/main/scala/colossus/metrics/collectors/Gauge.scala
```diff
@@ -76,8 +76,9 @@ class ConcreteGauge(params: GaugeParams) extends Gauge with LocalLocality with T
     }
   }
 
+  //TODO: how gauges are aggregated should be configurable, since it's not obvious which should be the default
```
nsauro: Yea..I can't think of a better way of doing it..and I think making it configurable is the only real option.

colossus-metrics/src/main/scala/colossus/metrics/Aggregation.scala
DanSimon: must only be looking at the plus sign, compiler complains about no implicit `Numeric` when I try sum, just gonna leave it as is.

nsauro:
:100:   Looks good to me.

colossus-metrics/src/main/scala/colossus/metrics/Aggregation.scala
DanSimon: Actually now that I'm thinking about, the way the aggregation works these will never get called if the lists are empty.  Basically aggregation happens while looping over value maps (`Map[TagMap, MetricValue]`) so either the seq will be non-empty or non-existent.  I'll still fix the issue for negative values though

DanSimon:
Ah nuts I just made all my fixes in another branch which branches off of this branch.  So I'll just merge this in now as is and get those fixes in with my next PR.

DanSimon:
Also gonna squash merge this, so closing here

PR #174, opened 2015-03-31
Rates, histograms use global metrics

### LINKED ISSUE #168: Ticked metrics must use global intervals
Currently rates and periodic histograms still directly configure their aggregation intervals from constructor parameters.  

1.  They should use the global intervals defined for the metric system
2.  For each tick, they should only emit the metrics for that tick's interval.  Right now a rate configured for 1 second and 1 minute aggregations will emit both aggregations regardless of which interval aggregator asked for them.

DanSimon:
Fixes #168 , This makes 2 changes

1.  Whereas rates and histograms previously took in aggregation intervals as constructor parameters, these are now automatically supplied by metric system.
2.  When ticks come in from interval aggregators, these collectors only returns metrics intended for the interval of the tick.  So a histogram collecting data for both 1 second and 1 minute intervals will only report the 1-second data for the 1-second tick.

I also have some of the fixes for things mentions in the previous PR.


nsauro:
+1

PR #177, opened 2015-04-01
MetricReporters now use MetricReporterFilter to specify what stats to re...

### LINKED ISSUE #175: Do away with querying in MetricReporters
Now that we don't need metric filters to squash or combine metrics in Reporters, we should switch MetricReporters to indicating what metrics to report in either a white list, black list.

nsauro:
...port, no longer uses a Seq[MetricFilter].  Also, removed unused classes.

fixes #175

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/StatReporter.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/StatReporter.scala
@@ -65,7 +65,6 @@ class MetricReporter(intervalAggregator : ActorRef, config: MetricReporterConfig
   private def filterMetrics(m : MetricMap) : RawMetricMap = {
     filters match {
       case MetricReporterFilter.All => m.toRawMetrics
-      case MetricReporterFilter.None => Map()
       case MetricReporterFilter.WhiteList(x) => {
         m.filterKeys(x.contains).toRawMetrics
       }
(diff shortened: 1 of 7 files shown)


colossus-metrics/src/main/scala/colossus/metrics/StatReporter.scala
```diff
@@ -64,6 +62,19 @@ class MetricReporter(intervalAggregator : ActorRef, config: MetricReporterConfig
     }
   }
 
+  private def filterMetrics(m : MetricMap) : RawMetricMap = {
+    filters match {
+      case MetricReporterFilter.All => m.toRawMetrics
+      case MetricReporterFilter.None => Map()
```
DanSimon: Will anyone actually ever use None?  Seems kinda weird if someone made a reporter and didn't report anything.

nsauro: Yea..i I figured as much..seems totally weird.  I just put it in in the even that they just wanted to quickly silence metric reporting.  I guess they could just comment out that line.  I can remove that case.

DanSimon:
Aside from my one point, this looks good

nsauro:
Removed the None option.,  Will merge this once travis greenlights it

PR #178, opened 2015-04-01
Connection info now available to services, fixiing initialization

DanSimon:
Two main things in here:

1.  There's a new ConnectionInfo trait that contains a bunch of low-level read-only values which is now exposed to all layers of a server connection handler.  There was already a `ConnectionInfo` class, which I renamed to `ConnectionSnapshot`which is more descriptive of what it is.

2.  Fixed a bug where basically the connection handler initialization in the service DSL was occurring before the connection handler actually was fully connected.  So if someone tried to grab the connection info it would have come back as None.  So now the initializer is invoked through the handler's connected method.


nsauro:
:+1:

PR #181, opened 2015-04-06
new TestSharedCollection, adding some more tests

DanSimon:
This is the beginnings of a simple testkit for colossus-metrics.  Don't think we'll need to break it into a separate project.

Also wrote some more tests.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/TestKit.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/TestKit.scala
@@ -1,0 +1,16 @@
+package colossus.metrics
+package testkit
+
+import akka.testkit.TestProbe
+import scala.reflect.ClassTag
+
+
+/**
+ * A Fake shared collection for creating event collectors using a provided test probe
+ */
+class TestSharedCollection(probe: TestProbe) extends Collection[SharedLocality] {
+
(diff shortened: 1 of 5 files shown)


colossus-metrics/src/main/scala/colossus/metrics/TestKit.scala
```diff
@@ -0,0 +1,16 @@
...
+package testkit
+
+import akka.testkit.TestProbe
+import scala.reflect.ClassTag
+
+
+/**
+ * A Fake shared collection for creating event collectors using a provided test probe
+ */
+class TestSharedCollection(probe: TestProbe) extends Collection[SharedLocality] {
```
nsauro: Weren't we going to add the DeadMetricSystem here?

nsauro:
:1

nsauro:
after this is merged, I will merge in my branch with its single commit for modding the access modifier on SharedCollection

colossus-metrics/src/main/scala/colossus/metrics/TestKit.scala
DanSimon: Right now there actually is no class or object, just a method on `MetricSystem`, so I'll leave it be for now

PR #182, opened 2015-04-06
few more tweaks for RC1

nsauro:
I"ll delete this branch after merging.


colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -76,6 +76,8 @@ case class HttpResponse(head: HttpResponseHead, body: Option[ByteString]) extend
 
   def withHeader(key: String, value: String) = copy(head = head.withHeader(key,value))
 
+  val code = head.code
```
DanSimon: This should be a def, otherwise we're storing it twice

nsauro:
fixed

DanSimon:
:+1:

PR #185, opened 2015-04-13
Fixing `colossus.protocol.redis.Commands.Setex` to take expiration

dangermike:
`Setex` was busted. It should work now.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/redis/Command.scala
+++ b/colossus/src/main/scala/colossus/protocols/redis/Command.scala
@@ -2,6 +2,7 @@ package colossus
 package protocols.redis
 
 import akka.util.{ByteString, ByteStringBuilder}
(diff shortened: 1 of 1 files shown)


nsauro:
Do we want to include a default value?

DanSimon:
I don't think so, you only ever want to call setex when you want to TTL something, so there isn't really any default that makes sense.

nsauro:
Ha..wow..yea that makes no sense.  I should have checked this command's definition.

DanSimon:
The only thing about using a `FiniteDuration` here is the fact that you can define one like `345.milliseconds` which is just gonna get turned into 0 seconds.  But I think it's not a big deal since the redis documentation itself is pretty clear that the ttl is always in seconds.  So LGTM :+1:

PR #192, opened 2015-04-27
New error handler in service DSL

### LINKED ISSUE #191: Allow services to override error handler
For requests that fail at some point during processing, the `handleError` method is invoked.  Normally this only occurs if an uncaught exception is thrown while processing a request, but it can also happen when a request times out or a few other situations.

Right now every codec has a method that basically converts any exception into an error message, but we should provide a hook in the service DSL to allow users to override this functionality for a service.

DanSimon:
Fixes #191 

This allows services to inject their own error formatting logic into services when a non-recoverable error occurs.  Any errors that fall through the provided `PartialFunction` will be handled by the default handler provided by the coded.


nsauro:
:100:

PR #193, opened 2015-04-28
Server graceful shutdown

### LINKED ISSUE #77: Graceful Shutdown for Servers
Right now there's really no way to gracefully shutdown a Server.  We can close individual connections gracefully, but we should be able to have Servers stop listening for new connections while properly waiting for open connections to either close on their own or timeout.

DanSimon:
Fixes #77 .

Servers now have a graceful shutdown process that involves two major changes:
* All server connection handlers now have a new `shutdownRequest()` method, which is called on every connection from a server when the server initiates its shutdown process.  For services, this just calls `gracefulDisconnect`.
* The server has a new `ShuttingDown` state, which it enters when it receives the `Shutdown` message.  this has the server broadcast the shutdown request to every worker, and the server then closes its socket listener and waits for all open connections to close.  There is also a new `shutdownTimeout` in `ServerSettings` that dictates how long a server should stay in this state until it times out and just shuts down (which also force-closes any connections still open)


colossus/src/main/scala/colossus/core/Worker.scala
```diff
@@ -211,6 +211,11 @@ private[colossus] class Worker(config: WorkerConfig) extends Actor with ActorMet
     case WorkerManager.UnregisterServer(server) => {
       unregisterServer(server.server)
     }
+    case WorkerManager.ServerShutdownRequest(server) => {
+      connections.collect{ case (id, connection: ServerConnection) if (connection.server == server) => {
```
nsauro: why not a foreach?  Not that it matters at all.

nsauro:
+1. Looks good.

PR #196, opened 2015-05-04
Document site fix

wolfapple:
1. Your Gemfile has no gem server sources. So I added a line like this to your Gemfile:
source 'https://rubygems.org'

2. Fixed syntax error in string interpolation ($ -> s)


DanSimon:
Looks good, thanks!

PR #197, opened 2015-05-06
rates, histograms can prune empty values

### LINKED ISSUE #187: Option on Rates/Histograms to filter out empty tags
If a tag has no values in it, there should be an option to not report it.

DanSimon:
Fixes #187 

Rates and Histograms have a new `PruneEmpty` option that when true, if a rate/histogram value for a set of tags received no hits during an interval, the value will be totally removed instead of being reported as zero.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
@@ -14,7 +14,8 @@ case class HistogramParams (
   address: MetricAddress, 
   bucketRanges: BucketList = Histogram.defaultBucketRanges, 
   percentiles: List[Double] = Histogram.defaultPercentiles,
-  sampleRate: Double = 1.0
+  sampleRate: Double = 1.0,
+  pruneEmpty: Boolean = false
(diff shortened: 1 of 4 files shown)


colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
```diff
@@ -182,8 +184,18 @@ class TaggedHistogram(val ranges: BucketList, percs: List[Double]) {
...
-    lastFullSnapshot = hists.map{case (tags, hist) => (tags, hist.snapshot(percs))}.toMap
-    reset()
+    val toRemove = collection.mutable.ArrayBuffer[TagMap]()
+    val snapBuild = collection.mutable.Map[TagMap, Snapshot]()
+    hists.foreach{case (tags, hist) => 
+      if (hist.count == 0 && pruneEmpty) {
+        toRemove += tags
+      } else {
+        snapBuild += (tags -> hist.snapshot(percs))
+        hist.reset()
```
nsauro: Do we want to invoke reset on hist, regardless of its prunability?

nsauro:
+1.  Looks gooe

PR #200, opened 2015-05-12
new repeatEOS parser, upping TestClient timeouts

DanSimon:
@dangermike 

This adds a new parser that will repeat another parser forever until the end of stream is reached.  

Also upping some timeouts in the tests.


nsauro:
oops.  My bad, I thought we released 0.6.3

nsauro:
+1

Issue #201, opened 2015-05-21
message from exception included in metrics tags on request errors


DanSimon:
in Colossus' built-in errors metric, we add the type of error as a metric tag, but it looks like it's including both the exception classname and the message body, which is way too much for a tag.  We should change it to only include the exception name:

This likely just involves overloading toString on some of the DisconnectCause case classes.

PR #204, opened 2015-06-02
improving error propagation, fixing bug with disabled writes

DanSimon:
This PR has several bug fixes and improvements

* Fixing a bug where if a client loses its connection after it disables writes, when the connection is re-established writes are never re-enabled, and the client refuses to send any requests.
* Improving the propagation of client connection errors to pending and outstanding requests.  Before if a service client lost its connection, any pending and outstanding requests would be completed with a generic error.  Now the actual error that caused the connection to be lost is included.
* Fixing #201 .  Now every `DisconnectCause` has both a name for the tag and a human-readable string
* Increasing the default sent buffer size for service-clients.  The value is still somewhat arbitrary but 20 was definitely too low and probably leading to unnecessary bottlenecking

### DIFF
--- a/colossus/src/main/scala/colossus/service/ServiceClient.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceClient.scala
@@ -39,7 +39,7 @@ case class ClientConfig(
   requestTimeout: Duration, 
   name: MetricAddress,
   pendingBufferSize: Int = 100,
-  sentBufferSize: Int = 20,
+  sentBufferSize: Int = 100,
   failFast: Boolean = false,
   connectionAttempts : PollingDuration = PollingDuration(250.milliseconds, None),
   idleTimeout: Duration = Duration.Inf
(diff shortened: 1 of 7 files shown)


colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -216,30 +214,30 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
     connectionAttempts = 0
   }
 
-  private def purgeBuffers(pendingException : => Throwable) {
+  private def purgeBuffers(reason : Throwable) {
     sentBuffer.foreach{s => 
       errors.hit(tags = hpTags)
```
nsauro: do we want tagstring here?

DanSimon: Maybe, I'm not sure if we should use the `DisconnectCause` that caused the purge (which as of now isn't passed into the function) or just the error itself.  Needs some more thought so I'll hold off on it for how.

PR #205, opened 2015-06-02
Metric Clock uses scheduleOnce for ticks instead of schedule

### LINKED ISSUE #202: Switch metric-clock ticking to use scheduleOnce
There definitely seems to be some weirdness around using `scheduler.schedule` where we have collectors getting old or possibly out-of-order messages.  using scheduleOnce instead should eliminate those issues.

DanSimon:
Fixes #202 .  

This should fix some weirdness we've been seeing where tick messages pile up and lead to race conditions.  By using `scheduleOnce`, we can now guarantee that only ever one message is schedule to be sent at a time.


nsauro:
+1

PR #213, opened 2015-06-12
Callback to future with failure handling

zgagnon:
Previously, callbacks were converted to futures by mapping a promise completion. This would not complete if the callback failed. Now promise is completed using both a map and a recover, allowing failed callbacks to be converted into failing futures.


colossus-tests/src/test/scala/colossus/service/CallbackSpec.scala
```diff
@@ -310,6 +310,23 @@ class CallbackSpec extends ColossusSpec {
...
+      value must equal(5)
+    }
+
+    "toFuture failed" in {
+      import scala.concurrent.ExecutionContext.Implicits.global
+
+      val badfunc = {f: (Try[Int] => Unit) => f(Failure[Int](new Exception))}
+      val callback = Callback(badfunc).map { i => i }
+      val result = Await.ready(callback.toFuture, 2 seconds).value.get
+      result mustBe a [Failure[Int]]
```
nsauro: This is throwing a compiler warning because of type erasure

DanSimon: yeah I've run into the same problem before, better to do `result.isFailure must equal(true)`

colossus/src/main/scala/colossus/service/Callback.scala
```diff
@@ -1,4 +1,4 @@
-package colossus
+jpackage colossus
```
nsauro: compiler error

colossus-tests/src/test/scala/colossus/service/CallbackSpec.scala
zgagnon: coolbeans will do.

DanSimon:
:+1:

nsauro:
looks good

PR #214, opened 2015-06-13
Removing broken Executor from testkit, new TestExecutor (that works)

DanSimon:
So after some careful examination, it's clear that the calling-thread callback executor did not properly work.  Furthermore, the Callback Matchers do not correctly handle Callbacks that work with Futures.

Going forward, the new `CallbackAwait` should be the only method to extract the value from a Callback in tests.  I've kept the Callback Matchers for now for backwards compatibility (and they do work with vanilla Callbacks), but i think we'll have to remove them.

Basically it works by spinning up an actual actor in a PinnedDispatcher that executes the callbacks and acts as the `CallbackExecutor`.  This ensures proper behavior when a Callback "jumps" out of thread, executes as a Future, and jumps back in thread via the Executor.

### DIFF
--- a/colossus-testkit/src/main/scala/colossus-testkit/FakeIOSystem.scala
+++ b/colossus-testkit/src/main/scala/colossus-testkit/FakeIOSystem.scala
@@ -9,7 +9,8 @@ import akka.actor._
 import akka.testkit.TestProbe
 import akka.testkit.CallingThreadDispatcher
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{Await, ExecutionContext, Future, Promise}
+import scala.concurrent.duration._
 
 object FakeIOSystem {
   def apply()(implicit system: ActorSystem): IOSystem = {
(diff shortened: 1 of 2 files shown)


colossus-testkit/src/main/scala/colossus-testkit/FakeIOSystem.scala
```diff
@@ -29,19 +30,52 @@ object FakeIOSystem {
...
-   */
-  def callingThreadCallbackExecutor(implicit system: ActorSystem): CallbackExecutor = {
-    class FakeExecutor extends Actor with CallbackExecution {
-      def receive = handleCallback
+  //prevents accidentally spinning up a new executor per test
+  private val exCache: collection.mutable.Map[ActorSystem, CallbackExecutor] = collection.mutable.Map()
+
+  def testExecutor(implicit system: ActorSystem): CallbackExecutor = {
+    exCache.get(system).getOrElse {
+      val ref = system.actorOf(Props[GenericExecutor].withDispatcher("server-dispatcher"))
```
nsauro: Is there any case to be made for making a generic-executor-dispatcher, which is pinned?  I know server-dispatcher will probably never change, but not sure if we want to make it pull double duty.  I'm fine with keeping it, but at least wanted to bring up the point

nsauro:
For the Scalatest Matcher, does it make sense to just switch that to using the Callback.Await under the covers, so to maintain backwards compatibility?

nsauro:
+1

colossus-testkit/src/main/scala/colossus-testkit/FakeIOSystem.scala
DanSimon: I doubt it.  You wouldn't be able to do anything that actually required callbacks since this executor is not a real event loop with an NIO selector.

DanSimon:
Re the matcher: yeah I think that's what we should do.  I didn't want to touch it with this PR since I wasn't sure it wouldn't change the semantics of how it works, but we should do that in another PR.

PR #216, opened 2015-06-15
Numerous optimizations and fixes

DanSimon:
This is a pretty large set of changes that result in significantly improved performance, especially with pipelined requests.

* Every connection now allocates its own `ByteBuffer` for writes.  This allows us to optimize such that every connection only performs at most one socket write per event loop iteration.  The buffer size right now is 64KB, so with many connections there is a significant increase in memory but its worth the improvement in throughput.  Going forward we can try to be smarter about using a pool or something.
* There is a new `ConstantCallback` which is now used when `Callback.successful` or `Callback.failed` are used.  The service layer pattern matches on these which in some cases lets it completely skip the request buffering process.
* Fixed a spot in the `InputController` where we were using recursion instead of a regular loop
* Fixed a bug that's been around since day 1 where a selection key with only the `OP_WRITE` interest would not be properly removed from the interest set iterator.  
* Fixed a bug where the connection handler for an `AsyncServiceClient` was attempting to bind to the worker twice.
* Added a new `BenchmarkService` example which is the optimized version of the service that we'll submit to the techempower benchmarks.

There is one other optimization I'm going to do, which will be to allow you to fully disable metrics (which actually has a much larger impact that we've previously thought), but I can do that in another PR.

### DIFF
--- a/colossus-examples/src/main/scala/colossus-examples/BenchmarkExample.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/BenchmarkExample.scala
@@ -1,0 +1,98 @@
+package colossus.examples
+
+
+import java.util.Date
+import java.text.SimpleDateFormat
+
+import colossus._
+import colossus.core.{ServerRef, ServerSettings}
+import service._
+import protocols.http._
+
+import net.liftweb.json._
+import JsonDSL._
+import akka.actor._
+import akka.util.ByteString
+import scala.concurrent.duration._
+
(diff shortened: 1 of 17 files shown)


DanSimon:
I'm going to close and re-open to merge into a different branch off master, and we'll publish this as a release candidate.  All the tests are passing but as these are some of the most intrusive changes ever made to the core layer, I'd like to spend some more time testing before we put this in a stable release.

PR #219, opened 2015-06-16
making worker constructor private, new fakeExecutorWorkerRef

DanSimon:
Having a fake CallbackExecutor is nice, but it's also good to be able to create `WorkerRef` objects that are fake, yet can still execute callbacks.

### DIFF
--- a/colossus/src/main/scala/colossus/core/Worker.scala
+++ b/colossus/src/main/scala/colossus/core/Worker.scala
@@ -37,7 +37,7 @@ case class WorkerConfig(
  * @param worker The ActorRef of the Worker
  * @param system The IOSystem to which this Worker belongs
  */
-case class WorkerRef(id: Int, metrics: LocalCollection, worker: ActorRef, system: IOSystem) {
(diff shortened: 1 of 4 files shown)


colossus/src/main/scala/colossus/core/Worker.scala
```diff
@@ -37,7 +37,7 @@ case class WorkerConfig(
  * @param worker The ActorRef of the Worker
  * @param system The IOSystem to which this Worker belongs
  */
-case class WorkerRef(id: Int, metrics: LocalCollection, worker: ActorRef, system: IOSystem) {
+case class WorkerRef private[colossus](id: Int, metrics: LocalCollection, worker: ActorRef, system: IOSystem) {
```
nsauro: Why is this private now?  Also, if we are going to make this private, what about ServerRef?

DanSimon: It's private since users should not be creating these on their own.  The same should be done for ServerRef, I'll add that in

PR #220, opened 2015-06-18
CallbackMatchers updates

nsauro:
CallbackMatchers using CallbackAwait now;  
Added Implicit conversion for Future=>Callback


DanSimon:
:+1:

PR #221, opened 2015-06-24
making the const parser return its value on endOfStream

dangermike:
Allow parsers that match on endOfStream to play nice with parsers that just end on a parse.


DanSimon:
:+1:

Issue #211, opened 2015-06-11
Callback.sequence with empty sequence never completes


DanSimon:
It should just fill in with `Callback.successful(Nil)`

PR #222, opened 2015-06-30
Executing a callback sequence with an empty sequence succeeds

jeffreythewang:
For issue: https://github.com/tumblr/colossus/issues/211

Fills in with `callback(Success(Nil))`

@DanSimon


DanSimon:
:+1:

PR #225, opened 2015-07-10
Specify metric system without supplying # of workers

aliyakamercan:
Added a new apply method with name & metric system only, # of workers default to available processors.


DanSimon:
:+1:

PR #226, opened 2015-07-13
fixing memory leak with failed connections, other changes

DanSimon:
When a client connection failed to connect, the connection was never removed from the worker's map of live connections.  This was never noticed because the memory overhead was very low, but now that every connection allocates its own buffer it's more apparent.

This fix resulted in some further refactoring about how client connection failures are handled

* adding new `ConnectFailed` `DisconnectError`
* removing `connectionFailed` method from `ClientConnectionHandler`
* minor logging fixups


nsauro:
+1

PR #228, opened 2015-07-13
fixing bug with respecting sentBuffer max size

DanSimon:
Service clients have two configurable buffer sizes:

* `sentBufferSize` - how many sent requests can be currently waiting for a response
* `pendingBufferSize` - how many requests can be waiting to be sent.

The expected behavior is that when the sent buffer maxes out, subsequent calls to `send` don't send the requests but instead add them to the pending buffer.  Once responses for the currently sent requests start coming in, the client dequeues requests from the pending buffer and sends them until the sent buffer fills up again.

However a bug caused the client to totally ignore the max sent buffer size when dequeuing from the pending buffer, so the entire pending buffer is then sent.

this PR fixes that bug and ensures that the sent buffer max size is respected even when draining the pending buffer.


nsauro:
+1

PR #230, opened 2015-07-14
added request formatter to error logging

### LINKED ISSUE #199: printing an HttpRequest prints the entire body
Which is not a bad thing...if the body is small..if you are posting an image, or other binary content..the logs become illegible. 

We should introduce some way of controlling what, on both a Request and Response get printed into a log.

nsauro:
fixes #199


DanSimon:
:+1:

PR #232, opened 2015-07-20
tag request metric with status code (http service handler)

aliyakamercan:
Start tagging requests metric with status code by default.


DanSimon:
:+1:

PR #235, opened 2015-07-27
check if it is a PR before publishing

aliyakamercan:
PR's branch is always master on travis. Make sure it is not a PR before publishing.


DanSimon:
Appears to be working, :+1:

PR #236, opened 2015-07-29
Add more redis commands

benblack86:
More redis commands. Not sure if any of these should be in the "commands we are ok with failing over to a slave" list.


DanSimon:
Travis is failing because there's a bug in our travis config with how we handle our encrypted stuff.  I'll try to get a fix in asap.  Code looks good otherwise.

DanSimon:
@benblack86 try merging master into your branch and re-running travis.  We just merged a fix to our travis config but I think that change has to be on your branch.

benblack86:
Nearly...

DanSimon:
:+1:

PR #237, opened 2015-07-30
don't decrypt keys when not on master or in a PR

DanSimon:
I think this will fix the issue we're seeing with travis runs on #236 .


aliyakamercan:
:+1:

PR #239, opened 2015-08-03
clean up redis/memcached after integration tests

aliyakamercan:
@DanSimon @nsauro


DanSimon:
:+1:

nsauro:
We don't want to do a flushall before/after each?

nsauro:
I would at least make that list a Set if we are going to keep track, no need to wipe dupes(if they occur)

aliyakamercan:
I don't think we want to flush all, cleaning up what is written is sufficient.

nsauro:
Not fer nothin', but most of the delete commands accept multiple parameters.

PR #240, opened 2015-08-05
WriteBuffer - only try to drain PartialBuffer when internal buffer drained

DanSimon:
This fixes a bug that can occur when attempting to write a large chunk of data on a connection in a service under heavy load.

When a user (usually the controller layer) calls `write` on a `WriteBuffer`, the buffer copies as much of the `DataBuffer` as it can to an internal buffer and holds onto the rest.  When the internal buffer is drained into the `SocketChannel`, it reads more data from the partial buffer into the internal buffer.

The bug was that when draining the internal buffer, we incorrectly assumed that it would be drained in one shot, however the `SocketChannel` may not accept the entire buffer, in which case we need to wait another event loop iteration to continue draining the buffer.  This means that the next chunk of the `PartialBuffer` was copied into a non-empty internal buffer, resulting in data corruption.

This fixes that.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/core/WriteBufferSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/core/WriteBufferSpec.scala
@@ -11,7 +11,7 @@ import WriteStatus._
 
 class WriteBufferSpec extends ColossusSpec {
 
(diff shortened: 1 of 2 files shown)


nsauro:
+1

PR #241, opened 2015-08-06
making compiledGlobalTags a function to always get latest tags

DanSimon:
The point of a `TagGenerator` is to generate a set of tags applied to every metric.  These tags are not immutable and may change during the course of an application's life.  Currently though, we only get the tags once when the reporter actor is created, and never fetch them again.  This fixes that.


nsauro:
+1

PR #242, opened 2015-08-10
Service Server Output backpressure

DanSimon:
This fixes a bug where responses could get dropped if a service server connection's output buffer backs up.  Now we don't drop any responses and instead keep the corresponding request/response in the request buffer until the output buffer opens up.


nsauro:
+1

PR #243, opened 2015-08-11
--WIP-- first iteration of typed memcached client

nsauro:
Just a preview of what might be a thing if we end up liking this.  What this branch does is allow us to create ServiceClients specific to each protocol, which have their own protocol specific API.  

This is also incorporates some higher kinded type stuff so unify both callback and future interfaces.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/memcache/package.scala
+++ b/colossus/src/main/scala/colossus/protocols/memcache/package.scala
@@ -1,7 +1,17 @@
 package colossus
 package protocols
 
+import akka.util.ByteString
+import colossus.core.WorkerRef
+import colossus.parsing.DataSize
+import colossus.protocols.memcache.MemcacheCommand._
+import colossus.protocols.memcache.MemcacheReply._
+import colossus.protocols.memcache._
 import service._
+
+import scala.language.higherKinds
+
+import scala.concurrent.{ExecutionContext, Future}
 
 package object memcache {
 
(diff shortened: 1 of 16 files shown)


colossus/src/main/scala/colossus/protocols/memcache/package.scala
```diff
@@ -16,3 +26,150 @@ package object memcache {
   }
 
 }
+
+trait ResponseAdapter[C <: CodecDSL, M[_]] {
```
nsauro: Instead of the usual I, O parameters we use for input/output, I'm using a CodecDSL instead.  There are pros and cons to this, for sure

nsauro:
The big remaining questions on this work are:
- The current typed interfaces are not complete(especially redis, since there are a lot of commands).  Do we want to include functions for all possible redis commands(so many commands!)?
- That question segues into the next: in the current typed interface, there is no way to execute an arbitrary command.  Do we want to expose the 'send' command on each interface?  Or, do we want to just have only a 'hard' interface on these clients, and then provide a  separate generic memcache/redis client (what we have now).  I can see arguments for both cases.  

If we do not provide a general 'send' command on the interface, and opt for the generic interfaces, then we should REALLY complete the redis/memcache typed interfaces(I would argue not completing it is not an option).  

Maybe 'do all the things' is the right answer: provide send function, complete typed interface, provide generic interface?  The only real thing here is..if I'm using a typed client to talk to redis, why would I want to send arbitrary, non standard commands?  Conversely, if I'm using a generic redis client to speak redis protocol to some non redis end point, why would i want the typed interface getting in my way?

colossus/src/main/scala/colossus/service/ResponseAdapter.scala
```diff
@@ -0,0 +1,45 @@
...
+  def flatMap[T](t : M[C#Output])(f : C#Output => M[T]) : M[T]
+
+  def success[T](t : T) : M[T]
+
+  def failure[T](ex : Throwable) : M[T]
+}
+
+trait CallbackResponseAdapter[C <: CodecDSL] extends ResponseAdapter[C, Callback] {
+
+  def client : ServiceClient[C#Input, C#Output]
```
DanSimon: maybe this should be `ServiceClientLike`, though I realize we don't really have the equivalent for an `AsyncServiceClient`.  I definitely think there is room to unify those with a HKT similar to what is done here, but we can save that for another PR.

nsauro:
bump.  this is ready to rock.

DanSimon:
:+1:

PR #245, opened 2015-08-18
fixing a bug where data in internal buffer could be overwritten

DanSimon:
If the internal buffer is in the process of being written out to the socket, and no PartialBuffer is set, a subsequent call to `write` could overwrite the data being written out.

This would occur under the following circumstances:

1. The service is under heavy load 
2. Several request/response message are written to a connection in a single event-loop iteration.  The internal buffer is not filled so the Partial buffer remains empty
3.  When the WriteBuffer tries to write the data to the socket, the socket only accepts part of the write.
4.  On the next event loop iteration, an additional request/response is written to the socket.  Since Partial buffer is empty, the data is copied into the internal buffer, overwriting the data that wasn't fully written out before.


nsauro:
good catch

PR #248, opened 2015-08-19
fixed bug with 'reset' function in BaseHttpServerCodec

### LINKED ISSUE #246: HttpServerCodec's reset function doesn't use maxSize
when the codec gets reset, it is using the default value, not the passed in value.

nsauro:
fixes #246


DanSimon:
:+1:

PR #249, opened 2015-08-24
response headers are bytestrings

DanSimon:
This significantly improves the performance of writing http responses with several headers.  I plan to do the same thing with request headers, but that'll have to wait for another PR to get merged.

This does introduce a breaking API change, but the migration is pretty easy.  Currently to add headers in the construction of a response, you do:

```scala
HttpResponse(headers = Vector(("name" -> "value")), //...
```

now it's just

```scala
HttpResponse(headers = Vector(HttpResponseHeader("name", "value")), //...
``` 

Doing `.withHeader` on a response is unaffected.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -82,9 +82,9 @@ trait HttpHeaderUtils {
    * Be aware that lacking this header may or may not be a valid request,
    * depending if the "transfer-encoding" header is set to "chunked"
    */
-  def contentLength: Option[Int] = headers.collectFirst{case (HttpHeaders.ContentLength, l) => l.toInt}
+  lazy val contentLength: Option[Int] = headers.collectFirst{case (HttpHeaders.ContentLength, l) => l.toInt}
 
-  def transferEncoding : TransferEncoding = singleHeader(HttpHeaders.TransferEncoding).flatMap(TransferEncoding.unapply).getOrElse(TransferEncoding.Identity)
+  lazy val transferEncoding : TransferEncoding = singleHeader(HttpHeaders.TransferEncoding).flatMap(TransferEncoding.unapply).getOrElse(TransferEncoding.Identity)
 }
 
 
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -82,9 +82,9 @@ trait HttpHeaderUtils {
    * Be aware that lacking this header may or may not be a valid request,
    * depending if the "transfer-encoding" header is set to "chunked"
    */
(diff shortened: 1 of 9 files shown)


colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -82,9 +82,9 @@ trait HttpHeaderUtils {
    * Be aware that lacking this header may or may not be a valid request,
    * depending if the "transfer-encoding" header is set to "chunked"
    */
-  val contentLength: Option[Int] = headers.collectFirst{case (HttpHeaders.ContentLength, l) => l.toInt}
+  def contentLength: Option[Int] = headers.collectFirst{case (HttpHeaders.ContentLength, l) => l.toInt}
```
nsauro: couldn't these be lazy vals?

DanSimon: They sure could!  Not sure why I did that further down but not here.

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -10,22 +10,52 @@ import controller._
...
+
+  //TODO: these are bytestrings, whereas in Head they're strings
+  val ContentLength = ByteString("content-length")
+  val TransferEncoding = ByteString("transfer-encoding")
+
+  val DELIM = ByteString(": ")
+
+  def apply(key: String, value: String): HttpResponseHeader = {
+    HttpResponseHeader(ByteString(key) , ByteString(value))
+  }
```
nsauro: Maybe have an unapply here as well?

nsauro: Or, at least a function on the case class to get the values out as Strings

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -82,9 +82,9 @@ trait HttpHeaderUtils {
...
 
-  val transferEncoding : TransferEncoding = singleHeader(HttpHeaders.TransferEncoding).flatMap(TransferEncoding.unapply).getOrElse(TransferEncoding.Identity)
+  def transferEncoding : TransferEncoding = singleHeader(HttpHeaders.TransferEncoding).flatMap(TransferEncoding.unapply).getOrElse(TransferEncoding.Identity)
 }
```
nsauro: Having a generic 'getHeader(key)' here might be nice as well

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
DanSimon: Good idea but I'm not going to do this quite yet as it'll conflict with another upcoming PR.  Basically we're going to use the same class for both request and response headers, so this code is going to get shifted around anyway.  I'll make sure I add that though.

PR #250, opened 2015-08-24
new faster http request head parser

DanSimon:
This significantly improves the performance of parsing HTTP requests.  Unfortunately I had to drop the parser combinators in parsing the head, though they're still being used in attaching the head to the body.  There's probably room to make this code a bit more functional but this should serve as a benchmark in terms of performance.

The new parser is more line-based and is designed to minimize the number of branches per input character.  It basically uses two sub-parsers to parse the first line and headers of the request head.  Furthermore, it parses `transfer-encoding`and `content-length` in the same pass as parsing all the headers.

All existing tests pass, but I think I might add a couple more to verify correctness before merging this in.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -5,6 +5,7 @@ import akka.util.ByteString
 import com.github.nscala_time.time.Imports._
 
 import scala.collection.immutable.HashMap
+import parsing.ParseException
 
 
 sealed abstract class HttpMethod(val name: String)
@@ -22,7 +23,7 @@ object HttpMethod {
   val methods: List[HttpMethod] = List(Get, Post, Put, Delete, Head, Options, Trace, Connect, Patch)
 
   def apply(str: String): HttpMethod = methods.find{_.name == str.toUpperCase}.getOrElse(
-    throw new HttpParsingException(s"Invalid http method $str")
+    throw new ParseException(s"Invalid http method $str")
   )
 }
 
@@ -35,7 +36,7 @@ object HttpVersion {
   case object `1.1` extends HttpVersion("1.1")
 
   def apply(str: String): HttpVersion = {
-    if (str == "HTTP/1.1") `1.1` else if (str=="HTTP/1.0") `1.0` else throw new parsing.ParseException(s"Invalid http version '$str'")
+    if (str == "HTTP/1.1") `1.1` else if (str=="HTTP/1.0") `1.0` else throw new ParseException(s"Invalid http version '$str'")
   }
 }
 
--- a/colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
@@ -7,17 +7,17 @@ import core.DataBuffer
 import parsing._
 import DataSize._
 import HttpParse._
+import Combinators._
 
 object HttpRequestParser {
-  import Combinators._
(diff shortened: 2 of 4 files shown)


colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -35,7 +36,7 @@ object HttpVersion {
   case object `1.1` extends HttpVersion("1.1")
 
   def apply(str: String): HttpVersion = {
-    if (str == "HTTP/1.1") `1.1` else if (str=="HTTP/1.0") `1.0` else throw new parsing.ParseException(s"Invalid http version '$str'")
+    if (str == "HTTP/1.1") `1.1` else if (str=="HTTP/1.0") `1.0` else throw new ParseException(s"Invalid http version '$str'")
```
nsauro: Should these Strings be vals? Or is the compiler going to optimize that?

colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
```diff
@@ -25,13 +25,152 @@ object HttpRequestParser {
...
+  val hparser = new HeaderParser
+        
+  var currentParser: MiniParser = fparser
+
+
+  def parse(d: DataBuffer): Option[HeadResult] = {
+    while (d.hasUnreadData) {
+      val b = d.next.toChar
+      if (b == '\r') {
+        headerState += 1
```
nsauro: This makes it a bit hard to follow, ie: using magic numbers vs something else a bit more descriptive.

colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
```diff
@@ -25,13 +25,152 @@ object HttpRequestParser {
...
-    HttpHead(HttpMethod(method), path, HttpVersion(version), headers)
-  }
+  protected def httpHead = new HttpHeadParser
   
-  protected def firstline  = stringUntil(' ', minSize = Some(1), allowWhiteSpace = false) ~ 
-    stringUntil(' ', minSize = Some(1), allowWhiteSpace = false) ~ 
-    stringUntil('\r', minSize = Some(1), allowWhiteSpace = false) <~ byte
+}
+
+case class HeadResult(head: HttpHead, contentLength: Option[Int], transferEncoding: Option[String] )
```
nsauro: If HttpHead had built in lazy vals for fetching contentLength/transferEncoding, would this class be necessary?

nsauro:
Looks good, my comments are pretty minor. :+1:

colossus/src/main/scala/colossus/protocols/http/Header.scala
DanSimon: I did some checking and I'm almost positive the JVM will intern the string literals, so not much else we can do here.

colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
DanSimon: I don't think there's really any other way to do it that's just as efficient, since we're literally just counting characters.  I can take a look at using some case objects, but I'd be concerned it might affect performance.  This is basically the hottest loop we have in the whole framework, so the simpler the better.

colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
DanSimon: Yes, it eliminates the extra collection traversals.

PR #251, opened 2015-08-25
Client request buffer metrics

zgagnon:
Adds two new metrics to the service client which measure the total load and the percentage of max load for the pending request buffer.

### DIFF
--- a/colossus/src/main/scala/colossus/service/ServiceClient.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceClient.scala
@@ -121,7 +121,7 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
   private val disconnects  = worker.metrics.getOrAdd(Rate(name / "disconnects"))
   private val latency = worker.metrics.getOrAdd(Histogram(name / "latency", sampleRate = 0.10, percentiles = List(0.75,0.99)))
   private val requestBufferRatio = worker.metrics.getOrAdd(Histogram(name / "request_buffer_ratio", sampleRate = 0.10, percentiles = List(0.75, 0.99)))
-  private val requestBufferLoad = worker.metrics.getOrAdd((Gauge(name / "request_buffer_load")))
+  private val requestBufferLoad = worker.metrics.getOrAdd((Gauge(name / "request_buffer_size")))
   lazy val log = Logging(worker.system.actorSystem, s"client:$address")
 
   private val responseTimeoutMillis: Long = config.requestTimeout.toMillis
--- a/colossus/src/main/scala/colossus/service/ServiceClient.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceClient.scala
@@ -35,8 +35,8 @@ import com.github.nscala_time.time.Imports.DateTime
  *
  */
 case class ClientConfig(
-  address: InetSocketAddress, 
-  requestTimeout: Duration, 
+  address: InetSocketAddress,
+  requestTimeout: Duration,
   name: MetricAddress,
   pendingBufferSize: Int = 100,
   sentBufferSize: Int = 100,
@@ -101,7 +101,7 @@ class StaleClientException(msg : String) extends Exception(msg)
  * that accepts a worker
  */
 class ServiceClient[I,O](
-  codec: Codec[I,O], 
+  codec: Codec[I,O],
   val config: ClientConfig,
   val worker: WorkerRef
 )(implicit tagDecorator: TagDecorator[I,O] = TagDecorator.default[I,O])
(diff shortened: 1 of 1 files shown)


colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -120,6 +120,8 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
   private val connectionFailures    = worker.metrics.getOrAdd(Rate(name / "connection_failures"))
   private val disconnects  = worker.metrics.getOrAdd(Rate(name / "disconnects"))
   private val latency = worker.metrics.getOrAdd(Histogram(name / "latency", sampleRate = 0.10, percentiles = List(0.75,0.99)))
+  private val requestBufferRatio = worker.metrics.getOrAdd(Histogram(name / "request_buffer_ratio", sampleRate = 0.10, percentiles = List(0.75, 0.99)))
+  private val requestBufferLoad = worker.metrics.getOrAdd((Gauge(name / "request_buffer_load")))
```
nsauro: wouldn't requestBufferSize be a bit more accurate?  Splitting hairs a bit, but we really are gauging the size.

colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -273,6 +275,8 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
       //don't allow any new requests, appear as if we're dead
       s.handler(Failure(new NotConnectedException("Not Connected")))
     } else if (isConnected || !failFast) {
+      requestBufferLoad.set(queueSize)
```
nsauro: So..I'm a bit confused.  These metrics are for the pending buffer(not yet sent).  However the only time they are being ticked is in purgeBuffers, which is called only on connectionLost/Closed(from what i'm gathering here).  Is this intentional?  Do you only care about the pending queue size when the connection is lost or closed?

If so, they should be renamed to something along those lines.

If not, then we need to also think through when to tick these metrics(especially the gauge), when we enqueue and dequeue pending requests.

zgagnon: There is a snip there in the diff. Those metrics are being ticked in attemptWrite, only in cases where it is not disconnecting. Both metrics are ticked right before a message is pushed into the pending buffer.

colossus/src/main/scala/colossus/service/ServiceClient.scala
zgagnon: that's a fair point though. Changed it to size.

nsauro:
That code fold was very well placed, and looked like all that code change was in that one function.  How annoying is that.  Doing this on send def seems like the better place to be doing it :)

colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -273,6 +275,8 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
       //don't allow any new requests, appear as if we're dead
       s.handler(Failure(new NotConnectedException("Not Connected")))
     } else if (isConnected || !failFast) {
+      requestBufferSize.set(queueSize)
+      requestBufferRatio.add(((queueSize.toDouble / config.pendingBufferSize.toDouble) * 100).toInt, Map("full" -> outputQueueFull.toString))
```
DanSimon: These two metrics seem to be basically tracking the same thing in slightly different ways, so I think we should cut one of them out, probably the size gauge.

Also, the "full" tag doesn't seem very useful, and insead the metric should be adding the `hpTags` to add host/port as tags.  Otherwise the values from every client will get merged together.

DanSimon:
I'm concerned these metrics won't add much value.  The size of the pending buffer is usually not worth paying attention to unless it's full, and there's other more direct things we can do when that happens.  Both the `queueSize` and `outputQueueFull` methods in `OutputController` are public, meaning they are directly exposed to any user of `ServiceClient`.  So it should be possible to completely avoid accidentally overloading a client.  Right now these values aren't really used.  For example, the `LoadBalancingClient`could use them to avoid accidentally overloading one of its managed connections.

Also, every time a client connection rejects a request due to the buffer being full, it fails the request with a `ClientOverloadedException`, again providing direct feedback.  I think if we do add a metric, we should be ticking the error rate every time this happens.  In fact, I think we should do that in every place we fail a request in the client.  That alone would probably provide a lot of insight as to when the buffer is full.

Lastly, I don't think a histogram is the right metric type to use, although I don't think there's a better choice right now.  Really something like a box-and-whisker metric that just gives you min, max, mean, and stddev is a better solution, since a histogram is a fairly expensive metric, but we don't have one yet.

I think this needs some more thought, though if anything it does highlight some shortcomings with metrics and back-pressure management in the `ServiceClient`.

PR #256, opened 2015-09-11
use return code tagger

aliyakamercan:
@DanSimon @nsauro 

Empty tagger was still being used by default.


colossus/src/main/scala/colossus/protocols/http/package.scala
```diff
@@ -47,8 +47,8 @@ package object http {
 
 
 
-  object ReturnCodeTagDecorator extends TagDecorator[BaseHttp#Input, BaseHttp#Output] {
-    override def tagsFor(request: BaseHttp#Input, response: BaseHttp#Output): TagMap = {
+  object ReturnCodeTagDecorator extends TagDecorator[Http#Input, Http#Output] {
```
DanSimon: I think we should find a way to make this work for both the Http and StreamingHttp codecs.  I think maybe making this a class with a type param `C <: BaseHttp` would work.

PR #257, opened 2015-09-14
fixing exception-catching holes in ConstantCallback

DanSimon:
This fixes a few places where if an exception was thrown when composing a constant callback, the exception wouldn't be properly captured.  This is extremely unlikely in the wild since you don't generally compose on a `Callback.successful`, but it popped up in testing something else.


nsauro:
+1

PR #259, opened 2015-09-21
Connection: close means close the connection

dangermike:
Support for the HTTP/1.1 `Connection: close` header ([spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.10)). Not a lot going on here other than checking that the request was HTTP/1.1 and that the `Connection` header is actually set to `close`.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -2,6 +2,7 @@ package colossus
 package protocols.http
 
 import akka.util.ByteString
+import colossus.protocols.http.Connection.{Close, KeepAlive}
 import com.github.nscala_time.time.Imports._
 
 import scala.collection.immutable.HashMap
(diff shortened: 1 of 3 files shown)


colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -197,6 +214,14 @@ case class HttpHead(method: HttpMethod, url: String, version: HttpVersion, heade
       reqString ++ ByteString("\r\n")
     }
   }
+
```
dangermike: This is where the logic lives

DanSimon:
So a better place for this to go is directly in the http request parser.  We're [already checking for several other headers](https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala#L53) and avoiding unnecessary scans of the entire set of headers.  The main difference about the connection:close header though is that the value needs to be stored in the request object itself so it can be used after request processing.  In fact we should really have dedicated values for all of the headers used directly by the http codec.

I think this is ok to merge in since it does improve functionality, but I think there is a performance cost in the extra scan so we should also create an issue to basically turn all the lazy vals in `HttpHeaderUtils` into actual values on the `HttpHead` case class.

PR #261, opened 2015-09-22
Check on post code and body

benblack86:
Fixing the test post helper method so that is actually checks the expected body.


DanSimon:
:+1:

PR #262, opened 2015-09-22
adding address for exception

amotamed:
logs remote address when parsing exception happens

### DIFF
--- a/colossus/src/main/scala/colossus/core/Connection.scala
+++ b/colossus/src/main/scala/colossus/core/Connection.scala
@@ -2,6 +2,7 @@ package colossus
 package core
 
 import akka.actor.ActorRef
+import colossus.parsing.ParseException
 
 import scala.concurrent.duration._
 import java.nio.channels.{SelectionKey, SocketChannel}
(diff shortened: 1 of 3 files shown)


colossus/src/main/scala/colossus/core/Connection.scala
```diff
@@ -194,7 +195,14 @@ private[core] abstract class Connection(val id: Long, val key: SelectionKey, _ch
   def handleRead(data: DataBuffer)(implicit time: Long) = {
     _lastTimeDataReceived = time
     myBytesReceived += data.size
-    handler.receivedData(data)
+    try {
+      handler.receivedData(data)
+    }
+    catch {
+      case parsingException:ParseException => {
+        throw new ParseException(s"error in request $remoteAddress", parsingException)
```
DanSimon: I'm not really a fan of this.  It seems too application-specific for the core layer, also kind of weird that you're rethrowing the same exception but just adding the remote address.  

I think it would be better to somehow add support for `ParseException` handling (and other "pre-request-handling" errors) into the service layer.  That way we could have a default behavior but allow the service to override handling to do things like adding more info to the message.  This would be different, though, from the existing `onError`, since that is for handling uncaught exception during request processing, while this happens before request processing even starts.  Perhaps it can be somehow unified with connection closed errors, since that happens outside of request processing as well.  I will create an issue for this.

amotamed: Cool, sounds good, I'll close this pr

PR #269, opened 2015-10-15
Remove unused dispatchers

benblack86:
It seems that only server/opentsdb dispatchers are used so I thought I'd clean up this file.


DanSimon:
Nice catch, some of those dispatchers are definitely a relic from a very old version. :+1:

PR #270, opened 2015-10-15
Include exception in error hit

benblack86:
+ Added a `failRequest` method that fails a request and ticks the error metric and is called from everywhere that was previously just calling `s.handler(Failure(_))`.
+ Differentiation between the `NotConnectedException` and `ConnectionLostException` in the `connectionLost` method.
+ `purgeBuffers` method uses the exception passed to fail the request instead of wrapping the exception in a `ConnectionLostException`.

This ignores requests that are pending.

Some things I could also do:
+ Get rid of the `droppedRequests`, `disconnects` and `connectionFailures` metrics since this information is now captured by the `errors` metric.
+ Extend RootDisconnectCause by Exception so that we can hit the error metric with a more detailed reason instead of the somewhat generic `NotConnectedException`/`ConnectionLostException` exceptions.

### DIFF
--- a/colossus/src/main/scala/colossus/service/ServiceClient.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceClient.scala
@@ -308,6 +308,7 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
   }
 
   private def failRequest(s: SourcedRequest, exception: Throwable): Unit = {
+    // TODO clean up duplicate code https://github.com/tumblr/colossus/issues/274
     errors.hit(tags = hpTags + ("type" -> exception.getClass.getName.replaceAll("[^\\w]", "")))
     s.handler(Failure(exception))
   }
--- a/colossus/src/main/scala/colossus/service/ServiceClient.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceClient.scala
@@ -215,10 +215,7 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
   }
 
   private def purgeBuffers(reason : Throwable) {
-    sentBuffer.foreach{s => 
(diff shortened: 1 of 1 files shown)


colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -308,4 +306,9 @@ extends Controller[O,I](codec, ControllerConfig(config.pendingBufferSize, config
       worker ! Kill(id.get, DisconnectCause.TimedOut)
     }
   }
+
+  private def failRequest(s: SourcedRequest, exception: Throwable): Unit = {
+    errors.hit(tags = hpTags + ("type" -> exception.getClass.getName.replaceAll("[^\\w]", "")))
```
DanSimon: I think we have the exact same classname sanitizing code in `ServiceServer`.  Let's try to move that somewhere common, or the very least add a `TODO` in the code along with a github issue to clean it up in a future PR.

benblack86: An alternative would be to use `e.getClass.getSimpleName`. If you want to keep this code, where would be a good common place?

DanSimon: Actually I definitely recall `getSimpleName` giving us some problems.... yeah I checked our old history (before open-sourcing) and we used to use `getSimpleName` but it was actually throwing exceptions for some case classes.

In terms of a common place, I don't think there's currently anywhere natural to put it.  That's why I was a little hesitant to try moving it in this PR, since I think it's going to involve a little code re-org in the process.

benblack86: Added a todo.

DanSimon:
:+1:

PR #275, opened 2015-10-20
fixing bug where reported metrics lagged by one tick interval

### LINKED ISSUE #268: Metrics are lagging by 1 tick period
This is bug in the ordering of two operations in `ActorMetrics`.  When a `Tick` is received, first the actor gets the values from each of its collectors and then calls `tick` on those collectors that need to be ticked (rates, histograms).  But this is actually the wrong order, the ticking should happen first, then retrieve the values.  This is because rates and histograms always return values generated from the previous tick period, so getting a value before ticking is returning the value aggregated from the previous tick, not the latest one.

DanSimon:
Fixes #268 

Basically ticked collectors like rates and histograms were reporting metrics off by 1 interval (and dropping the very first interval of data) due to the fact that two operations, ticking the collector and getting the data, were happening in the wrong order.  Those metrics always return data from the last tick period (to always show the last fully collected period), but since the tick was happening after the collection, it was actually returning data from the previous tick interval, not the one that just ended.

Turns out these two operations were implemented in separate methods that didn't actually need to be separate, so now they're merged and in the right order.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/ActorMetrics.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/ActorMetrics.scala
@@ -21,8 +21,7 @@ trait ActorMetrics extends Actor with ActorLogging {
       case InvalidEvent => log.error(s"Invalid event $m")
     }
     case Tick(v, interval) => {
-      val agg = metrics.aggregate(interval)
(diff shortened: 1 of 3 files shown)


benblack86:
:+1:

nsauro:
+1

Issue #189, opened 2015-04-22
Add better filtering abilities to metrics reporters


DanSimon:
Right now the filters for metrics reporters are either simple whitelists or blacklists.  We need some better filtering capabilities, particularly prefix and postfix wildcards.  For example we could filter out everything like `/foo/*` or `/*/count` where the wildcards could be one or more path components.

PR #277, opened 2015-10-26
support wild card filtering on metrics

aliyakamercan:
@DanSimon @nsauro 
resolves https://github.com/tumblr/colossus/issues/189


DanSimon:
:+1: I definitely think we're going to want to bring back the metric query parser, but that'll wait until we have more powerful aggregation capabilities.

PR #281, opened 2015-11-06
Adding timeout for OpenTsdbSender

DanSimon:
This prevents blocking forever on trying to send data to tsdb.

We can't make the timeout configurable without breaking binary compatibility, so we'll just hard-code it now for a minute and schedule configurability (along with possibly a bigger refactoring) for 0.7


reeve:
:+1:

aliyakamercan:
:+1:

nsauro:
as a temp solution, i think this is fine :+1:

PR #282, opened 2015-11-10
fixing bug where exception thrown in connection close kills worker

DanSimon:
If a connection handler (or any code executed from within) throws an exception, this will currently kill the worker.  This can eventually lead to a connection leak if while the dead worker is closing all its connections in `postStop`, another handle also throws an exception.


aliyakamercan:
:+1:

benblack86:
:+1:

nsauro:
:+1:

PR #283, opened 2015-11-11
Allow high ascii (0x80-0xFF) in memcached keys

dangermike:
Based on memcache's [protocol.txt](https://github.com/memcached/memcached/blob/master/doc/protocol.txt)

Also rejecting zero-length keys and added some tests.

@nsauro @DanSimon


DanSimon:
I personally vote we drop all validation logic.  This is very clearly adding a performance hit and we are simply attempting to duplicate logic that memcached is already going to do itself.  Especially considering that any such invalid keys are going to be very far and few.

Furthermore, it looks like what it says in protocol.txt is not true.  I successfully set memcache keys that included tabs, backspace characters, and bell characters (which amusingly actually beeps my terminal when I retrieve the key).  So while I cannot imagine a legit use case of setting such keys, we could end  up in a situation where they exist and we'd be unable to work with them.

As a compromise, what we can do is perhaps only do the validation when we receive either an `ERROR` or `CLIENT_ERROR` so we can perhaps return a better error message to the user.

dangermike:
Per Dan's point, validation is a mistake unless you can do it 100% correctly. It shouldn't be hard, but the documentation in memcached about this isn't great. So I dumped the validation. 

What I added in its place was some better handing of the errors we are getting already, including propagating the server-provided error messages up to the caller. At that point, the invalid key stuff became meaningless. The only test we should _consider_ putting in there is that the key cannot include the `SPACE` character because we know that would break the protocol -- it is used to delimit between the key and the flags.

dangermike:
Sorry, Dan. We need to ensure we aren't breaking the protocol. The way the memcache protocol is structured, we won't necessarily get an error from the server if there is a busted key. In fact, under some circumstances the server may TRUNCATE the key on storage and then REJECT the key on fetch. 

The validation is quite fast, on the order of 1.5E8 key characters/sec on my MacBook, regardless of how long I make the keys.

dangermike:
I ran a benchmark, but I don't love it. I chained GETs or SETs together using `Future.flatMap`, hitting the same key, making it essentially single threaded. The memcached instance was running on my Mac, as was the test code. Results below are sorted by operation, validation mode, and speed. The keys were 250 chars, maximizing the pain for the validation version. The performance difference was below the threshold of measurement even though we were running at latencies around 1.4ms.

```
GET with validation completed 1000 in 1381.830 ms. 723.67752 iterations/sec
GET with validation completed 1000 in 1412.020 ms. 708.2048 iterations/sec
GET with validation completed 1000 in 1416.739 ms. 705.845915 iterations/sec
GET with validation completed 1000 in 1424.332 ms. 702.083436 iterations/sec
GET with validation completed 1000 in 1435.869 ms. 696.442160 iterations/sec
```

```
GET without validation completed 1000 in 1384.155 ms. 722.462326 iterations/sec
GET without validation completed 1000 in 1386.960 ms. 721.001166 iterations/sec
GET without validation completed 1000 in 1393.307 ms. 717.716683 iterations/sec
GET without validation completed 1000 in 1398.464 ms. 715.07005 iterations/sec
GET without validation completed 1000 in 1439.648 ms. 694.613903 iterations/sec
```

```
SET with validation completed 1000 in 1413.784 ms. 707.321216 iterations/sec
SET with validation completed 1000 in 1428.993 ms. 699.79306 iterations/sec
SET with validation completed 1000 in 1430.047 ms. 699.277595 iterations/sec
SET with validation completed 1000 in 1449.220 ms. 690.025914 iterations/sec
SET with validation completed 1000 in 1520.941 ms. 657.487421 iterations/sec
```

```
SET without validation completed 1000 in 1416.443 ms. 705.993433 iterations/sec
SET without validation completed 1000 in 1418.774 ms. 704.833432 iterations/sec
SET without validation completed 1000 in 1425.187 ms. 701.66206 iterations/sec
SET without validation completed 1000 in 1426.371 ms. 701.079613 iterations/sec
SET without validation completed 1000 in 1711.524 ms. 584.274328 iterations/sec
```

DanSimon:
:+1:

PR #285, opened 2015-11-16
Support for infinite timeouts in the ServiceServer

dangermike:
Hard to run out of time when the timeout is infinite. Infinite timeouts, such as are in the example project, will break the server without this little change.


DanSimon:
:+1:

PR #286, opened 2015-11-16
don't throw exception when gracefulDisconnect it call while disconnecting

DanSimon:
Right now if `gracefulDisconnect` is called after the connection has already disconnected (which can happen when requests finish processing after the connection is terminated), we end up throwing an exception.  Really instead we should just ignore the second call, so this does that.


jlbelmonte:
:+1:

PR #287, opened 2015-11-17
enabling memcached key validation by default

dangermike:
enabling after performance testing


dangermike:
merging per blessing from @DanSimon

DanSimon:
So there is a little weirdness here how `validate` is sometimes called in the constructor and other times when encoding to bytes.  Not a blocker, but we should keep it in mind whenever we do some refactoring on this code.

:+1:

PR #291, opened 2015-12-08
Improve warning message

benblack86:
It is probably my top favorite colossus warning:

```
colossus.metrics.IntervalAggregator - Currently processing tick# 57996.  Received a tock message for an outdated tick#: 57995.  Ignoring
```

However, it would be nice to know where the tock message came from. I've added the sender value and generally tidied the code.

@DanSimon

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/MetricClock.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/MetricClock.scala
@@ -42,16 +42,16 @@ class IntervalAggregator(namespace: MetricAddress, interval: FiniteDuration, sna
       context.system.scheduler.scheduleOnce(interval, self, SendTick)
     }
 
-    case Tock(m, v) => {
-      if (v == latestTick) {
-        if(collectors.contains(sender())) {
-          build = build <+> m
+    case Tock(metrics, tick) => {
+      if (tick == latestTick) {
+        if(collectors.contains(sender)) {
+          build = build <+> metrics
           incrementCollected()
-        }else {
-          log.warning(s"Received metrics from an unregistered EventCollector: ${sender()}")
+        } else {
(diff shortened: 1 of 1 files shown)


colossus-metrics/src/main/scala/colossus/metrics/MetricClock.scala
```diff
@@ -42,16 +42,16 @@ class IntervalAggregator(namespace: MetricAddress, interval: FiniteDuration, sna
...
           incrementCollected()
-        }else {
-          log.warning(s"Received metrics from an unregistered EventCollector: ${sender()}")
+        } else {
+          log.warning(s"Ignoring metrics from an unregistered EventCollector sent by $sender")
         }
-      }else{
-        log.warning(s"Currently processing tick# $latestTick.  Received a tock message for an outdated tick#: $v.  Ignoring")
+      } else {
+        log.warning(s"Ignoring tock from $sender with tick #$tick as metric clock is at tick #$latestTick")
```
nsauro: I guess I don't see see the need for totally changing the message, wouldn't just adding $sender in there somewhere suffice.  This message feels a bit more confusing to me, but it's a subjective observation.

benblack86: Obviously I'm in favor of the new way. At the very least, the existing message has weird punctuation. I also don't see how it is "Currently processing tick". It is processing a tock message and "currently" is redundant language. Where would you suggest I put $sender in the existing log msg?

nsauro: +1.  I really don't care one way or the other :), I think as long as $sender is in there somewhere, we are better off.

DanSimon:
:+1:

PR #293, opened 2015-12-14
adding NX, EX, XX capabilities to SET function for redis

### LINKED ISSUE #292: Add support for NX, EX, XX options in redis SET command
Right now `set` just takes key/value, but according to the [docs](http://redis.io/commands/set), there are several other parameters that can be provided.

DanSimon:
Fixes #292


nsauro:
+1.  So annoyed I missed this :)

diptamay:
Any reason to use `FiniteDuration` and not `Duration` for TTL?

nsauro:
I would argue that these timeouts should never be infinitely long.

DanSimon:
Actually I like that idea, setting `Duration.inf` can be the same as not adding the `PX` parameter.  Eliminates having to wrap in an option.

PR #294, opened 2016-01-04
[WIP] Pull-based writing

DanSimon:
This is a major redesign of how data is written on open connections.  It's been on the back-burner for a while, but I've now got a solution working.  Tests do not pass (actually right now they don't even compile), but I think it's worth putting out now.

_note this is to merge into the 0.7.0 branch, not master_

### The Problem

Back in the day when a connection had data to write, it simply just immediately wrote it to the socket.  This ended up being inefficient if a connection had multiple writes during a single event-loop iteration, since each call to `SocketChannel.write` has a significant overhead (which I believe is spin-waiting for an OS-resource).  Thus it was far more efficient for connections to buffer their writes and do them all at once.  

This was originally solved by giving each connection a fixed-size buffer that would be flushed once per event-loop iteration.  While this did improve performance, it came at the cost of significantly higher memory usage (and probably reduced performance for large writes, but I never benchmarked this).  Every connection needed a 64KB buffer, which can add GB's of memory when there are thousands of open connections.  

Furthermore, all the code to accomplish this was complex and buggy, especially when the underlying socket could not accept all the data trying to be written.

### The Solution

A single large (default 4MB) output buffer is shared by all connections per worker.  Connections no longer can write whenever they choose and must instead signal a write request.  They then do the actual write via an event callback function `handleWrite` in which they get temporary exclusive access to the output buffer.  If the connection fills the output buffer, the return value of `handleWrite` signals to the worker whether the connection has more data to write.

### Side-effect : streaming overhaul

I also ended up almost completely rewriting the `OutputController` to handle the new changes.  This further resulted in changing how some of the streaming stuff works.  In particular, I have unified the `Http` and `StreamingHttp` protocols, such that there is now just a single `Http` codec, and the body of a `HttpResponse` is now an ADT that can either be a single Bytestring or a stream.  The main downside is that you no longer have direct access to the body, but doing this greatly simplified all the other code and I think is worth it.

### Results

This will immediately significantly reduce memory usage on applications with many open connections.  It will also completely eliminate the "OOM: Direct Buffer Memory" exceptions that some people have run into.  

The cleanup to the OutputController will go a long way to improving streaming support and pave the way for things like websocket and http/2 support, since it's now technically possible to transmit two streams at the same time (as defined in http/2), but this will require more code changes to actually implement.

Also initial benchmarks seem to indicate a 5-7% improvement in performance, but so far I've only benchmarked on my laptop.  At the very least there appears to be no loss of performance.

### To-do before merging

I need to get all the tests passing and probably write some new ones.  A lot of the Streaming Http tests are now obsolete.  There are definitely bugs as well, since sometimes during benchmarking connections doing heavy streaming lock up.

Furthermore, there is still a lot of unification that needs to happen with Http, but I'd prefer to put that in a separate branch.


DanSimon:
I have moved all of the http changes into another branch, as it ended up being independant of the rest of this PR and has some unresolved issues.

Basically all tests are now passing, a few I marked as ignore but they are not critical.

Also at least locally this PR is now improving performance of pipelined requests by about 12%.  I'm not entirely sure what is causing the performance increase, but I'm not complaining.  Performance of non-pipelined requests appears unaffected.

colossus/src/main/scala/colossus/controller/OutputController.scala
```diff
@@ -72,50 +163,85 @@ object OutputResult {
...
+      case Alive(msgq, dataq, livestate, disconnecting, writesEnabled) => {
+        livestate match {
+          case Streaming(source, post) => {
+            source.terminate(reason)
+            post(OutputResult.Failure(reason))
+          }
+          case _ => {}
+        }
+        if (disconnecting) {
+          //todo: flush msgq
```
reeve: is this important? :)

DanSimon: yup, fixed

Issue #279, opened 2015-11-05
Http Request parser fails on request with no headers


DanSimon:
Get's a parse exception "invalid content in header first line"

DanSimon:
Cannot reproduce, parsing the request `GET /foo HTTP/1.1\r\n\r\n` works just fine.  It's possible our sighting of this error in the wild was due to some other issue and the lack of headers was a red herring.

PR #295, opened 2016-01-08
overhauling http query parameters, fixing parsing issues

### LINKED ISSUE #288: Stop using java.net.URI to parse urls, it doesn't work
given a url like `http://foo.bar.com?key=a%3Db`, where `%3D` is a url-encoded `&`, `java.net.URI` incorrectly parses this by first url-decoding and then parsing the query string, so whereas the value of `key` should be `a&b`, it ends up just being `a`.

The solution here is to just write our own parser.  This should be very easy since we just have to split on the first `?` and then split on non-encoded `&` and `=`.

DanSimon:
Fixes #288 .  The new parser correctly handles url encoding inside query parameters.  Also switches from a map to a seq of tuples to handle multiple parameters with the same name (which is allowed) and wrapped the values in a case class with some convenience methods.

I also added a test that shows #279 appears to be invalid.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -178,19 +178,21 @@ object Connection {
 
 case class QueryParameters(parameters: Seq[(String, String)]) extends AnyVal{
 
-  def apply(key: String) = singleValue(key).get
+  def apply(key: String) = getFirst(key).get
 
   /**
-   * Get the value of a query string parameter when only at most one value is expected
+   * Get the value of a query string parameter when only at most one value is
+   * expected.  If there are multiple instances of the parameter then only the
+   * value of the first is returned
    **/
-  def singleValue(key: String): Option[String] = parameters.collectFirst{case (k,v) if k == key => v}
+  def getFirst(key: String): Option[String] = parameters.collectFirst{case (k,v) if k == key => v}
 
   /**
    * Get the values of all instances of key
    *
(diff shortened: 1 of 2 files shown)


reeve:
:+1: 
I assume we don't expect that scanning the seq for headers is expensive enough to make it worth using a multimap or something?

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -176,13 +176,51 @@ object Connection {
...
 }
 
+case class QueryParameters(parameters: Seq[(String, String)]) extends AnyVal{
+
+  def apply(key: String) = singleValue(key).get
+
+  /**
+   * Get the value of a query string parameter when only at most one value is expected
+   **/
+  def singleValue(key: String): Option[String] = parameters.collectFirst{case (k,v) if k == key => v}
```
amotamed: how do you feel about getSingle and get (or maybe get and getAll since pulling a single value from a parameter is probably the default case)

DanSimon:
I don't think it's an issue.  It's rare to have more than a few parameters and if we do hit a scenario where requests with many parameters are causing a performance hit during scanning, we'll re-address.

colossus/src/main/scala/colossus/protocols/http/Header.scala
DanSimon: hmm good point, actually I think `getFirst`and `getAll` would be more appropriate, since that's what they actually do.

amotamed: getFirst and getAll FTW!

amotamed:
:+1:

PR #296, opened 2016-01-12
Remove println

benblack86:
Using `AsyncServiceClient` in Osedax and noticed that it was using println instead of a logger.


diptamay:
Like it!

DanSimon:
These lines can be removed altogether.  I actually did this in my open PR.

benblack86:
:+1:

PR #297, opened 2016-01-12
Complete Metrics overhaul

DanSimon:
This is a major redesign of colossus-metrics which vastly simplifies everything without affecting performance.

* There is no longer local vs shared collections.  There are simply collections which are thread-safe
* No more metrics values, or metric map merging, since now only one copy of each collector exists.
* no more ActorMetrics, as it's no longer needed
* Lots of old mostly-unused code removed.

Whereas the old version kept everything thread-unsafe and behind/within actors, the new version just uses tried-and-true `ConcurrentHashMap` and `AtomicLong`

The API is almost backwards compatible, however usage in colossus has changed somewhat significantly.  Whereas each worker used to have it's own `LocalCollection`, now an entire `MetricSystem` has a single `base` collection (though users are free to add their own collections, though that currently offers no advantage).

This is not yet feature-complete, there's a few weird things to fix, and I need to fill in a bunch more tests, but at first glance it appears to be working.  Initial benchmarks show no performance difference.


colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -3,130 +3,11 @@ package colossus.metrics
 import akka.actor._
```
nsauro: unused import :)

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -142,169 +23,89 @@ trait TickedCollector extends EventCollector {
...
-      metrics.putIfAbsent(params.address, created: Local[EventCollector]) match {
-        case null => created
-        case exists: T => {
-          //because of type erasure, if we try to match against
-          //Local[T] or T with LocalLocality[T], it creates false positives (it
-          //thinks a Local[Counter] is the same type as a Local[Rate]), bad :( 
-          exists.asInstanceOf[Local[T]] 
-        }
-        case other => throw new DuplicateMetricException(s"An event collector of type ${other.getClass.getSimpleName} already exists")
+  def increment(tags: T, num: Long = 1) {
```
nsauro: small comment: this function is nearly identical to the one underneath it, maybe merge the two?

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -142,169 +23,89 @@ trait TickedCollector extends EventCollector {
...
   }
 
-  def handleEvent(event: MetricEvent): EventResult = {
-    Option(metrics.get(event.address)).map{c =>
-      c.event.andThen{_ => Ok}.orElse[MetricEvent,EventResult]({
-        case _ => InvalidEvent
-      })(event)
-    }.getOrElse{
-      UnknownMetric
+  def set(tags: T, num: Long) {
```
nsauro: maybe, if the feeling arises:
```scala
def update(tags : T, num : Long, f : (AtomicLong, Long) => _) {
    Option(map.get(tags)) match {
      case Some(got) => f(got, num)
      case None => {
        map.putIfAbsent(tags, new AtomicLong(num))
      }
    }
  }
```

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -142,169 +23,89 @@ trait TickedCollector extends EventCollector {
...
-  import Collection.ParamsFor
+  val collectors = new ConcurrentHashMap[MetricAddress, Collector]
 
-  def getOrAdd[T <: EventCollector : ClassTag, P : ParamsFor[T]#Type](params: P)(implicit creator: Generator[T, P]): Shared[T] = {
-    //create the actual collector
-    local.getOrAdd(params)
-    //get a shared interface for it
-    creator.shared(params, CollectorConfig(local.intervals))(collector)
+  def add(collector: Collector) {
+    collectors.put(collector.address, collector)
```
nsauro: Hrm..do you think it would be wise to have a few different flavors of this?  ie: in this impl, you always clobber what is already there.  Does it make sense for a few variations, or flags on this function which control whether or not you evict something if it is already there?  More of a defensive measure than anything, but still a minor point

nsauro: Ha, well, I guess if i just kept reading, I would have saw getOrAdd

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -142,169 +23,89 @@ trait TickedCollector extends EventCollector {
...
-  val globalTags: TagMap = TagMap.Empty, 
-  val intervals: Seq[FiniteDuration] = Nil,
-  metrics:ConcurrentHashMap[MetricAddress, Local[EventCollector]] = new ConcurrentHashMap[MetricAddress, Local[EventCollector]],
-  parent: Option[LocalCollection] = None
-) extends Collection[LocalLocality] {
-  import Collection.ParamsFor
-  import LocalCollection._
 
-  val localProducers = collection.mutable.ArrayBuffer[MetricProducer]()
+class CollectionMap[T] {
```
nsauro: This name def confused me :)

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -142,169 +23,89 @@ trait TickedCollector extends EventCollector {
...
-
-
-trait Collection[L <: Locality] {
-  import Collection.ParamsFor
-  //notice - the type lambda is basically the same as a view bounds (P <%
-  //MetricParams[U]), but view bounds are deprecated.  We need this instead of
-  //just subtype (P <: MetricParams[U]) becuase for some reason type inference
-  //fails when we do that
-  def getOrAdd[T <: EventCollector : ClassTag, P : ParamsFor[T]#Type](params: P)(implicit generator: Generator[T,P]): T with L
+trait Collector {
```
nsauro: Docs?

colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
```diff
@@ -36,13 +35,11 @@ class MetricInterval private[metrics](val namespace : MetricAddress, val interva
  */
 case class MetricSystem private [metrics](namespace: MetricAddress, metricIntervals : Map[FiniteDuration, MetricInterval]) {
 
-  def sharedCollection(globalTags: TagMap = TagMap.Empty)(implicit fact: ActorRefFactory) = {
-    implicit val me = this
-    SharedCollection(globalTags)
-  }
+  implicit val base = new Collection(CollectorConfig(metricIntervals.keys.toSeq))
```
nsauro: I just realized that on the apply for the MetricSystem object, metricIntervals is a Seq not a Set, we should probably fix that.

colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
```diff
@@ -2,24 +2,13 @@ package colossus.metrics
 
 import akka.actor._
```
nsauro: not used

colossus-metrics/src/main/scala/colossus/metrics/collectors/Rate.scala
```diff
@@ -4,115 +4,19 @@ import akka.actor._
 
 import scala.concurrent.duration._
```
nsauro: akka.actor._ not used :)

colossus-metrics/src/main/scala/colossus/metrics/collectors/Rate.scala
```diff
@@ -4,115 +4,19 @@ import akka.actor._
...
 
-import EventLocality._
+class Rate(val address: MetricAddress)(implicit collection: Collection) extends Collector {
```
nsauro: This class shares a lot in common with Counter.  Maybe extract a Trait or something?  I know in general, they are closely related metric types, so it kind of makes sense that their implementations are so identical

nsauro:
Overall, I like the direction of this PR, it solves a LOT of problems, mainly around the API and usage of Metrics.  Although they've always worked, it was a bit hard to understand how the pita was made. 
:100:

DanSimon:
Ok this is ready to roll so I will merge shortly.  I've created #299 and #298 as follow-ups to this PR

PR #301, opened 2016-01-19
Build artifacts from master/develop branches and all release tags

DanSimon:
Right now artifacts are only built off of master, which is a problem when it comes to publishing releases as well as building artifacts from multiple branches.  This is especially important since I with to publish a M1 release of 0.7.0 and currently can't do so from travis.

The plan going forward will be this:

* "master" and "develop" will be our snapshot branches.  Master is for ongoing work on 0.6.x and develop is for 0.7.0, so as soon as we release 0.7.0 we'll go back to just master.
* To publish a stable release (whether it's a milestone, candidate, or final release), we'll cut a branch just to change the version, the create a tag off that branch.  This way master and develop are always on "x.x.x-SNAPSHOT" versions and we never accidentally trigger a stable build from one of them.
* any tag that starts with "v*", for example "v0.7.0-M1" will be considered a release tag and have artifacts published.

I'm open to other ideas, but I think this will work.


DanSimon:
I wasn't really able to test this in a PR, but I did get it working properly in the develop branch, so we'll just leave this as-is until that branch is merged to master.

PR #302, opened 2016-01-25
Connection Handler hot swapping : Phase 1

DanSimon:
This adds all the lower-level support for correctly swapping a connection handler for a live connection.  This is a prerequisite for WebSocket support.  The service DSL and overall API are not really affected by this.  Phase 2 will add API support.

Connection Handler swapping is needed for Websocket support because a websocket connection starts out as Http and "upgrades" to the Websocket protocol.  This means a server connection has to switch protocols, and instead of somehow trying to support this functionality within a single connection handler or a single codec, it's much easier to just swap out the Http handler with a new Websocket one.

In general, the way hot-swapping works is: the graceful disconnect logic in the controller and service layers has been generalized such that when the controller is finally ready to disconnect, it instead can trigger a ConnectionHandler swap.  This is done so that when the old handler is removed, we know it is not in the middle of anything important.

In fact, the connection handler itself doesn't know what action, disconnecting or swapping, is going to take place, it is only aware of the fact that it needs to shutdown and signal back to the connection when it has finished its shutdown procedure.  For services this means ensuring that every request has been processed (either successfully or with an error) and for controllers it means all pending messages have been sent (or failed).

This also includes a fairly significant refactoring of some of the traits and classes dealing with connections and connection handlers.

### DIFF
--- a/colossus-examples/src/main/scala/colossus-examples/ChatExample.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/ChatExample.scala
@@ -123,9 +123,9 @@ class ChatHandler(broadcaster: ActorRef) extends Controller[String, ChatMessage]
     }
   }
 
-  def shutdownRequest(){
+  override def shutdownRequest(){
     push(Status("The server is shutting down, goodbye")){_ => ()}
-    gracefulDisconnect()
+    super.shutdownRequest()
   }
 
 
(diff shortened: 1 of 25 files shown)


DanSimon:
Gonna merge this with my other branches and re-open

PR #303, opened 2016-01-29
Connection hot swapping, new Service DSL

DanSimon:
This is a major change to a significant portion of the framework.  The overall goal here was supporting connection hot-swapping, but this has also resulted in a pretty big refactoring.  Here is an overview of what's in here:

* Connection Hot-swapping : I'll refer to my previous closed PR #302 for more details.  This PR is simply a continuation of that one with a lot of cleanup.
* New `CoreHandler` : This is a new class that sits in between a raw connection handler and the Controller layer.  It encapsulates all the logic dealing with the new shutdown process and provides an interface that can be used by any subsequent layer (controller, service, DSL, etc).
* New `ServerDSL` that provides a unified interface for starting servers
* Completely overhauled `ServiceDSL` that works with the server DSL to provide a unified interface for starting services.
* New `ChannelActions` trait that encapsulates all the raw operations that need to happen on a socket channel.  The `LiveConnection`trait implements this for actual connections and is only mixed in when actually created in the worker.  This all significantly simplifies testing connection and connection handler logic, and I suspect almost all of our integration tests can be replaced with unit tests.

So starting a service looks like this now:

```scala
val server = Server.start("my-server", 9000){ context =>
  
  context onConnect { connection =>

    connection accept new Service[Http] {
      
      def handle = {
        case req @ Get on Root / "foo" => req.ok("hello")

        case req @ Get on Root / "close" => {
          gracefulDisconnect()
          req.ok("bye!")
        }

        case req @ Get on Root / "websocket" => {
          this become new WebSocketController {
            //websocket stuff in here
          }
          req.ok("ok")
        }
      }
    }
  }
}

```

The full class hierarchy for `Service[T]` is :
```
WorkerItem => ConnectionHandler => CoreHandler => Controller => ServiceServer => Service
```

This means that users now have full access to all the public and protected methods of every layer, which means unlike the existing DSL, all functionality is now available to users without having to add weird accessor traits.

Lastly notice that there aren't actually any full tests of the connection hot swapping, but I plan to add those in.  There are a few other minor things to fix as well, but this PR is already getting big enough.


colossus/src/main/scala/colossus/service/ServiceDSL.scala
```diff
@@ -86,126 +55,19 @@ trait ClientCodecProvider[C <: CodecDSL] {
...
-  (config: ServiceConfig[C#Input, C#Output], worker: WorkerRef, provider: CodecProvider[C], val initializer: HandlerGenerator[C])
-  (implicit ex: ExecutionContext, tagDecorator: TagDecorator[C#Input, C#Output] = TagDecorator.default[C#Input, C#Output])
-  extends ServiceServer[C#Input, C#Output](provider.provideCodec(), config, worker) 
-  with DSLHandler[C] {
+abstract class Service[C <: CodecDSL]
+(val config: ServiceConfig[C#Input, C#Output])
+(implicit val provider: CodecProvider[C], io: IOSystem) 
+extends ServiceServer[C#Input, C#Output](provider.provideCodec(), config)(io) {
+
+  //TODO: fix it, this should pull from some kind of default config
```
DanSimon: I've created #304 which will allow us to fix this.

colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -223,32 +222,18 @@ extends Controller[I,O](codec, ControllerConfig(config.requestBufferSize, Output
...
-
-  private def checkGracefulDisconnect() {
-    if (disconnecting && requestBuffer.size == 0) {
-      super.gracefulDisconnect()
-    }
-  }
-
-  def shutdownRequest() {
-    gracefulDisconnect()
+    //TODO: there is probably a bug here if this gets called outside of processing a request, we should probably just push onto the request buffer all the time
```
DanSimon: I've created #305 to address this.

colossus-testkit/src/main/scala/colossus-testkit/MockWriteBuffer.scala
```diff
@@ -84,3 +103,9 @@ class MockWriteBuffer(val maxWriteSize: Int) extends WriteBuffer {
 
     
 }
+
+class MockWriteBuffer(val maxWriteSize: Int) extends WriteBuffer with MockChannelActions {
+  def completeDisconnect(){channelClose()}
+
+  def testWrite(d: DataBuffer): WriteStatus = write(d)
```
aliyakamercan: why do you need this ?

colossus-tests/src/test/scala/colossus/core/ConnectionHandlerSpec.scala
```diff
@@ -66,10 +64,12 @@ class ConnectionHandlerSpec extends ColossusSpec {
         if (sendBind) probe ! "BOUND"
       }
       override def onUnbind() {
+        println("UNBOUND")
         if (sendUnbind) probe ! "UNBOUND"
       }
 
       override def connected(endpoint: WriteEndpoint) {
+        println("CONNECTED")
```
aliyakamercan: println's

aliyakamercan:
`InputController`,  `OutputController` interfaces looks unnatural taking both `Input` & `Output` but I guess that was always like that.

:+1:

PR #306, opened 2016-02-10
Update memcache exception reason

benblack86:
Got this exception.

    Caused by: colossus.protocols.memcache.MemcacheInvalidCharacterException: Key contains invalid character at position 46.

Would be nice to know what the key looks like.


DanSimon:
:+1:

PR #310, opened 2016-02-18
Add mean

benblack86:
+ Added mean to histogram, which is calculated by keeping a running total.
+ Changed the metric calculated so that mean, max and min will be included with the percentiles. Tag name has changed from `percentile` to `label`.


DanSimon:
:+1:

Issue #247, opened 2015-08-19
Better support for configuring ServerCodec's from the functional interface


nsauro:
Right now, if I want to to tweak parameters to the Codec, I need to subclass the CodecProvider and then override the provideCodec function.  Would be nice if there was a way to pass in parameters via the interface for things like maxSize, etc.

PR #311, opened 2016-02-19
More API Changes

DanSimon:
This introduces a few significant changes to the API across all layers and does a lot of cleanup:

* The process of handling binding of `WorkerItems` has changed.  Previously a `WorkerItem` did not get its id and reference to the worker during construction, but afterwards when `onBind` is called.  Not `WorkerItem` is now a class which requires a `Context`, which can only be generated by a `WorkerRef`.  This context includes the unique id for the worker item.  While the worker item is still not bound during construction, now `WorkerItem` has access to its id and to the worker during construction.  
* The Server DSL has been modified so now both layers of initialization involve constructing classes.  A new `Initializer` replaces the old `ServiceContext` that had mutable variables.
* Most of the code around `Tasks` has been gutted.  I will be creating a new issue to address cleaning this up.

Starting a service now looks like this:
```scala
Server.start("hello", 8080){ worker => new Initializer(worker) {
  
  def onConnect = context => new Service[Http](context) {
    
    def handle = {
      case req => Callback.successful(req.ok("hello"))
    }
  }
}}

```

Using the `Server.basic` method (which defaults to a vanilla initializer), the code can still be shortened as such:

```scala
Server.basic("hello", 8080){ new Service[Http](_) {
  def handle = {
    case req => Callback.successful(req.ok("hello"))
  }
}}
```

As usual, the vast majority of code changes are in the tests.


DanSimon:
created #312 to address the task code

DanSimon:
Decided to just include the task fixing code in this PR.  Wasn't very much, so I'm going to close #312

DanSimon:
Added in a new Constructor to `Service` to address #247

PR #313, opened 2016-02-21
Websocket protocol

DanSimon:
_Note : This branch merges into my other unmerged PR.  Once that PR is merge I'll change this to merge into develop._

This is a mostly complete, working implementation of the websocket protocol.  It properly handles the upgrade http request and the basics of parsing and encoding websocket frames as defined in [RFC 6455](https://tools.ietf.org/html/rfc6455).

The new example server can be tested by running the examples, browsing to http://www.websocket.org/echo.html and connecting to `ws://localhost:9008`.  I've so far tested with Chome, Firefox, and Safari.

The `WebsocketController` is very basic, and I think there's a lot of room to make it more flexible and easier to use.  In particular, we could make it automatically handle PING and CLOSE opcodes, and we can probably build another layer on top that is type-parameterized for a higher-level protocol.  There's also no websocket client yet, which would be useful for integration testing.

### DIFF
--- a/colossus-examples/src/main/scala/colossus-examples/Main.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/Main.scala
@@ -38,5 +38,7 @@ object Main extends App {
 
   val benchmarkServer = BenchmarkService.start(9007)
 
+  val websocketServer = WebsocketExample.start(9008)
+
 
 }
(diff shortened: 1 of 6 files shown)


dangermike:
that is a shockingly small amount of code! well done.

nsauro:
for a first pass, looks great.  I think at some point we will recognize and abstract some of the controlling messages, so that we can provide default implementations.

PR #314, opened 2016-02-22
Small http request parsing optimizations

DanSimon:
This eliminates a couple string function calls that has a noticeable improvement in performance in benchmarks.


nsauro:
This just trims whitespace chars?

DanSimon:
I inlined two functions, one that lowercases key names in headers, the other trims leading whitespace for header values.

PR #315, opened 2016-02-22
Removing Json metrics sender

DanSimon:
This is not used anywhere and serves no real purpose, so it is being removed.


nsauro:
+1

PR #316, opened 2016-02-22
minor improvement to metrics api

### LINKED ISSUE #298: Remove cyclic dependency from metric collections
Right now (in the 0.7.0 branch) when doing `getOrAdd` on a collection, the new collector being created needs an implicit parameter for the collection.  This is so the collector has access to the set of collection intervals during construction, but it introduces a somewhat weird cyclic dependency.

It might make more sense to implicitly require the metric-system instead, or rework things to make it more similar to how it worked pre 0.7.0, where we have a separate parameters class and a typeclass for constructors.

DanSimon:
This fixes #298 .

Now it's easier than ever to get handles to metrics.  Where in the base branch you need to do:

```scala
implicit val collection = context.worker.system.metrics.base
val myrate = collection.getOrAdd(new Rate("/foo"))
```

Now you can do:

```scala
import context.worker.metrics
val myrate = Rate("/foo")
```

so you no longer need to use the collection both explicitly and implicitly.

### DIFF
--- a/colossus-examples/src/main/scala/colossus-examples/BenchmarkExample.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/BenchmarkExample.scala
@@ -49,8 +49,6 @@ object BenchmarkService {
       tcpBacklogSize = Some(1024)
     )
     val serviceConfig = ServiceConfig(
-      name = "/sample",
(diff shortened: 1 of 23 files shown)


benblack86:
:+1:

PR #317, opened 2016-02-22
new ServerContext, removing name from ServiceConfig

DanSimon:
This is the first of probably several cleanup PR's.  Trying to keep these as small as possible.

* Server connection handlers now have access to their server via the new `ServerContext`, which they accept as the constructor parameter instead of the regular `Context`.  Client connection handlers remain unchanged since they're not associated with a server
* `ClientConfig` no longer has a `name` parameter.  Instead, service servers simply use the name of their server.

### DIFF
--- a/colossus/src/main/scala/colossus/service/ServiceServer.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceServer.scala
@@ -26,8 +26,7 @@ import Codec._
  * access to the ServerRef
  */
 case class ServiceConfig(
-  name: MetricAddress,
-  requestTimeout: Duration,
+  requestTimeout: Duration = Duration.Inf,
   requestBufferSize: Int = 100,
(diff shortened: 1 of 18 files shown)


colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -26,8 +26,7 @@ import Codec._
  * access to the ServerRef
  */
 case class ServiceConfig(
-  name: MetricAddress,
-  requestTimeout: Duration,
+  requestTimeout: Duration = Duration.Inf,
```
benblack86: Does this mean that by default requests will never timeout?

DanSimon: yes

benblack86:
:+1:

PR #318, opened 2016-02-23
fix histogram min, don't report empty metrics

DanSimon:
Fixing a few minor metrics bugs:

* Right now `min` is reported as `2147483647` for a histogram with no values...oops!
* counters and rates are returning non-empty metric map with empty value maps when they have no values.


benblack86:
:+1:

PR #319, opened 2016-02-25
gracefulDisconnect => disconnect, disconnect => forceDisconnect

DanSimon:
Minor API change.  This is mostly to signify that the existing `gracefulDisconnect` should be considered the "normal" process of disconnecting, whereas `forceDisconnect` is a more disruptive action.


benblack86:
:+1:

nsauro:
makes sense

PR #321, opened 2016-02-25
Fixing member scope in ServiceServer

DanSimon:
Noticed there were some values and methods that should have been marked private.


benblack86:
:+1:

PR #322, opened 2016-02-28
Write-side optimizations

DanSimon:
This PR introduces several changes that seem to significantly increase performance.   

* Instead of a using a fixed-length output buffer, workers now use a dynamically growable buffer.  The "base" size uses a direct ByteBuffer, but if it overflows a non-direct buffer is used instead, doubling in size each time it overflows.   The direct buffer remains fixed but the overflow buffer is discarded after data has been read out.  The idea here is that once a "soft-overflow" has occurred, the writer backs off until the buffer clears, but now there's no worry about not having enough space for a single write operation.
* This significantly cleans up the logic in `OutputController`.  Now it can properly encode items on demand and will continue to write messages until it overflows the buffer.  Normally if the buffer's base size is large enough (which can be very large since there's only one per worker) it will rarely or never overflow.  The base size is not yet configurable, but I'll add an issue for it.
* This also means that output messages can now directly encode into the output buffer, totally eliminating any interim copying to other buffers.  This is because the previous fixed-size buffer may not have had enough free space to encode into, so we had to encode into a `ByteStringBuilder`, convert to a `ByteBuffer`, and then copy as much as we could and possibly hold onto the remainder for the next event-loop iteration.  All that logic is now gone.
* A bunch of optimizations have been made to `HttpResponse` in particular.  All use of `ByteString` has been eliminated, and replacing the `Vector.foreach` with a regular while loop had a big improvement.  The `String.getBytes` call on the content length was also a big offender, so it too has been replaced.

All tests pass.  I suspect code-coverage might be a little low so I may add a few more before merging.


DanSimon:
Added some more tests, fixed some bugs, code-coverage is back up.  There's definitely a bit of weirdness lurking in `OutputController` that I'll try to address in 0.7.1 or 0.8.0

colossus/src/main/scala/colossus/controller/OutputController.scala
```diff
@@ -408,39 +382,68 @@ trait OutputController[Input, Output] extends MasterController[Input, Output] {
...
+                drainSource(ls)
+                continue = false
+              }
+            }
           }
-          if (state.dataQ.itemSize == 0) {
-            continue = false
+          //TODO this can be cleaned up somehow
+          if (continue) {
+            if (state.disconnecting || state.msgQ.size > 0) {
```
nsauro: You could put these into objects with unapplys that return booleans, that way they can just be case statements l

colossus/src/main/scala/colossus/core/DataOutBuffer.scala
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  def available = Long.MaxValue //maybe limit this somehow?
+
+  def write(from: ByteBuffer) {
+    copyDestination(from.remaining).put(from)
+  }
+
+
+  def write(from: DataBuffer) {
```
nsauro: Maybe move up these impls into the trait since you probably always want these functions feeding into each otehr like this as the default behavior.

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -23,14 +23,18 @@ object HttpMethod {
 
   val methods: List[HttpMethod] = List(Get, Post, Put, Delete, Head, Options, Trace, Connect, Patch)
 
-  def apply(str: String): HttpMethod = methods.find{_.name == str.toUpperCase}.getOrElse(
-    throw new ParseException(s"Invalid http method $str")
-  )
+  def apply(str: String): HttpMethod = {
```
nsauro: The list is so small it probably doesn't matter, but this is a pretty hot path, did you try this look up using a Map or just good ol' if statements :)

nsauro:
+1.  My comments are pretty minor.

colossus/src/main/scala/colossus/controller/OutputController.scala
DanSimon: I've created #323 to address this.  Changing the way `OutputState` works will get rid of a lot of this.

colossus/src/main/scala/colossus/core/DataOutBuffer.scala
DanSimon: Actually probably going to remove the trait as there's really no need for other implementations.

DanSimon: created #324

PR #325, opened 2016-03-02
fixing tag aggregation for histograms, fixing 1.0 sample-rate

DanSimon:
Fixing a silly issue where values with different tags were overwriting each other instead of properly merging.

Also fixing another silly issue where a sample rate of 1 wouldn't collect anything


nsauro:
+1

PR #327, opened 2016-03-04
new queue-time metric for service clients

DanSimon:
Currently the "latency" metric counts both queue time and transit time.  This can sometimes make it difficult to determine if a latency is being caused on the local or remote end.

This has now been split into two metrics.  `queue_time` will track just the time a request is spent being queued, while `latency` is now strictly recording the time from transmit of the request's raw data to full decoding of the response.


DanSimon:
After some chatting with @aliyakamercan , we decided it was best to keep the total latency metric, since that is likely the metric you'll want to see most often.  So I added it back and also reduced the sample rate of the queue-time and transit-time metrics.  I think these values will give meaningful info without incurring too much of a performance hit.

nsauro:
LGTM

PR #328, opened 2016-03-06
Http data structures improvements, part 1

DanSimon:
_Notice this is to merge into a new `develop-0.7.1` branch, so this is not going into 0.7.0_

Made a bunch of improvements that both clean things up a lot and considerably improve performance.

1.  Previously headers were handled differently in requests versus responses, mainly because things were optimized for parsing requests and encoding responses.  Now there's a single `HttpHeader` trait that abstracts all that away.  Under the hood there's still two versions, one optimized for decoding and the other encoding, but now it can all be hidden from the user.
2.  Replacing the `Vector` with a plain old `Array` for the collection of headers resulted in a huge performance boost.  I tried a few different approaches to try to keep using a `Seq` or `IndexedSeq` but this one worked best.  So again to hide this detail from users and keep the API as an immutable collection, there's a new `HttpHeaders` class, used in both requests and responses.
3.  Basically did the same thing for the Http Body as well.

This isn't a complete unification but it does go a long way and this branch was starting to get pretty big.  So I'll save the rest for another PR.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/protocols/http/client/HttpResponseParserSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/protocols/http/client/HttpResponseParserSpec.scala
@@ -17,7 +17,7 @@ class HttpResponseParserSpec extends WordSpec with MustMatchers {
 
   import DecodedResult.Static
 
-  import HttpResponseHeader.Conversions._
+  import HttpHeader.Conversions._
 
   "HttpResponseParser" must {
 
@@ -25,7 +25,7 @@ class HttpResponseParserSpec extends WordSpec with MustMatchers {
       val res = ByteString("HTTP/1.1 204 NO_CONTENT\r\n\r\n")
       val parser = HttpResponseParser.static(res.size.bytes)
       val data = DataBuffer(res)
-      parser.parse(data) must equal(Some(Static(HttpResponse(HttpResponseHead(HttpVersion.`1.1`, HttpCodes.NO_CONTENT, Vector()), None))))
+      parser.parse(data) must equal(Some(Static(HttpResponse(HttpVersion.`1.1`, HttpCodes.NO_CONTENT, HttpHeaders()))))
       data.remaining must equal(0)
     }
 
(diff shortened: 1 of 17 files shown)


colossus-tests/src/test/scala/colossus/protocols/http/client/HttpResponseParserSpec.scala
```diff
@@ -35,12 +35,18 @@ class HttpResponseParserSpec extends WordSpec with MustMatchers {
...
+      val expected = HttpResponse(
         HttpVersion.`1.1`,
         HttpCodes.OK,
-        Vector("host"->"api.foo.bar:444", "accept"->"*/*", "authorization"->"Basic XXX", "accept-encoding"->"gzip, deflate", "content-length"->"0")
-      ))
-      parser.parse(DataBuffer(ByteString(res))) must equal(Some(expected))
+        Seq("host"->"api.foo.bar:444", "accept"->"*/*", "authorization"->"Basic XXX", "accept-encoding"->"gzip, deflate", "content-length"->"0")
+      )
+      val actual = parser.parse(DataBuffer(ByteString(res))).get.value
+      println(actual.head.headers.toSeq.toSet.toString)
```
nsauro: probably want to remove these printlns.

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -62,37 +64,96 @@ object HttpHeaders {
...
-  def headers : Seq[(String, String)]
+  def toEncodedHeader = new EncodedHeader(encoded, key.length, key.length + 2)
+}
+
+
+object HttpHeader {
+  def apply(key: String, value: String): HttpHeader = (new DecodedHeader(key, value)).toEncodedHeader
 
 
+  object Conversions {
```
nsauro: I like this idea.  Does it make sense to look into one that is a conversion for (String, Any), which just invokes .toString on Any?

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -62,37 +64,96 @@ object HttpHeaders {
...
+    implicit def stringTuple2Header(t: (String, String)): HttpHeader = HttpHeader(t._1, t._2)
+    implicit def seqStringTuple2Headers(t: Seq[(String, String)]): HttpHeaders = new HttpHeaders(t.map{stringTuple2Header}.toArray)
+  }
+
+}
+    
+class HttpHeaders(private val headers: Array[HttpHeader]) {
   def singleHeader(name: String): Option[String] = {
```
nsauro: I think i prefer the name 'firstHeader', it's a bit more specific to what this function is doing.

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -62,37 +64,96 @@ object HttpHeaders {
...
     val l = name.toLowerCase
-    headers.collectFirst{ case (`l`, v) => v}
+    headers.collectFirst{ case x if (x.key == l) => x.value }
   }
 
   def multiHeader(name: String): Seq[String] = {
```
nsauro: More opinionated, but i think the name 'headerValues' is a bit more descriptive.

colossus/src/main/scala/colossus/protocols/http/HttpRequest.scala
```diff
@@ -3,13 +3,77 @@ package protocols.http
 
 import akka.util.ByteString
 
-case class HttpRequest(head: HttpHead, entity: Option[ByteString]) {
+case class HttpRequestHead(method: HttpMethod, url: String, version: HttpVersion, headers: HttpHeaders) {
+
+  lazy val (path, query) = {
```
nsauro: Can we put a type on this, or at least make it a case class

colossus/src/main/scala/colossus/protocols/http/Header.scala
DanSimon: This is just old code that got moved around.  I think allowing `Any` leaves a lot of room for people to make mistakes, but I do think there might be room for something more flexible than just string.

colossus/src/main/scala/colossus/protocols/http/Header.scala
DanSimon: I actually don't think we need to repeat "header" in the name, but maybe the names `firstValue` and `allValues` might be better.

colossus/src/main/scala/colossus/protocols/http/HttpRequest.scala
DanSimon: It's just extracting the url into two lazy vals.  Again old moved code.

PR #334, opened 2016-03-09
Proxy actors for all WorkerItems

DanSimon:
This is a pretty significant improvement that takes some of the concepts originally in Tasks and generalizes them for all WorkerItems, including both server and client connection handlers.

Since the beginning of time, WorkerItems could receive actor-like messages from the outside world by sending a `Message(id, message)` to the worker.  This wasn't used much since there was no easy way to know from the outside the id and worker of a particular connection.  The coalescence of these values into the new `Context` introduced in 0.7.0 paved the way for having a cleaner reference to WorkerItems, but these contexts were still not easily exposable to the outside world, mainly because there were only created by the Worker.

Tasks, on the other hand, circumvented this by starting a proxy actor.  The actor would be returned to the user creating the task, and any messages sent to that actor would get properly routed to the running task.

So this PR does a few things:

1.  Every WorkerItem gets a proxy actor through its `Context`.  This actor is lazily constructed, so we don't incur any overhead of spinning up actors that never get used.
2.  A new `ProxyActor` trait pulls in the lazy value as the implicit `self` and provides semantics for `receive`and `sender()`, similar to Akka actors.
3.  `IOSystem` can now generate contexts.  It does this by round-robining across workers.  The current implementation is a little hacky, since it has to get the set of workers from the `WorkerManager`, but I suspect we'll be able to clean this up in the future.
4.  A new `BindWithContext` `IOCommand` is used to provide the link of creating a new WorkerItem from outside the IOSystem and getting a reference to the proxy actor.  Basically the IOSystem itself generates the context, sends that along with the `Context => WorkerItem` function, and returns the context's proxy actor.
5.  As a side effect, I did come considerable refactoring of `WorkerItemManager`, eliminating some of the complexity by having the `IOSystem`provided as a constructor parameter instead of through an initialization actor message.

Both the chat and websocket examples use these now.

There is one main outstanding issue which will have to be addressed in another PR, which is that if the proxy actor is killed, it should trigger a shutdown and unbinding of the WorkerItem.  We can't do this easily now because we need different shutdown logic for Connection handlers vs generic WorkerItems, and I didn't see a quick way to do this.


colossus/src/main/scala/colossus/core/IOSystem.scala
```diff
@@ -77,6 +115,14 @@ case class IOSystem private[colossus](workerManager: ActorRef, config: IOSystemC
    */
   val namespace = MetricAddress.Root / config.name
 
+  //ENSURE THIS IS THE LAST THING INITIALIZED!!!
+  private[colossus] val workerManager : ActorRef = managerFactory(workers, this)
+
+  // >[*]< SUPER HACK ALERT >[*]<
+  while (config.numWorkers > 0 && workers().size == 0) {
```
nsauro: Hrm...yea....the only other way I could see of doing this, is to have this class be construcable only via Future.  You would need some intermediate object which was checking the state of the Workers, and then filling a Promise when they are finished.  I don't think it's a ton of work either, but just an idea in any case.

colossus/src/main/scala/colossus/core/WorkerItem.scala
```diff
@@ -105,3 +113,47 @@ abstract class WorkerItem(val context: Context) {
...
+object ProxyActor {
+
+  type Receive = PartialFunction[Any, Unit]
+
+}
+
+class WorkerItemProxy(id: Long, worker: WorkerRef) extends Actor {
+  
+  def receive = {
+    case Worker.MessageDeliveryFailed(_,_) => {} //do anything here?
```
nsauro: log debug at the very least :)

colossus/src/main/scala/colossus/core/WorkerItem.scala
```diff
@@ -105,3 +113,47 @@ abstract class WorkerItem(val context: Context) {
...
+  def becomeReceive(receive: Receive) {
+    currentReceiver = receive
+  }
+  
+  private var lastSender = ActorRef.noSender
+  def sender() : ActorRef = lastSender
+
+  def receivedMessage(message: Any, sender: ActorRef) {
+    lastSender = sender
+    currentReceiver.applyOrElse(message, (_: Any) => ())
```
nsauro: One idea, and maybe not for this PR,(and also if I'm reading this right), is to have something akin to deadLetters here.

DanSimon: Yeah I was thinking that as well, I think we can actually just use Akka's deadletters.

PR #335, opened 2016-03-09
Cleanup output controller

DanSimon:
This is a refactoring of `OutputController` that cleans up a lot of the logic.  Previously I had moved all the mutable state into the `OutputState` ADT, but unfortunately this mostly ended up in tons of pattern matching everywhere.  So they've been moved back into the class itself.


nsauro:
looks good.  nice cleanup

PR #336, opened 2016-03-10
Metrics prune

DanSimon:
This was a feature that almost got lost in the overhaul of metrics, but it's now added back in and supports the same functionality.  `pruneEmpty` can be used to prevent rates and histograms from filling up with lots of 0 values if they are hit with many different combinations of tags.


nsauro:
LGTM

PR #337, opened 2016-03-13
More Http cleanup

DanSimon:
* Both requests and responses now use `HttpBody`, which also now includes an optional content-type header
* A new `HttpBodyEncoder` typeclass for easily turning typed data into http bodies with the correct content type header.
* Created a new `DateHeader` which can be used to generate correct values for the http date header.
* removed all the json stuff from the metrics
* cleaned up some of the interface code in `DataOutBuffer`

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -282,7 +282,7 @@ class DateHeader(start: Long = System.currentTimeMillis) extends HttpHeader {
   import java.util.Date
   import java.text.SimpleDateFormat
 
-  private val formatter = new SimpleDateFormat("EEE, MMM d yyyy HH:MM:ss z")
+  private val formatter = new SimpleDateFormat(DateHeader.DATE_FORMAT)
   
   private def generate(time: Long) = HttpHeader("Date", formatter.format(new Date(time)))
   private var lastDate = generate(start)
@@ -300,6 +300,12 @@ class DateHeader(start: Long = System.currentTimeMillis) extends HttpHeader {
     }
     lastDate.encoded
   }
+
+}
+
+object DateHeader {
+
+  val DATE_FORMAT = "EEE, MMM d yyyy HH:MM:ss z"
 
 }
 
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -269,8 +269,32 @@ case class QueryParameters(parameters: Seq[(String, String)]) extends AnyVal{
 
 }
 
+class DateHeader extends HttpHeader {
+  import java.util.Date
+  import java.text.SimpleDateFormat
 
-class HttpBody(private val body: Array[Byte])  {
+  private val formatter = new SimpleDateFormat("EEE, MMM d yyyy HH:MM:ss z")
+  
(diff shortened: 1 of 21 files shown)


colossus/src/main/scala/colossus/protocols/http/package.scala
```diff
@@ -107,12 +108,5 @@ package object http {
     )
 
     */
-  implicit object ByteStringLikeString extends ByteStringLike[String] {
-    override def toByteString(t: String): ByteString = ByteString(t)
-  }
-
-  implicit object ByteStringLikeByteString extends ByteStringLike[ByteString] {
```
nsauro: Thank god.  I hated this class.

colossus/src/main/scala/colossus/protocols/http/HttpRequest.scala
```diff
@@ -31,74 +31,64 @@ case class HttpRequestHead(method: HttpMethod, url: String, version: HttpVersion
...
+    HttpResponse(HttpResponseHead(version, code, headers), HttpBody(data))
   }
 
-  def ok(data: String, headers: List[(String, String)] = Nil)              = respond(OK, data, headers)
-  def notFound(data: String = "", headers: List[(String, String)] = Nil)   = respond(NOT_FOUND, data, headers)
-  def error(message: String, headers: List[(String, String)] = Nil)        = respond(INTERNAL_SERVER_ERROR, message, headers)
-  def badRequest(message: String, headers: List[(String, String)] = Nil)   = respond(BAD_REQUEST, message, headers)
-  def unauthorized(message: String, headers: List[(String, String)] = Nil) = respond(UNAUTHORIZED, message, headers)
-  def forbidden(message: String, headers: List[(String, String)] = Nil)    = respond(FORBIDDEN, message, headers)
+  def ok[T : HttpBodyEncoder](data: T, headers: HttpHeaders = HttpHeaders.Empty)              = respond(OK, data, headers)
```
nsauro: moar pls? :)

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -268,4 +278,82 @@ case class QueryParameters(parameters: Seq[(String, String)]) extends AnyVal{
...
+  override def toString = bytes.utf8String
+
+}
+
+trait HttpBodyEncoder[T] {
+  def encode(data: T): HttpBody
+}
+
+trait HttpBodyEncoders {
+  implicit object ByteStringEncoder extends HttpBodyEncoder[ByteString] {
```
nsauro: I keep on finding reasons not to do it, but I still can't stop myself from suggesting that we should look into a way to have a JsonEncoder in here somehow.  Here is the wrong place, but just wanted to throw an idea out.  I can think of like 5 reasons why its bad right off the bat however :)

nsauro:
+1

colossus/src/main/scala/colossus/protocols/http/Header.scala
DanSimon: The problem is that we shouldn't prescribe a particular JSON library that should be used.  If you look in the benchmark example I have one for json4s, it's only 2 lines so I don't think it's that serious if we don't include it ourselves.

nsauro: That was reason num1 for me, was forcing a JSON package.  Pretty much the biggest reason to not do it :)

PR #340, opened 2016-03-17
first pass of getting typesafe config into colossus

nsauro:
This is just for colossus-metrics.

Specifically, this PR focuses on being able to load a MetricSystem, and MetricSystem Collectors(Rate, Histogram) from a TypeSafe config source.

I'm going to leave some details out of this only because I'm hoping the ScalaDocs are informative enough to describe it.  If is not, I will update both this PR and those docs.

### DIFF
--- a/colossus-metrics/src/main/resources/reference.conf
+++ b/colossus-metrics/src/main/resources/reference.conf
@@ -1,8 +1,8 @@
 colossus {
   metrics {
     collect-system-metrics: true
-    metric-intervals: ["1 second", "1 minute"]
-    metric-address: "/"
+    collection-intervals: ["1 second", "1 minute"]
+    namespace: "/"
     collectors-defaults {
       rate {
         prune-empty: false
@@ -16,5 +16,9 @@ colossus {
 
       }
     }
+    ##Metrics can be listed here as well:
+    # /my/path/to/my/histogram {
+     # ...configuration
+    ##}
   }
 }
--- a/colossus-metrics/src/main/resources/reference.conf
+++ b/colossus-metrics/src/main/resources/reference.conf
@@ -3,17 +3,20 @@ colossus {
     collect-system-metrics: true
     collection-intervals: ["1 second", "1 minute"]
     namespace: "/"
+    enabled : true
     collectors-defaults {
       rate {
+        enabled : true
         prune-empty: false
       }
       histogram {
+        enabled : true
         percentiles: [0.75, 0.9, 0.99, 0.999, 0.9999]
         sample-rate: 1.0
         prune-empty: false
       }
       counter {
-
+        enabled : true
       }
     }
     ##Metrics can be listed here as well:
--- a/colossus-metrics/src/main/resources/reference.conf
+++ b/colossus-metrics/src/main/resources/reference.conf
@@ -1,0 +1,20 @@
+colossus {
+  metrics {
+    collect-system-metrics: true
+    metric-intervals: ["1 second", "1 minute"]
+    metric-address: "/"
+    collectors-defaults {
+      rate {
+        prune-empty: false
+      }
+      histogram {
+        percentiles: [0.75, 0.9, 0.99, 0.999, 0.9999]
+        sample-rate: 1.0
+        prune-empty: false
+      }
+      counter {
+
+      }
+    }
+  }
+}
--- a/colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
@@ -13,7 +13,9 @@ case class BucketList(buckets: Vector[Int]) extends AnyVal
  *
  * Each bucket handles an increasingly large range of values from 0 to MAX_INT.
  */
-object Histogram {
+object Histogram extends CollectorConfigLoader{
+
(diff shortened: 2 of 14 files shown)


nsauro:
I also cleaned up some minor scaladoc and whitespace issues.

nsauro:
ALso, just realized I broke some tests. Will fix those and update this.

colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
```diff
@@ -43,7 +45,54 @@ object Histogram {
...
+    * passed into the MetricSystem.apply function):
+    * 1) configPath.address
+    * 2) metricSystemConfigPath.collectors-defaults.histogram
+    * 3) colossus.metrics.collectors-defaults.histogram
+    * @param address The address relative to the Collection's MetricSystem Root.
+    * @param configPath The path in the ConfigFile that this Histogram is located.
+    * @param collection The Collection this Metric will become a part of.
+    * @return
+    */
+  def apply(address : MetricAddress, configPath : String)(implicit collection : Collection) : Histogram = {
```
DanSimon: Can't `configPath` have a default argument of "colossus.metrics" and then you can eliminate the other constructor?

nsauro: I think it depends on if we want to allow the construction of metrics with referencing any config sources.

nsauro: Sorry, that was a pre-coffee comment.  I see what you mean.

nsauro:
Made some more updates.  Take a peek when you can

colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
```diff
@@ -43,7 +45,49 @@ object Histogram {
...
+    * @return
+    */
+  def apply(address : MetricAddress)(implicit collection : Collection) : Histogram = {
+    apply(address, MetricSystem.ConfigRoot)
+  }
+
+  /**
+    * Create a Histogram with following address.  Note, the address will be prefixed with the MetricSystem's root.
+    * Configuration is resolved and overlayed as follows('metricSystemConfigPath' is the configPath parameter, if any, that was
+    * passed into the MetricSystem.apply function):
```
DanSimon: I think it would make more sense to document how configuration is loaded all in one place, possibly just on `MetricSystem` or even an empty package object, since right now it seems to be scattered a bit all over and there's a lot of duplication.

colossus-metrics/src/main/resources/reference.conf
```diff
@@ -0,0 +1,20 @@
+colossus {
+  metrics {
+    collect-system-metrics: true
+    metric-intervals: ["1 second", "1 minute"]
+    metric-address: "/"
```
DanSimon: probably better names for these parameters are "aggregation-intervals" and "root-address"....or something else, I'm open to ideas, but prefixing them with "metric" seems redundant.

colossus-metrics/src/main/scala/colossus/metrics/collectors/Rate.scala
```diff
@@ -24,15 +26,65 @@ class Rate private[metrics](val address: MetricAddress, pruneEmpty: Boolean)(imp
...
+    * @param pruneEmpty Instruct the collector to not report any values for tag combinations which were previously empty.
+    * @param collection The collection which will contain this Collector.
+    * @return
+    */
   def apply(address: MetricAddress, pruneEmpty: Boolean = false)(implicit collection: Collection): Rate = {
     collection.getOrAdd(new Rate(address, pruneEmpty))
   }
 }
 
+private[metrics] trait CollectorConfigLoader {
```
DanSimon: This trait should probably get moved into its own file, since other things besides rate mix it in.

colossus/src/main/resources/reference.conf
```diff
@@ -92,4 +92,24 @@ colossus{
...
+  metric-address : "/"
+}
+
+io-system{
+
+}
+
+
+
+imageReg{
```
DanSimon: ?

colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
nsauro: Cool, that makes sense.  I want to keep the ScalaDocs in place, but can def move some of this more repetitive stuff onto the MetricSystem's docs.

colossus-metrics/src/main/resources/reference.conf
nsauro: Yea, I can dig that.  aggregation-intervals is probably the better name, maybe reporting-intervals or collection-intervals?  

root-address or maybe namespace ?

colossus-metrics/src/main/scala/colossus/metrics/collectors/Rate.scala
nsauro: Sure, wasn't sure how this was going to evolve, but can move it

colossus/src/main/resources/reference.conf
nsauro: doh

nsauro:
updated

DanSimon:
:+1:

PR #341, opened 2016-03-18
Metrics enabling/disabling

nsauro:
This PR allows for metrics to be enabled or disabled by either specifying it directly via parameter passing into a constructor, or by including it into the Metric's configuration.

I opted to lift a Trait out for each metric type, and then provide a "Default" and "Nop" version implementations.  The Default implementations are the original implementations, whereas the Nop versions do not perform any function.  This was to avoid having to sully up the Default implementations with additional state.  I also wanted to avoid having a situation where a Metric was disabled, and still collecting values, which would be the beginnings of a memory leak(especially for a Histogram).

In addition to this new functionality, I decided to add some ScalaDocs to the existing APIs, since you know..docs are usually good, unless they are poorly written, in which case they are just bad.  I don't think I wrote bad docs, though they might not be complete.  I would def appreciate a quick once over to make sure that my docs(and me) aren't lying through our teeth.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/collectors/Counter.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/collectors/Counter.scala
@@ -68,7 +68,7 @@ class NopCounter private[metrics](val address : MetricAddress) extends Counter {
 
   override def set(tags: TagMap, value: MetricValue): Unit = {}
 
-  override def get(tags: TagMap): MetricValue = -1
+  override def get(tags: TagMap): MetricValue = 0
 }
 
 object Counter extends CollectorConfigLoader{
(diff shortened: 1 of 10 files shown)


colossus-metrics/src/main/scala/colossus/metrics/collectors/Counter.scala
```diff
@@ -22,13 +57,63 @@ class Counter private[metrics](val address: MetricAddress)(implicit collection:
...
+//Dummy implementation of a counter, used when "enabled=false" is specified at creation
+class NopCounter private[metrics](val address : MetricAddress) extends Counter {
+  val empty : MetricMap = Map()
+  override def tick(interval: FiniteDuration): MetricMap = empty
+
+  override def increment(tags: TagMap, amount: MetricValue): Unit = {}
+
+  override def set(tags: TagMap, value: MetricValue): Unit = {}
+
+  override def get(tags: TagMap): MetricValue = -1
```
DanSimon: This should probably either return 0 or be wrapped in an option, not entirely sure which though.

nsauro:
bump.  added an option to disable an entire MetricSystem as well

DanSimon:
:+1: looks good

PR #343, opened 2016-03-28
Client cleanup and unification

DanSimon:
This is the first part of a pretty significant cleanup of protocols, codecs, and clients.

There's now a fully unified way to create both callback and future clients for every protocol.  Basically for any protocol it will look like

```scala
import colossus.protocols.protocol._
import Protocol.defaults._

val client: ProtocolClient[Callback] = Protocol.client(//...

val futureClient: ProtocolClient[Future] = Protocol.futureClient(//...
```

The nice thing about this is both callback and future clients share the same set of constructors, eliminating a lot of duplicated code.  Creating non-protocol-specific clients is also still available through `ServiceClient`.

Furthermore, all of the default codec providers have a unified import structure and are no longer automatically imported with the protocol itself.  This will make it considerably easier to customize the codec providers (for example to provide custom config to a codec) without making it much harder to use the defaults.

There is still _lots_ more cleanup and improvements that can be done, but this PR is already getting large enough.  In particular some of new traits introduced in this PR can be eliminated if we have `ServiceClient` and `AsyncServiceClient` use the `CodecDSL` type param instead of separate `I,O` params and implement a common trait.


colossus-tests/src/test/scala/colossus/Util.scala
```diff
@@ -40,6 +41,33 @@ object RawProtocol {
     type Input = ByteString
     type Output = ByteString
   }
+  implicit object ServiceClientLifter extends ServiceClientLifter[Raw, RawClient[Callback]] {
```
nsauro: Can we call this the CallbackClientLifter so that its name jives with the FutureClientLifter?

colossus-tests/src/test/scala/colossus/service/ServiceClientSpec.scala
```diff
@@ -397,7 +397,7 @@ class ServiceClientSpec extends ColossusSpec {
             address = new InetSocketAddress("localhost", TEST_PORT + 1),
             connectionAttempts = PollingDuration(50.milliseconds, Some(2L)))
 
-          val client = AsyncServiceClient(config, RawProtocol.RawCodec)(io)
+          val client = AsyncServiceClient[Raw](config)//, RawProtocol.RawCodec)(io)
```
nsauro: why are these comments here?

colossus-tests/src/test/scala/colossus/service/ServiceServerSpec.scala
```diff
@@ -158,7 +159,7 @@ class ServiceServerSpec extends ColossusSpec {
             requestTimeout = 800.milliseconds,
             connectionAttempts = PollingDuration.NoRetry
           )
-          val client = new RedisFutureClient(AsyncServiceClient(clientConfig, new RedisClientCodec))
+          val client = Redis.futureClient(clientConfig)//new RedisFutureClient(AsyncServiceClient(clientConfig, new RedisClientCodec))
```
nsauro: and here?

PR #344, opened 2016-03-29
Client cleanup and unification : part 2

DanSimon:
_Note - this PR is branched off my other PR_

Building on my last PR, this has a bunch more cleanup and refactoring:

* `CodecDSL` is now `Protocol`
* `ServiceClient`, `AsyncServiceClient`, and several other class now use a single `Protocol` type parameter instead of the dual `I,O` params and implement the new `Sender`trait.
* `ResponseAdapter` has been replaced with a new `Async` typeclass.  The advantage here is the protocol-specific traits no longer need to mixin the adapter and instead use it as a typeclass, which eliminates a bunch of boilerplate.
* Did a bunch more cleanup with the client factories.  In particular all protocols use a common `LiftedClient` to mix in their protocol-specific trait and each protocol now needs to only implement one instance of `ClientLifter`.

The API for users remains the same as my other PR, so this is largely an internal-only cleanup.

### DIFF
--- a/colossus/src/main/scala/colossus/service/Async.scala
+++ b/colossus/src/main/scala/colossus/service/Async.scala
@@ -41,6 +41,10 @@ trait Async[M[_]] {
 
 }
 
+/**
+ * A Typeclass for building Async instances, used internally by ClientFactory.
+ * This is needed to get the environment into the Async.  
+ */
 trait AsyncBuilder[M[_], E] {
 
   def build(env: E): Async[M]
@@ -58,17 +62,9 @@ object AsyncBuilder {
 }
 
 
-object Async {
-
-  implicit def cbAsync: Async[Callback] = CallbackAsync
-
-  implicit def fAsync(implicit io: IOSystem) : Async[Future] = new FutureAsync
-
-}
-
 object CallbackAsync extends Async[Callback] {
   
-  type E = Unit //we don't actually need the workerRef, but having the same environment var as the constructor for the base client works nicely
+  type E = Unit //we don't actually need any environment for callbacks
 
   implicit val environment: E = ()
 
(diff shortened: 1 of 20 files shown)


colossus/src/main/scala/colossus/service/Async.scala
```diff
@@ -0,0 +1,94 @@
...
+  def success[T](t: T): Callback[T] = Callback.successful(t)
+
+  def failure[T](ex: Throwable): Callback[T] = Callback.failed(ex)
+}
+
+class FutureAsync(implicit val environment: IOSystem) extends Async[Future] {
+
+  type E = IOSystem
+
+  import environment.actorSystem.dispatcher
```
nsauro: Hrm..what if we want a separate ExecutionContext?

DanSimon: I thought about that but I don't think it's a big deal.  This is only used by Futures that the client itself generates, which right now are only doing simple map operations.  After that, you just have a regular Future and use whatever context you like.

nsauro: I think the only issue with this is that forces the user to have an IOSystem in scope at the creation site, mainly just so it can access the ActorSystem's ExecutionContext.  I'm not sure if requiring the user to carry around an IOSystem to various points in their application where they would want to create an AsyncClient is ideal.  I think switching this to using an ExecutionContext maintains the most flexibility, and doesn't preclude this case.

nsauro: Also, totally unrelated..but it just popped into my head, and I'm not sure if this a great or terrible idea:
What if the IOSystem implemented the ExecutionContext interface, and just proxied those functions to the its underlying ActorSystem?  Then, an IOSystem could be used in the same exact manner.  It both plays into and contradicts my previous comment :)

DanSimon: Well you need an IOSystem to create an AsyncServiceClient, so there's really no way around that.

Issue #263, opened 2015-09-22
Add support for handling ParseExceptions and other connection-level errors


DanSimon:
Right now if a codec throws a `ParseException` (or really any kind of exception), the connection is closed with no response.  Currently there is no ability to override this behavior in any way.

We should somehow allow users to override this behavior.  For example, for http it is actually expected for a server to send back a 400 when a malformed request is sent.  This could also be used to allow services to do their own logging/metrics for such errors.

DanSimon:
So the plan is to overhaul how `onError` works so we can handle both request exceptions and other exceptions like `ParseException`.  Any exception throw relative to a request would be wrapped in another exception with a reference to both the request and the thrown exception.

In every case a response is returned, even for parse exceptions.  This should be ok for all codecs.

PR #345, opened 2016-03-30
Yaka/req size

aliyakamercan:
Removing input size tracking from codec into input controller. Max input size can be configured using with `ControllerConfig`

@DanSimon @nsauro


colossus-tests/src/test/scala/colossus/controller/InputControllerSpec.scala
```diff
@@ -60,6 +62,17 @@ class InputControllerSpec extends ColossusSpec with CallbackMatchers{
       //now it should be done
       executed must equal(true)
     }
+    
+    "reject data above the size limit" in {
+      val input = ByteString("hello")
+      val config = ControllerConfig(4, 50.milliseconds, "test-controller", 2L)
+      val con = static(config)
+      intercept[ParseException] {
```
nsauro: Do we want to take this time and change this exception to something more specific. Technically, this is not a parsing issue, its a data size issue.  I would vote we throw a different type of exception specific to the actual cause of too large input.

colossus/src/main/scala/colossus/protocols/websocket/Websocket.scala
```diff
@@ -166,7 +166,8 @@ object FrameParser {
   }
 }
 
-abstract class WebsocketHandler(context: Context) extends Controller(new WebsocketCodec, ControllerConfig(50, scala.concurrent.duration.Duration.Inf), context) {
+abstract class WebsocketHandler(context: Context)
```
nsauro: Shouldn't this be taking in a configuration, and then passing that to the parent?

colossus-tests/src/test/scala/colossus/controller/InputControllerSpec.scala
DanSimon: I think you're right but it probably shouldn't be done here.  Right now users can't catch these exceptions anyway do it doesn't really matter what the type is.  #263 needs to happen

colossus/src/main/scala/colossus/protocols/websocket/Websocket.scala
DanSimon: I agree, but it's not really relevant to this PR, I'll create an issue for this.

DanSimon:
:+1:  the code coverage decrease seems to just be in constructors so I'm not too worried about that.

nsauro:
:pizza:

PR #347, opened 2016-03-30
typesafe config integration for IOSystem take 1

nsauro:
Typesafe Integration for IOSystems.

This borrows the same pattern that I used for the MetricSystem constructors.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/Metric.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/Metric.scala
@@ -1,7 +1,5 @@
 package colossus.metrics
 
-import scala.util.Try
-
 trait MetricFormatter[T] {
   def format(m: MetricFragment, timestamp: Long): T
 }
@@ -15,7 +13,7 @@ object OpenTsdbFormatter extends MetricFormatter[String] {
 
 
 case class MetricAddress(components: List[String]) {
-  def /(c: String): MetricAddress = copy(components = components :+ c)//.toLowercase removed for backwards compatability
+  def /(c: String): MetricAddress = /(MetricAddress(c))
 
   def /(m: MetricAddress): MetricAddress = copy(components = components ++ m.components)
 
@@ -61,19 +59,20 @@ case class MetricAddress(components: List[String]) {
 object MetricAddress {
   val Root = MetricAddress(Nil)
 
-  def fromString(str: String): Try[MetricAddress] = Try {
-    if (str == "/") MetricAddress.Root else {
-      val pieces = str.split("/")
-      if (pieces(0) != "") {
-        throw new IllegalArgumentException("Address must start with leading /")
+  def fromString(str: String): MetricAddress = {
+    str match {
+      case "" | "/" => MetricAddress.Root
+      case _ => {
+        val s = if (str startsWith "/") str else "/" + str
+        val pieces = s.split("/")
+        MetricAddress(pieces.toList.tail)
       }
-      MetricAddress(pieces.toList.tail)
     }
   }
 
-  def apply(str: String) = fromString(str).get
+  def apply(str: String) = fromString(str)
 
-  implicit def string2Address(s: String) : MetricAddress = fromString(if (s startsWith "/") s else "/" + s).get
+  implicit def string2Address(s: String) : MetricAddress = fromString(s)
 
 }
 
--- a/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
@@ -82,7 +82,7 @@ object MetricSystem {
    * @return
    */
   def apply(namespace: MetricAddress, collectionIntervals: Seq[FiniteDuration] = Seq(1.second, 1.minute),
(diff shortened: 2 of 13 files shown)


colossus-metrics/src/main/scala/colossus/metrics/Metric.scala
```diff
@@ -15,7 +13,7 @@ object OpenTsdbFormatter extends MetricFormatter[String] {
```
nsauro: These changes came about when I found a fun issue when concatting a MetricAddress.Root with "/".  When I found out what the culprit was, it looks like this class needed a bit of an update.  Everything now should work identically:
construction and appending can be invoked with strings that don't have a preceding "/".  When invoking "/" with a String that contained "/" in them it didn't properly break out the components.  I fixed that by having the /(str:String) just create a MetricAddress and invoke the other "/" function, so that behavior is identical in all cases.

colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
```diff
@@ -82,7 +82,7 @@ object MetricSystem {
    * @return
    */
   def apply(namespace: MetricAddress, collectionIntervals: Seq[FiniteDuration] = Seq(1.second, 1.minute),
-            collectSystemMetrics: Boolean = true, config : Config = ConfigFactory.defaultReference())
+            collectSystemMetrics: Boolean = true, config : Config = ConfigFactory.load())
```
nsauro: After reading typesafe config's docs, they recommended having libraries call this load function.  From what I gather from Akka's source, they call this former function. I switched as, I would rather take the advice from the library creator, rather than a user of  the library

colossus/src/main/scala/colossus/core/IOSystem.scala
```diff
@@ -15,42 +16,70 @@ import scala.concurrent.{ExecutionContext, Future}
 
 object IOSystem {
```
nsauro: Finally, actual implementation code.  Way the hell down here :)

colossus/src/main/scala/colossus/core/IOSystem.scala
```diff
@@ -15,42 +16,70 @@ import scala.concurrent.{ExecutionContext, Future}
...
   }
 
-}
+  /**
+    * Create a new IOSystem by loading its config from the specified configPath.  This configPath will be overlaid on top of the default "colossus.io-system" config path.
+    * @param configPath The path to the configuration
+    * @param config The Config source to query
+    * @param sys
+    * @return
+    */
```
nsauro: If push comes to shove, L54 might be pushed into some utility object.  I'm holding off on that until I do more integration with the Server and Clients.

colossus/src/main/scala/colossus/core/IOSystem.scala
```diff
@@ -15,42 +16,70 @@ import scala.concurrent.{ExecutionContext, Future}
...
+  def apply(name : String, workerCount : Option[Int], metrics : MetricSystem)(implicit system : ActorSystem) : IOSystem = {
+    val numWorkers = workerCount.getOrElse(Runtime.getRuntime.availableProcessors())
+
+    val workerManager = system.actorOf(Props(classOf[WorkerManager], WorkerManagerConfig(numWorkers, metrics)), s"iosystem-${actorFriendlyName(name)}-manager")
+    val sys = IOSystem(workerManager, name, numWorkers, metrics, system)
+    workerManager ! WorkerManager.Initialize(sys)
+    sys
+  }
+  private def actorFriendlyName(name : String) = {
```
nsauro: This is needed, as "/" is a nono in Actor names, so if you create your name as a multi segmented MetricAddress, Akka gets angry

nsauro: This has me thinking about the "name" parameter, and how it kind of doubles as the MetricAddress as well..something feels off about this.  It feels like its pulling double duty, and as such, causing some minor infractions.  Would you be opposed to having both a "name" which is a log/akka safe string, and MetricAddress as a separate member?

colossus-metrics/src/test/scala/colossus/metrics/MetricAddressSpec.scala
```diff
@@ -0,0 +1,44 @@
+package colossus.metrics
```
nsauro: Figure we should have these, just to solidify the desired behavior.

colossus/src/main/scala/colossus/core/IOSystem.scala
DanSimon: I'm not sure we need another name parameter, I think it's just going to make things more confusing.  It's already confusing enough that there's separate parameters for iosystem and metricssytem.

nsauro: We could always just kill name here.

DanSimon: I think it's ok to leave it for now since it defaults to just "/".  As long as we don't force users to provide names for both metric system and iosystem, I think we're ok.

DanSimon:
:+1: :+1: :+1:

PR #351, opened 2016-03-31
HttpRequest correctly encodes content length

DanSimon:
This fixes a bug where HttpRequest was not including a content-length header when it has a body.  Also cleaned up some of the code so both requests and responses use the same code for encoding the content-length header.


nsauro:
:+1:

PR #354, opened 2016-04-01
uniquely named codec defaults implicits

### LINKED ISSUE #353: Codec defaults implicit values need to be uniquely named
Right now the defaults for all protocols are just called `serverDefaults` and `clientDefaults`.  But apparently importing two implicits with the same same will cause the last one to overwrite the others.  This means trying to import defaults for multiple protocols doesn't really work.

DanSimon:
Fixes #353 

Now you can import defaults for multiple codecs without them stepping over each other.


nsauro:
+1

PR #355, opened 2016-04-04
client API cleanup, new HttpRequestBuilder traits

### LINKED ISSUE #350: Add Support for lifting any Sender to a type-specific client
Every client factory should have an apply method that takes a generic `Sender[C,M]` and lifts it to the type-specific client.  For example:

```scala
val loadBalancer: new LoadBalancingClient[Memcache](//...
val typed: MemcacheClient[Callback] = Memcache(loadBalancer)
```

this should be really easy since all the necessary machinery already exists.

### LINKED ISSUE #352: Fill in methods for Http Client
Right now the Http client is empty, we should add all the basic methods one would expect of an http client.

In particular we should have `.<method>` methods for each http method (get, post, put), with each method having support for different sets of arguments.

DanSimon:
* Fixes #352 - Or at least this is a start for adding more functionality to `HttpClient`.  The new `HttpRequestBuilder` and `HttpRequestBuilding` traits should make it much easier to build http requests.  Since `HttpRequest` itself implements these methods, you can do something like  `val req = HttpReqeust.get("/foo").withHeader("bar", "baz")`

* Fixes #350 , now you can create a type-specific client of protocol `P` from any generic `Sender[P,M]`.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/HttpClient.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/HttpClient.scala
@@ -2,13 +2,19 @@ package colossus
 package protocols.http
 import scala.language.higherKinds
 
-import core._
 import service._
 import scala.concurrent.{ExecutionContext, Future}
 
-trait HttpClient[M[_]] extends LiftedClient[Http, M] {
+trait HttpClient[M[_]] extends LiftedClient[Http, M] with HttpRequestBuilder[M[HttpResponse]]{
+
+  protected def build(req: HttpRequest) = client.send(req)
+
+  val base = HttpRequest.base
+
 
 }
+
+
 
 object HttpClient {
 
--- a/colossus/src/main/scala/colossus/protocols/http/HttpRequest.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/HttpRequest.scala
@@ -32,8 +32,29 @@ case class BuildFL(method: HttpMethod, path: String, version: HttpVersion) exten
   }
 }
 
+case class BuiltHead(firstLine: BuildFL, headers: HttpHeaders) extends HttpRequestHead
 
-case class HttpRequestHead(firstLine: FirstLine, headers: HttpHeaders) extends Encoder {
+case class ParsedHead(firstLine: ParsedFL, headers: HttpHeaders) extends HttpRequestHead
+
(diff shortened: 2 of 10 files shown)


colossus/src/main/scala/colossus/protocols/http/HttpClient.scala
```diff
@@ -2,14 +2,20 @@ package colossus
...
 import scala.language.higherKinds
 
-import core._
 import service._
 import scala.concurrent.{ExecutionContext, Future}
 
-trait HttpClient[M[_]] extends LiftedClient[Http, M] {
+trait HttpClient[M[_]] extends LiftedClient[Http, M] with HttpRequestBuilder[M[HttpResponse]]{
+
+  protected def build(req: HttpRequest) = client.send(req)
```
nsauro: This returns an M[HttpResponse]?  Build feels like a bit of a misnomer, since it is actually invoking the client and sending the request.

colossus/src/main/scala/colossus/protocols/http/HttpRequest.scala
```diff
@@ -116,13 +141,53 @@ case class HttpRequest(head: HttpRequestHead, body: HttpBody) extends Encoder {
...
+
+trait HttpRequestBuilder[T] {
+
+  def base: HttpRequest
+
+  protected def build(f: HttpRequest): T
+
+  def startMethod(method: HttpMethod, path: String) = build(base.withMethod(method).withPath(path))
+
+  def get(path: String)     = startMethod(HttpMethod.Get, path)
```
nsauro: Can we get a return type on these?

colossus/src/main/scala/colossus/protocols/http/HttpRequest.scala
```diff
@@ -116,13 +141,53 @@ case class HttpRequest(head: HttpRequestHead, body: HttpBody) extends Encoder {
       
   }
 
-  def withHeader(key: String, value: String) = copy(head = head.withHeader(key, value))
 }
 
-object HttpRequest {
+trait HttpRequestBuilding[T] {
```
nsauro: This isn't Real Scala without fancy symbolic names. :) (Please, don't do it :) )

nsauro:
LGTM

PR #356, opened 2016-04-05
Metric system config loading updates

nsauro:
- Changed Metric constructors to only process Configuration within a getOrAdd block, as to avoid unnecessarily loading Config for Metrics which already exist
- Changed how Metric configuration definitions are looked up a bit.  Now, it only looks in the defaults or in the metricSytemRoot.configPath.
- Cleaned up ConfigLoadingSpec
- Cleaned up ScalaDocs a bit

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
@@ -7,9 +7,8 @@ import scala.concurrent.duration._
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicLong
 
-
 /**
- * This is passed to new event collectors in addition to their own config.
+  * This is passed to new event collectors in addition to their own config.
  *
  * TODO: we might want to include global tags in here as well, and remove them
  * from CollectionContext.  This would mean event collectors would be
@@ -29,12 +28,14 @@ trait Collector {
 
   /**
     * The MetricAddress of this Collector.  Note, this will be relative to the containing MetricSystem's metricAddress.
+    *
     * @return
     */
   def address: MetricAddress
 
   /**
     * TODO
+    *
     * @param interval
     * @return
     */
@@ -98,6 +99,7 @@ class Collection(val config: CollectorConfig) {
 
   val collectors = new ConcurrentHashMap[MetricAddress, Collector]
 
+  //not used
   def add(collector: Collector) {
     collectors.put(collector.address, collector)
   }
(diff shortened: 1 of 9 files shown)


colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -102,8 +102,11 @@ class Collection(val config: CollectorConfig) {
     collectors.put(collector.address, collector)
   }
 
-  def getOrAdd[T <: Collector](created: T): T = {
-    collectors.putIfAbsent(created.address, created) match {
+  //The MetricAddress here should be identical to the MetricAddress that comes from T.
+  //Created is 'call by name', as to only invoke it if it is missing from from the collectors Map
+  //This is to prevent potential expensive operations, like configuration loading, from happening if the Metric is already present.
+  def getOrAdd[T <: Collector](address : MetricAddress)(created : => T): T = {
+    collectors.putIfAbsent(address, created) match {
```
DanSimon: This is still going to invoke `created` every time, since the parameter on `putIfAbsent` isn't call-by-name.  You need to do a `containsKey` call first, followed by a `get` call if it returns true and doing the same pattern matching block as below.

nsauro: Ah, right, good catch.  Will change that now

nsauro:
bump. updated

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -98,15 +99,30 @@ class Collection(val config: CollectorConfig) {
...
-  def getOrAdd[T <: Collector](created: T): T = {
-    collectors.putIfAbsent(created.address, created) match {
-      case null => created
-      case exists: T => exists
-      case other => throw new DuplicateMetricException(s"An event collector of type ${other.getClass.getSimpleName} already exists")
+  //The MetricAddress here should be identical to the MetricAddress that comes from T.
+  //Created is 'call by name', as to only invoke it if it is missing from from the collectors Map
+  //This is to prevent potential expensive operations, like configuration loading, from happening if the Metric is already present.
+  //TODO: Fix compiler warnings.  Look into TypeTags or ClassTags to fix this.
+  //This does not throw a DuplicateMetricException, rather it throws a ClassCastException for bad adds.
```
DanSimon: Let's create an issue for this, it should be pretty easy to fix

DanSimon:
👍

PR #357, opened 2016-04-05
handle parse exceptions gracefully

aliyakamercan:
Adding support to handle parse exception at service level instead of killing the connection immediately, this allows to return a proper response.

error handling functions take in a `ProcessingFailure` instead of `request`, `reason` so service can reply back even if there is not valid requests.

@DanSimon @nsauro

### DIFF
--- a/colossus-tests/src/test/scala/colossus/service/ServiceServerSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/service/ServiceServerSpec.scala
@@ -25,7 +25,8 @@ class ServiceServerSpec extends ColossusSpec {
   class FakeService(handler: ByteString => Callback[ByteString], srv: ServerContext) extends ServiceServer[ByteString, ByteString](
       config = ServiceConfig (
         requestBufferSize = 2,
-        requestTimeout = 50.milliseconds
+        requestTimeout = 50.milliseconds,
+        maxRequestSize = 300L
       ),
       codec = RawCodec,
       serverContext = srv
@@ -33,10 +34,12 @@ class ServiceServerSpec extends ColossusSpec {
 
     def processFailure(error: ProcessingFailure[ByteString]) = ByteString("ERROR")
 
+ //   override def processBadRequest(reason: Throwable) = Some(processFailure(IrrecoverableError(reason)))
+    
     def processRequest(input: ByteString) = handler(input)
 
     def receivedMessage(x: Any, s: ActorRef){}
-
+    
     def testCanPush = canPush //expose protected method
   }
 
@@ -147,8 +150,8 @@ class ServiceServerSpec extends ColossusSpec {
       val serviceConfig = ServiceConfig (
         requestTimeout = 50.milliseconds
       )
-      withIOSystem{implicit io => 
-        val server = Server.basic("test", serverSettings) { new Service[Redis](_){ 
+      withIOSystem{implicit io =>
+        val server = Server.basic("test", serverSettings) { new Service[Redis](_){
           def handle = {
               case req => Callback.schedule(500.milliseconds)(Callback.successful(StatusReply("HEllo")))
           }
@@ -170,6 +173,14 @@ class ServiceServerSpec extends ColossusSpec {
           }
         }
       }
+    }
+    
+    "gracefully handle bad input" in {
+      val t = fakeService()
+      // this is above  max request size
+      t.typedHandler.receivedData(DataBuffer(ByteString("G" * 301)))
+      t.iterate()
+      t.expectOneWrite(ByteString("ERROR"))
     }
 
   }
--- a/colossus-tests/src/test/scala/colossus/service/ServiceServerSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/service/ServiceServerSpec.scala
@@ -30,7 +30,7 @@ class ServiceServerSpec extends ColossusSpec {
       codec = RawCodec,
       serverContext = srv
   ) {
-    // def processFailure(request: ByteString, reason: Throwable) = ByteString("ERROR")
+
     def processFailure(error: ProcessingFailure[ByteString]) = ByteString("ERROR")
 
     def processRequest(input: ByteString) = handler(input)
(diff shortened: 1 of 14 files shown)


colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -243,10 +250,19 @@ with ServerConnectionHandler {
...
 
   //DO NOT CALL THIS METHOD INTERNALLY, use handleFailure!!
-  protected def processFailure(request: I, reason: Throwable): O
+ 
+ // protected def processFailure(request: I, reason: Throwable): O
+  protected def processFailure(error: ProcessingFailure[I]): O
+  
+}
 
+trait ProcessingFailure[C] {
```
nsauro: sealed?

colossus-tests/src/test/scala/colossus/service/ServiceServerSpec.scala
```diff
@@ -30,7 +30,8 @@ class ServiceServerSpec extends ColossusSpec {
       codec = RawCodec,
       serverContext = srv
   ) {
-    def processFailure(request: ByteString, reason: Throwable) = ByteString("ERROR")
+    // def processFailure(request: ByteString, reason: Throwable) = ByteString("ERROR")
```
nsauro: commented

colossus/src/main/scala/colossus/controller/InputController.scala
```diff
@@ -1,11 +1,14 @@
 package colossus
 package controller
 
+import colossus.controller.OutputResult.Failure
```
nsauro: Personally, I would change this name, so it doesn't clash with a class name that is widely used by Scala.

colossus/src/main/scala/colossus/controller/InputController.scala
```diff
@@ -206,5 +214,10 @@ trait InputController[Input, Output] extends MasterController[Input, Output] {
 
 
   protected def processMessage(message: Input)
+  
+  protected def processBadRequest(reason: Throwable): Output = {
+    disconnect()
+    throw reason
```
DanSimon: So there's a few issues here.  

1.  it shouldn't throw the exception, it should just disconnect.  
2.  There need to be two new methods.  `processBadRequest`needs to be abstract, so that we force subclasses to implement it.  Then we need another method, let's call it `fatalInputError` that does 3 things: it calls `processBadRequest`, it writes the `Output`object that it returns, and it calls `disconnect()`.  It has to work this way because when an error like this occurs, we have no option but to close the connection.  So therefore it should not be up to subclasses to call disconnect, and we don't want that behavior changed if they override the method.
2.  The methods need to be defined in `Controller`, not `InputController`, because it needs to be able to push the output object, which `InputController`cannot do.

colossus/src/main/scala/colossus/controller/InputController.scala
```diff
@@ -133,18 +133,24 @@ trait InputController[Input, Output] extends MasterController[Input, Output] {
...
+              }
+              case Some(DecodedResult.Stream(msg, sink)) => {
+                decoding = false
+                inputState = ReadingStream(sink)
+                processAndContinue(msg, data)
+              }
             }
-            case None => {
+          } catch {
+            case reason: Throwable =>
```
DanSimon: This should have curly braces wrapping the case block

colossus/src/main/scala/colossus/service/ServiceDSL.scala
```diff
@@ -119,7 +121,8 @@ extends ServiceServer[C#Input, C#Output](codec, config, srv) {
     
   protected def processRequest(i: C#Input): Callback[C#Output] = handler(i)
   
-  protected def processFailure(request: C#Input, reason: Throwable): C#Output = errorHandler((request, reason))
+  //protected def processFailure(request: C#Input, reason: Throwable): C#Output = errorHandler((request, reason))
```
DanSimon: probably should remove this

colossus/src/main/scala/colossus/protocols/websocket/package.scala
```diff
@@ -30,7 +30,7 @@ package object websocket {
     def provideCodec() = new WebsocketCodec
 
     //TODO : looks like we need to break this out from codec provider
-    def errorResponse(request: Frame, reason: Throwable): Frame = ???
+    def errorResponse(error: ProcessingFailure[Frame]): Frame = ???
```
DanSimon: I created #361 for this

colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -205,25 +201,34 @@ with ServerConnectionHandler {
...
+    Some(handleFailure(IrrecoverableError(reason)))
+  }
 
-  private def handleFailure(request: I, reason: Throwable): O = {
-    addError(request, reason)
-    processFailure(request, reason)
+  private def handleFailure(error: ProcessingFailure[I]): O = {
+    error match {
+      case RecoverableError(request, reason) => addError(request, reason)
+      case _: IrrecoverableError[I] => // TODO: tick/log w/o request
```
DanSimon: I feel like this should be implemented in this PR.  I think you can just have `addError`take an `Option[Input]`

colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -243,10 +248,19 @@ with ServerConnectionHandler {
   protected def processRequest(request: I): Callback[O]
 
   //DO NOT CALL THIS METHOD INTERNALLY, use handleFailure!!
-  protected def processFailure(request: I, reason: Throwable): O
+ 
+ // protected def processFailure(request: I, reason: Throwable): O
```
DanSimon: more commented code

DanSimon:
I think this needs a couple more tests before it can get merged, otherwise aside from my minor points it looks pretty good

DanSimon:
👍

PR #358, opened 2016-04-05
new MetricNamespace and MetricContext

DanSimon:
This PR solves a lot of issues we've faced in metrics with relative paths.  Now there is a new `MetricNamespace` that is required when creating new metrics.  `MetricSystem` implements this trait and can be used as a namespace, but now we have the ability to create sub-spaces with extended prefixes.

This is now used by all components of Colossus that create their own metrics.

### DIFF
--- a/colossus-examples/src/main/scala/colossus-examples/WebsocketExample.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/WebsocketExample.scala
@@ -52,10 +52,10 @@ object WebsocketExample {
     
     val generator = io.actorSystem.actorOf(Props[PrimeGenerator])
 
-    Server.basic("websocket", port){ new Service[Http](_) {
+    Server.basic("websocket", port){ ctx => new Service[Http](ctx) {
       def handle = {
         case UpgradeRequest(resp) => {
-          become(new WebsocketHandler(_) with ProxyActor {
+          become(_ => new WebsocketServerHandler(ctx) with ProxyActor {
 
             private var sending = false
 
--- a/colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
@@ -83,7 +83,7 @@ object Histogram extends CollectorConfigLoader{
     * @param collection The collection which will contain this Collector.
     * @return
     */
-  def apply(address : MetricAddress)(implicit collection : Collection) : Histogram = {
+  def apply(address : MetricAddress)(implicit ns : MetricNamespace) : Histogram = {
     apply(address, MetricSystem.ConfigRoot)
   }
 
@@ -97,11 +97,11 @@ object Histogram extends CollectorConfigLoader{
     * @param collection The collection which will contain this Collector.
     * @return
     */
-  def apply(address : MetricAddress, configPath : String)(implicit collection : Collection) : Histogram = {
+  def apply(address : MetricAddress, configPath : String)(implicit ns : MetricNamespace) : Histogram = {
 
     import scala.collection.JavaConversions._
 
-    val params = resolveConfig(collection.config.config, s"$configPath.$address", DefaultConfigPath)
+    val params = resolveConfig(ns.collection.config.config, s"$configPath.$address", DefaultConfigPath)
 
     val percentiles = params.getDoubleList("percentiles").map(_.toDouble)
     val sampleRate = params.getDouble("sample-rate")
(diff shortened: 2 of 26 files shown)


colossus-examples/src/main/scala/colossus-examples/WebsocketExample.scala
```diff
@@ -52,10 +52,10 @@ object WebsocketExample {
     
     val generator = io.actorSystem.actorOf(Props[PrimeGenerator])
 
-    Server.basic("websocket", port){ new Service[Http](_) {
+    Server.basic("websocket", port){ ctx => new Service[Http](ctx) {
       def handle = {
         case UpgradeRequest(resp) => {
-          become(new WebsocketHandler(_) with ProxyActor {
+          become(_ => new WebsocketServerHandler(ctx) with ProxyActor {
```
DanSimon: I've created #359 to address the fact that the input context is now not used here.

colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
```diff
@@ -236,7 +236,8 @@ class ServerSpec extends ColossusSpec {
         }
       }
 
-      "open up spot when connection closes" in {
+      //TODO : fix this test
```
DanSimon: I've created #360 to address this

colossus/src/main/scala/colossus/core/Worker.scala
```diff
@@ -157,11 +156,10 @@ private[colossus] class Worker(config: WorkerConfig) extends Actor with ActorLog
 
   private val workerIdTag = Map("worker" -> (io.name + "-" + workerId.toString))
 
-  import config.io.metrics.base
-
-  val eventLoops              = Rate(io.namespace / "worker" / "event_loops")
-  val numConnections          = Counter(io.namespace / "worker" / "connections")
-  val rejectedConnections     = Rate(io.namespace / "worker" / "rejected_connections")
+  implicit val ns = io.metrics
```
nsauro: I might be mistaken, but isn't this changing the address of these metrics?  Before, they were using the io.namespace, now they are using the MetricSystem namespace, wouldn't that mean they lose any iosystem prefix that was there previously?

DanSimon: Ah yeah I forgot the IOSystem needs its own namespace

colossus/src/main/scala/colossus/controller/Controller.scala
```diff
@@ -42,6 +41,8 @@ trait MasterController[Input, Output] extends ConnectionHandler {
   protected def codec: Codec[Output, Input]
   protected def controllerConfig: ControllerConfig
 
+  implicit def namespace : metrics.MetricNamespace
```
nsauro: Not sure for Controller, but shouldn't this also be using an IOSystem's namespace?

colossus/src/main/scala/colossus/core/Server.scala
```diff
@@ -95,6 +95,8 @@ case class ServerRef private[colossus] (config: ServerConfig, server: ActorRef,
 
   def serverState = serverStateAgent.get()
 
+  val namespace : MetricNamespace = system.metrics / name
```
nsauro: here as well.

colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -113,20 +113,20 @@ with ClientConnectionHandler with Sender[P, Callback] with ManualUnbindHandler {
 
   import colossus.core.WorkerCommand._
   import config._
-  import context.worker.metrics
+  implicit val namespace = context.worker.system.metrics / config.name
```
nsauro: This actually brings up a question, which I"m not sure I know if we answered or not, which is , how should Client addresses be namespace?  I guess IOSystem?  I can't think of anything lower that would make sense

colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
```diff
@@ -97,11 +97,11 @@ object Histogram extends CollectorConfigLoader{
     * @param collection The collection which will contain this Collector.
     * @return
     */
-  def apply(address : MetricAddress, configPath : String)(implicit collection : Collection) : Histogram = {
+  def apply(address : MetricAddress, configPath : String)(implicit ns : MetricNamespace) : Histogram = {
 
     import scala.collection.JavaConversions._
 
-    val params = resolveConfig(collection.config.config, s"$configPath.$address", DefaultConfigPath)
```
nsauro: config.config makes me sad.  I've bought this upon us, I'll see if i can make it better somehow

colossus/src/main/scala/colossus/controller/Controller.scala
DanSimon: This one is abstract, but which namespace it uses depends on whether it's a server or client.

colossus/src/main/scala/colossus/service/ServiceClient.scala
DanSimon: yeah it should be iosystem.

colossus/src/main/scala/colossus/controller/Controller.scala
nsauro: Should it not have anything assigned to it, to force subs to implement it ?

DanSimon: it doesn't.

nsauro: Lol.  wow, I need to learn to read :)

DanSimon:
Fixed the IOSystem namespace issues, fixed conflicts from develop, and fixed the `ClassCastException` issue with `getOrAdd`.

nsauro:
👍

PR #362, opened 2016-04-07
CoreHandler calls shutdown in NotConnected State, fixing ServiceClient

### LINKED ISSUE #278: Client keeps trying to connect when disconnect is called in connecting state
So if a service client fails to connect to a server, and the retry policy is set to retry connecting, the client will keep attempting to connect forever, even if disconnect is called.

DanSimon:
Fixes #278 .  Now service clients will properly stop trying to connect if `disconnect` is called during the connection process.  This was basically solved by having `CoreHandler`invoke the shutdown process while in the `NotConnected`state.  Previously it would only do it in the `Connected` state.


nsauro:
👍

Issue #361, opened 2016-04-06
errorResponse needs to be separate from CodecProvider


DanSimon:
`errorResponse` is only used with services, but it currently defined in `CodecProvider` which is used for other servers as well.  For example, right now the websocket protocol has to implement this, but it doesn't really make sense in that situation.

It may make sense to move it into a new `ServiceCodecProvider` that simply extends `CodecProvider`

PR #363, opened 2016-04-07
service codec provider extends codec provider

aliyakamercan:
@DanSimon @nsauro 

Added `ServiceCodecProvider` which extends `CodecProvider` and moved `errorResponse` to it. This way codecs used by server but not services won't have to implement `errorResponse`.


aliyakamercan:
fixes https://github.com/tumblr/colossus/issues/361

nsauro:
+1.  Looks good, unless I'm missing something

DanSimon:
👍

PR #365, opened 2016-04-08
remove context from lambda in become

### LINKED ISSUE #359: ConnectionHandler become needs to work with both Context and ServerContext
Right now become take a function of type `Context => ConnectionHandler`, but this doesn't work for anything that requires a `ServerContext`.  This logic should be reworked so either the function can somehow take either one (like a common trait or something) or no longer takes a lambda as an argument.

aliyakamercan:
This fixes #359 

Connection type become lambda no longer takes a context.

@DanSimon @nsauro

### DIFF
--- a/colossus/src/main/scala/colossus/core/CoreHandler.scala
+++ b/colossus/src/main/scala/colossus/core/CoreHandler.scala
@@ -8,7 +8,7 @@ sealed abstract class ShutdownAction(val rank: Int) {
 
 object ShutdownAction {
   case object DefaultDisconnect extends ShutdownAction(0)
-  case class Become(newHandler: Context => ConnectionHandler) extends ShutdownAction(1)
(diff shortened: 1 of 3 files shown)


colossus/src/main/scala/colossus/core/CoreHandler.scala
```diff
@@ -82,10 +84,12 @@ abstract class CoreHandler(ctx: Context) extends WorkerItem(ctx) with Connection
    * Replace this connection handler with the given handler.  The actual swap
    * only occurs when the shutdown process complete
    */
-  final def become(nh: Context => ConnectionHandler): Boolean = if (setShutdownAction(ShutdownAction.Become(nh))) {
+  final def become(nh: () => ConnectionHandler): Boolean = if (setShutdownAction(ShutdownAction.Become(nh))) {
```
nsauro: what if we used call by name instead?

DanSimon:
👍

Issue #332, opened 2016-03-07
Make Http Bodies content-type aware


DanSimon:
Right now we always expose a http message's body simply as a string or bytestring.  We can probably improve on this by allowing bodies to be typed based on the content-type header.

On the encoding end, we kind of started to do this with the `ByteStringLike` typeclass, but not only do we need to switch from ByteString's to HttpBody, but we need to include logic so that the type can also write the correct header.  So if I set my response's body as a json structure, it automatically adds the `content-type: application/json` header.

On the decoding end we can add a `.as[T]` where `T` is some type that has an associated typeclass that can check the content-type header and parse the body into some other type.

DanSimon:
The encoding side of this has already been solved in #337 .

PR #367, opened 2016-04-11
New HttpBody constructors

### LINKED ISSUE #349: Improve HttpBody constructors and implicit lifting
Right now strings encode by default with `text/plain` content-type, and overriding that (say, for rendering a string of HTML) requires more boilerplate than it needs to:

```scala
new HttpBody(htmlString.getBytes("UTF-8"), Some(HttpHeader("Content-Type", "text/html")))
```

We should make it easier to override the content-type of various types and provide some alternative constructors so users don't need to construct a full header.

DanSimon:
Fixes #349.  Now you can easily create a `HttpBody` using an encoder, but with a custom content-type header.

I also moved `HttpBody` to it's own file


aliyakamercan:
:+1:

DanSimon:
I also added a new `HttpBodyDecoder` typeclass, which fixes #332, as well as some tests

PR #368, opened 2016-04-12
Metric address as config path

### LINKED ISSUE #364: Metrics are not being stored in a MetricContext's Collection correctly
MetricContexts share Collections.  When a Metric is added to a Collection, it is being stored using its relative MetricAddress, not the full MetricAddress which is prefixed by the MetricContext.  

Cases that break:

```scala
//not using implicits here, just to illustrate the point.  Imagine these 2 metrics are created miles away from each other
val namespace = MetricContext("/", new Collection(CollectorConfig(Seq(1.minute, 1.second))))
val subName = namespace / "bar"

val rate = Rate("bar", false, true)(namespace)
//miles away
val rate2 = Counter("bar", true)(subName)
//the above line throws a DuplicateMetricException
```
another:
```scala
//not using implicits here, just to illustrate the point.  Imagine these 2 metrics are created miles away from each other
val namespace = MetricContext("/", new Collection(CollectorConfig(Seq(1.minute, 1.second))))
val subName = namespace / "bar"

val rate = Rate("bar", false, true)(namespace)
//miles away
val rate2 = Rate("bar", true)(subName)
 namespace.collection.collectors.size()  //should be 2 .. is 1
```

Potential Solutions:
We have a few options:
- Don't share collections.  Every sub should have a new collection, which gets registered with the MetricSystem on creation
- proxy "getOrAdd" on the MetricContext, so that it can be make sure the namespace is being properly applied

One thing which is bothering me a tad, is that even if we did proxy the getOrAdd function, actual Metric creation happens outside of the Context, so there is no guarantee that the Metric's address is actually in the namespace.  Not sure if there is much we can do about that, at least not without going crazy.

nsauro:
A few things in here:

- MetricAddresses are now used as a config path, and are the first place a Metric is configured from
- Fixed #364.  MetricNamespace now exposes a getOrAdd function to ensure that addresses are properly prefixed when added to a Collection.  At some point, we might want to make Collection fully private, but that will be a bit of a change which we can punt on for now
- import cleanup, to remove compiler warnings

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/collectors/Counter.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/collectors/Counter.scala
@@ -43,7 +43,7 @@ trait Counter extends Collector {
 }
 
 //Working implementation of a Counter
-class DefaultCounter private[metrics](val address: MetricAddress)(implicit ns : MetricNamespace) extends Counter {
+class DefaultCounter private[metrics](val address: MetricAddress) extends Counter {
 
   private val counters = new CollectionMap[TagMap]
 
@@ -85,7 +85,7 @@ object Counter extends CollectorConfigLoader{
     * Create a Counter with the following address.   See the documentation for [[colossus.metrics.MetricSystem]] for details on configuration
     *
     * @param address The MetricAddress of this Counter.  Note, this will be relative to the containing MetricSystem's metricAddress.
-    * @param collection The collection which will contain this Collector.
+    * @param ns The namespace to which this Metric is relative.
     * @return
     */
   def apply(address : MetricAddress)(implicit ns : MetricNamespace) : Counter = {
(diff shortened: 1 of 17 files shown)


colossus-metrics/src/main/scala/colossus/metrics/collectors/Counter.scala
```diff
@@ -98,12 +98,13 @@ object Counter extends CollectorConfigLoader{
...
-    * @param collection The collection which will contain this Collector.
+    * @param ns The namespace to which this Metric is relative.
     * @return
     */
   def apply(address : MetricAddress, configPath : String)(implicit ns : MetricNamespace) : Counter = {
-    ns.collection.getOrAdd(address){
-      val params = resolveConfig(ns.collection.config.config, s"$ConfigRoot.$configPath", s"$ConfigRoot.$DefaultConfigPath")
+    ns.getOrAdd(address){ (fullAddress, config) =>
+      val addressPath = fullAddress.pieceString.replace('/','.')
+      val params = resolveConfig(config.config, addressPath, s"$ConfigRoot.$configPath", s"$ConfigRoot.$DefaultConfigPath")
```
DanSimon: looks like this is duplicated in every collector.

DanSimon:
overall looks good to me

colossus-metrics/src/main/scala/colossus/metrics/collectors/Counter.scala
nsauro: I've fixed this in my next branch.

PR #370, opened 2016-04-15
Server typesafe config

nsauro:
## What
Ability to configure Servers via typesafe config

## How
Following the same pattern as `IOSystem` and `MetricSystem`'s constructor functions.  This one is a bit different, due to the `ServerDSL` trait, which has a bunch of convenience functions for creating `Server`s.  I followed the pattern for IO/MetricSystem on there, instead of the `Server.apply` functions.  

Some difficulty snuck in with configuring a Server's metrics: 

A `MetricSystem` during its creation, takes any user overrides and applies them onto the defaults.  This merged object is then placed back onto the original Config, and this modified Config is then set on the `MetricSystem` as its configuration source.  `Collector`s then use this already merged config source to pull their configurations from, vs having to manage the overrides and defaults themselves.

`Server`s can provide their own `metrics` definition, which contains configuration for its internal metrics.  This configuration needs to be overlaid on top of the underlying `MetricSystem`'s merged configuration in order for them to be properly applied.  Since this happens after a `MetricSystem` creation, this new config needs to somehow be surfaced to the `Server`'s metric constructors, so that they can be properly configured.

The `Server` creates this new configuration, by overlaying its `metrics` configuration (if it exists) on the underlying `MetricSystem`'s configuration.  From there, there were 2 options(probably more), that presented themselves:

- Inside `Server`, create a new `MetricNamespace`, whose underlying `Collection`, contained a pointer to the new configuration, and use that `MetricNamespace` in the `Collector`'s constructors.  

- Modify the contructors for `Collector`s to allow them to specify a Config object.  This Config object would take precedence over the underlying `MetricSystem`'s configuration.

I opted for the second, only because I thought of it first, and it was pretty straightforward.  The only problem is now all `Collector`s have this new constructor, which is kind of not great.  However, having this extra constructor does afford more obvious configuration options for `Collector` creators, which I think is a good thing.

I'm still slightly torn on which way is better(or if there is a better option all together), but think for a first pass this works, and we can always improve on it.  If there is strong opinion to go another route, I'm all ears.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
@@ -195,7 +195,7 @@ object MetricSystem {
     */
   def apply(configPath : String, config : Config)(implicit system : ActorSystem) : MetricSystem = {
 
-    import MetricSystemConfigHelpers._
+    import ConfigHelpers._
 
     val userPathObject = config.getObject(configPath)
     val metricsObject = userPathObject.withFallback(config.getObject(ConfigRoot))
@@ -218,13 +218,29 @@ object MetricSystem {
 }
 
 //has to be a better way
-object MetricSystemConfigHelpers {
+object ConfigHelpers {
 
-  implicit class FiniteDurationLoader(config : Config) {
+  implicit class ConfigExtractors(config : Config) {
 
     import scala.collection.JavaConversions._
 
+    def getIntOption(path : String) : Option[Int] = getOption(path, config.getInt)
+
+    def getLongOption(path : String) : Option[Long] = getOption(path, config.getLong)
+
+    private def getOption[T](path : String, f : String => T) : Option[T] = {
+      if(config.hasPath(path)){
+        Some(f(path))
+      }else{
+        None
+      }
+    }
+
     def getFiniteDurations(path : String) : Seq[FiniteDuration] = config.getStringList(path).map(finiteDurationOnly(_, path))
+
+    def getFiniteDuration(path : String) : FiniteDuration = finiteDurationOnly(config.getString(path), path)
+
+    def getScalaDuration(path : String) : Duration =  Duration(config.getString(path))
 
     private def finiteDurationOnly(str : String, key : String) = {
       Duration(str) match {
--- a/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
@@ -63,6 +63,27 @@ trait MetricNamespace {
    * Get a new namespace by appending `subpath` to the namespace
    */
   def /(subpath: MetricAddress): MetricNamespace = MetricContext(namespace / subpath, collection)
+
+
+  /**
+    * Create a new namespace using the supplied config overlaid on top of the existing MetricSystem configuration.
+    * This function is used for components that want to add Metrics into this namespace, but want to provide additional or overriding config.
+    * Note, if there are name clashes with existing elements in the current Config, the existing elements will be overwritten for this new MetricNamespace instance.
+    * @param config  Config object which has a `metrics` field on it.  This field is expected to be in the shape of a `colossus.metrics` config definition.
+    *                If no `metrics` field is detected, then this Config object is not used.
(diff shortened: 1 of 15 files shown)


colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
```diff
@@ -218,14 +218,30 @@ object MetricSystem {
 }
 
 //has to be a better way
```
nsauro: This object grew a bit, as `ServerSetting` had to expand this a bit. Having this here feels weird, but there is no other location where this can be shared between metrics and core.

colossus-metrics/src/main/scala/colossus/metrics/collectors/CollectorConfigLoader.scala
```diff
@@ -20,4 +21,10 @@ private[metrics] trait CollectorConfigLoader {
       }
     }
   }
+
```
nsauro: This is an abstraction of that dupe code that was previously in every Collector.

colossus-metrics/src/main/scala/colossus/metrics/collectors/Histogram.scala
```diff
@@ -104,12 +106,28 @@ object Histogram extends CollectorConfigLoader{
     * @return
     */
   def apply(address : MetricAddress, configPath : String)(implicit ns : MetricNamespace) : Histogram = {
+    addToNamespace(address, configPath, None)
+  }
+
+  /**
+    * Create a Histogram with the following address.  Source the config from the provided Config object,
+    * instead of the MetricNamespace's Config
+    *
```
nsauro: New constructor which takes in a Config object.  I modified all `Collector` constructors so that both this constructor and the other one with no Config object, funnel into the same function.

colossus/src/main/scala/colossus/core/Server.scala
```diff
@@ -64,6 +65,32 @@ case class ServerSettings(
...
+    import colossus.metrics.ConfigHelpers._
+
+    //one-offing these, since they are going away soon
+    val bindingAttemptInterval = config.getFiniteDuration("binding-attempt-interval")
+    val bindingAttemptMaxTries = config.getLongOption("binding-attempt-max-tries")
+    val binding = PollingDuration(bindingAttemptInterval, bindingAttemptMaxTries)
+    val delegatorCreationInterval = config.getFiniteDuration("delegator-creation-interval")
+    val delegatorCreationMaxTries = config.getLongOption("delegator-creation-max-tries")
+    val delegator = PollingDuration(delegatorCreationInterval, delegatorCreationMaxTries)
+
```
nsauro: Without bringing in a third party lib, currying the constructor and just chaining applies seemed to be the cleanest and most readable way of doing this.

colossus/src/main/scala/colossus/core/Server.scala
```diff
@@ -182,15 +212,31 @@ private[colossus] class Server(io: IOSystem, config: ServerConfig, stateAgent :
   val address = new InetSocketAddress(settings.port)
   ssc.register( selector, SelectionKey.OP_ACCEPT )
 
-  val me = ServerRef(config, self, io, stateAgent)
+  val me = ServerRef(serverConfig, self, io, stateAgent)
```
nsauro: Some of the ugly I was talking about

colossus/src/main/scala/colossus/core/ServerDSL.scala
```diff
@@ -37,30 +39,173 @@ class DSLDelegator(server : ServerRef, _worker : WorkerRef, initializer: Initial
...
+    * "colossus.server" configuration path.
+    *
+    * @param name Name of this Server. Name is also the MetricAddress relative to the containing IOSystem's MetricNamespace
+    * @param port Port to run on
+    * @param initializer
+    * @param io
+    * @return
+    */
+  def start(name: String, port: Int)(initializer: WorkerRef => Initializer)(implicit io: IOSystem): ServerRef = {
+
```
nsauro: Feels a bit sneaky, but also allowed me to continue to have these functions all chain into each other

nsauro:
closing this.  After long discussions, I think we are changing the approach a bit.  Reopening with updates in a day or 2

PR #375, opened 2016-04-21
Server typesafe config

nsauro:
Take 2.

In this PR:

##  MetricSystem changes
MetricSystem's configuration still comes from `colossus.metrics` however, this definition has been updated so that values that configure how the `MetricSystem` behaves are now moved into `colossus.metrics.system`.  Any value residing within `colossus.metrics` is expected to be a `Metric` definition.

Removed the ability for user overrides to come from a location other than `colossus.metrics`.  This just helps simplify a lot of things.  If a user wants to override behavior, or add definitions to a `MetricSystem`, they will need to just override `colossus.metrics` in their application.conf.  Note, that the `MetricSystem` still has a constructor which takes in a `Config` object, which should allow users to load their Config however they want.

## IOSystem changes
IOSystem configuration defaults now come from `colossus.defaults.io-system`.  Users can apply overrides by putting in their application.conf:
```javascript
colossus.io-system.myIOSystem{
...
}
```
This definition then gets applied on top of the reference defaults.

IOSystem's constructors have also been simplified and reduced a bit.

Worker metrics are now configurable as well.

## Server changes

Servers now too are configurable via typesafe config.  It works exactly like IOSystem, with regards to how and where defaults are stored, and user overrides applied.  The only difference being that defaults are stored in `colossus.defaults.server` and user overrides are expected to be in `colossus.server.$serverName`

## Colossus' Metric definitions.

Colossus components have a lot of metrics, which we'd like to open up and make configurable.  One of the biggest problems we had in the last attempt, was configuring these metrics.  It was tricky as metrics are usually not created at the same time or place of its containing component(ie: IOSystem, Server).  This was causing some pain and hoop jumping in order to bubble down configuration to these call sites.

This PR, scrapped all of that.  Instead, we are now going define ALL metrics that Colossus components collect in `colossus.defaults.metrics`.  The `IOSystem`, at time of its creation, will take this definition, and overlay on on the `colossus.metrics` definition, and use that merged configuration to then create its `MetricSystem`.  This `MetricSystem` is then the system used throughout Colossus to create metrics, and as such, it should contain all of the metrics' definitions.

## Creating / overriding Metric definitions

given the following snippet:
```scala
//for brevity
implicit val ns = MetricNamespace("/prefix")
val rateA = Rate("myrate")
val rateB = Rate("myOtherRate, "pruned")
```
RateA's configuration sources are:
- `colossus.metrics.prefix.myrate`  -
- `colossus.metrics.system.collectors.defaults.rate` - Note overriding this overrides ALL rate defaults.

RateB's configuration sources are:
- `colossus.metrics.prefix.myOtherRate`
- `colossus.metrics.pruned` - This is a metric definition that can be shared amongst different metrics.  Overriding this overrides it for all users of this definition.
- `colossus.metrics.system.collectors.defaults.rate` - Note overriding this overrides ALL rate defaults.

### Overriding Colossus' metrics

There are 2 ways to override Colossus' metrics:

by address:
`colossus.metrics.$iosystem.worker.connections` - where $ioSystem is the name of a created IOSystem.

`colossus.defaults.metrics.worker-connections` - Overriding the default will override it for all metric instances (ie if there are multiple IOSystems)

## /etc   (might change in this PR)

The metric namespace hierarchy that existed between `MetricSystem`, `IOSystem` and `Server` has changed. `MetricSystem` has a global namespace(defaulted to `/` which is applied to ALL metrics created by it(do we still want this?)  However, IOSystem does not stack its name on top of Server metrics anymore.  `MetricNamespace`, while still used, could be removed, but I think it still has a purpose(though we are not using its `/` function to create subspaces anymore).

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
@@ -97,8 +97,9 @@ private[metrics] class CollectionMap[T] {
 
 class DuplicateMetricException(message: String) extends Exception(message)
 
-class Collection(val config: CollectorConfig,
-                 val collectors : ConcurrentHashMap[MetricAddress, Collector] = new ConcurrentHashMap[MetricAddress, Collector]()) {
+class Collection(val config: CollectorConfig) {
+
+  val collectors : ConcurrentHashMap[MetricAddress, Collector] = new ConcurrentHashMap[MetricAddress, Collector]()
 
   /**
    * Retrieve a collector of a specific type by address, creating a new one if
@@ -143,7 +144,6 @@ class Collection(val config: CollectorConfig,
 
 object Collection{
   def withReferenceConf(intervals : Seq[FiniteDuration]) : Collection = {
-    new Collection(CollectorConfig(intervals, ConfigFactory.defaultReference().getConfig(MetricSystem.ConfigRoot)),
-      new ConcurrentHashMap[MetricAddress, Collector]())
+    new Collection(CollectorConfig(intervals, ConfigFactory.defaultReference().getConfig(MetricSystem.ConfigRoot)))
   }
 }
--- a/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
@@ -21,7 +21,7 @@ import scala.reflect.ClassTag
  * @param intervals The aggregation intervals configured for the MetricSystem this collection belongs to
  * @param config The Configuration of the underlying [[MetricSystem]]
  */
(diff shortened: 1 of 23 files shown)


colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -146,3 +141,9 @@ class Collection(val config: CollectorConfig) {
   }
 
 }
+
+object Collection{
+  def withReferenceConf(intervals : Seq[FiniteDuration]) : Collection = {
```
DanSimon: It looks like this is only used in tests?  Might be a good idea either to add a comment about what it's used for, or just move it into tests.

DanSimon:
Looks good.  I suspect we'll be iterating on this over the next couple versions, but this is definitely a good start on all this. 👍

PR #376, opened 2016-04-25
Master merge into develop, fix RetryPolicy

DanSimon:
This is simply merging master into develop, however I had to make some fairly significant changes since `RetryPolicy` was not in develop yet.  So most of this is just from master except for the config parsing code in `RetryPolicy` and the related changes in `Server`.

I also moved the config helpers in metrics into their own file.

side-note - I wonder if there's a better way to handle situations like this.


nsauro:
LGTM

PR #380, opened 2016-05-03
making error and errorHander lazy vals to avoid init NPE's

### LINKED ISSUE #378: Investigate using lazy val for fullHandler in Service
Right now the `Service` class internally defines 

```scala
private val handler: PartialHandler[C] = handle orElse unhandled
```
which combines the user's `handle` with the partial function for handling requests that fall through.  But because of initialization order, if the user defines `handle` as a val, or uses any vals in their handle, it results in null-pointer exceptions.

Making this a lazy val will solve the issue, since it won't get called until initialization is complete.  However, since this is hot code in the request path, we should verify with thorough benchmarking that there isn't a significant performance hit.  If there is, we'll need to come with with another solution, possibly with API changes.

DanSimon:
Fixes #378 .  This will solve any silly initialization errors that can arise when extending this class and using vals for/in handlers or error handlers.  I did some benchmarking and there was no difference in performance whatsoever.

### DIFF
--- a/colossus/src/main/scala/colossus/service/ServiceDSL.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceDSL.scala
@@ -92,8 +92,8 @@ extends ServiceServer[C#Input, C#Output](codec, config, srv) {
     throw new ReceiveException("No sender")
   }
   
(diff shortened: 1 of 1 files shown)


DanSimon:
opened this against the wrong branch, closing

PR #382, opened 2016-05-04
Http typed headers

### LINKED ISSUE #260: Separate core http headers from other headers during parsing
As mentioned in #259 , we need to have dedicated values in the `HttpHeader` case class for headers like `Content-Length` and `Connection` that Colossus internally uses to make decisions during parsing/request processing.  Doing so will make all of the logic cleaner as well as avoid repeated scans of all the http headers.

DanSimon:
Fixes #260.  This was already sorta fixed with the new parsers introduced in 0.7, but now it's cleaned up and exposed to the public.

For general http headers that colossus itself needs to check, which currently are `content-length`, `transfer-encoding`, and `connection`, we now efficiently parse and identify during processing of the request/response.  The "structured" versions of these headers are kept separate from the rest of the headers, so when they are accessed we don't need to scan the full list of headers.


colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -196,9 +220,9 @@ class HttpHeaders(private val headers: JList[HttpHeader]) {
    */
   def contentLength: Option[Int] = firstValue(HttpHeaders.ContentLength).map{_.toInt}
 
-  def transferEncoding : TransferEncoding = firstValue(HttpHeaders.TransferEncoding).flatMap(TransferEncoding.unapply).getOrElse(TransferEncoding.Identity)
+  def transferEncoding : TransferEncoding = firstValue(HttpHeaders.TransferEncoding).map(TransferEncoding(_)).getOrElse(TransferEncoding.Identity)
 
-  def connection: Option[Connection] = firstValue(HttpHeaders.Connection).flatMap(Connection.unapply)
+  def connection: Option[Connection] = firstValue(HttpHeaders.Connection).map(Connection(_))
```
nsauro: one idea is to have firstValue take in those map functions as a parameter so that they can be directly applied to the collectFirst function.

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -296,8 +312,9 @@ object TransferEncoding {
   }
 
   private val all = Seq(Identity, Chunked)
-  def unapply(str : String) : Option[TransferEncoding] = {
-    all.find(_.value == str.toLowerCase)
+  def apply(str : String) : TransferEncoding = {
```
nsauro: lot of dupe code in these 3 functions, should be an easy way to dedupe it.

nsauro:
my comments are minor. 👍

PR #383, opened 2016-05-06
Parser cleanup

DanSimon:
Doing some overdue cleanup on the parsers:

* Changed several parsers from creating anonymous classes to concrete classes.  This is mostly to help with making stack traces more readable
* Removed the very outdated ParseBuffer classes.  Their functionality is now incorporated in the `bytes` and `bytesUntil` parsers, as well as the `FastArrayBuilder`.
* Standardizing on `Array[Byte]` for all byte-based parsers.  This is a little sad since `ByteString`s are very nice, but unfortunately significantly slower than raw byte arrays.  So I added `.asByteString` as an extension class to at least make the translation easier for situations when we do want to use ByteStrings.


nsauro:
👍

PR #385, opened 2016-05-09
Connection handler config

nsauro:
A few things here:

- ServiceConfig can now be loaded from configuration sources.  It's values are nested in `colossus.server.connection-handler`.
- Service metrics are added to `colossus.metrics`, so that they can be loaded at runtime.
- `DataSize` got a bit of love:
  - Added the ability to create one from a String(needed for config loading)
  - Added an unapply for pattern matching
  - Added some implicit objects to make it smell a bit like Durations ie: `1.KB`.  This was there before, but each invocation would create multiple instances.  This can be backed out if there is any strong resistance.  (Mainly b/c after I did this, I realized that it already existed by way of the existing implicit and instance functions 🙈  )

Note:  There is now naming ambiguity in this branch. ConnectionHandlers are called, well, ConnectionHandlers and/or Service.  This PR doesn't help this at all.  We should make a full name change on this pretty quickly to clean up this ambiguity.  I'm fine with doing in this branch, or saving it for another.

### DIFF
--- a/colossus/src/main/scala/colossus/util/Parsing.scala
+++ b/colossus/src/main/scala/colossus/util/Parsing.scala
@@ -14,28 +14,16 @@ case object Complete extends ParseStatus
 
 class ParseException(message: String) extends Exception(message)
 
-case class DataSize(value: Long) extends AnyVal {
-  def MB: DataSize = DataSize(value * 1024 * 1024)
-  def KB: DataSize = DataSize(value * 1024)
-  def bytes = value
+case class DataSize(bytes: Long) extends AnyVal {
+  def megabytes = MB
+  def MB: DataSize = DataSize(bytes * 1024 * 1024)
+  def kilobytes = KB
+  def KB: DataSize = DataSize(bytes * 1024)
 }
 
 object DataSize {
-  //implicit def longToDataSize(l: Long): DataSize = DataSize(l)
 
-  implicit class IntToSize(val amount : Int) extends DataSizeConversions[Int]
-  implicit class LongToSize(val amount : Long) extends DataSizeConversions[Long]
-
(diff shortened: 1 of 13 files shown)


colossus/src/main/scala/colossus/util/Parsing.scala
```diff
@@ -6,22 +6,54 @@ import core.DataBuffer
...
+
+  implicit class IntToSize(val amount : Int) extends DataSizeConversions[Int]
+  implicit class LongToSize(val amount : Long) extends DataSizeConversions[Long]
+
+  trait DataSizeConversions[T] {
+    def amount : T
+
+    def bytes = B
+    def B = DataSize(s"$amount B")
+    def MB = DataSize(s"$amount MB")
```
DanSimon: There's definitely some duplication of logic here.  I think it makes sense to just put all these methods in `DataSize` itself and get rid of the implicit classes.  Also I'm not a huge fan of these using string parsing in these methods since it's pretty inefficient.

colossus/src/main/scala/colossus/util/Parsing.scala
```diff
@@ -6,22 +6,42 @@ import core.DataBuffer
...
 sealed trait ParseStatus
 case object Incomplete extends ParseStatus
 case object Complete extends ParseStatus
 
 class ParseException(message: String) extends Exception(message)
 
-case class DataSize(value: Long) extends AnyVal {
-  def MB: DataSize = value * 1024 * 1024
-  def KB: DataSize = value * 1024
-  def bytes = value
```
DanSimon: We still need to keep this method, otherwise you can't do `5.bytes`.

DanSimon:
👍

PR #386, opened 2016-05-11
Websocket server improvements

DanSimon:
This makes some notable improvements to the API for starting and using a Websocket server.

* new sub-protocols package for defining websocket sub-protocols.  These work similarly to regular codecs, but are designed to parse the contents of Websocket frames, so no streaming support is needed.
* New API for starting a Websocket server.  The `WebsocketServer` properly encapsulates starting a HTTP server and handling the Websocket upgrade request.
* The Websocket server handler now properly handles the Ping and Close opcodes.  Support for the Continue code is still needed.
* Created a new `DataBlock` class that wraps a byte array.  This is similar to Akka's `ByteString`, but at least from previous benchmarking ByteString has some serious performance problems.  This is an attempt to get something with similar functionality without the overhead, though I haven't done much benchmarking on it yet.

This is still not a complete solution and I think there's probably more ways to improve this.  But we can take this incrementally and figure out what works and what doesn't.

### DIFF
--- a/colossus/src/main/scala/colossus/core/DataBuffer.scala
+++ b/colossus/src/main/scala/colossus/core/DataBuffer.scala
@@ -5,6 +5,45 @@ import akka.util.ByteString
 import java.nio.ByteBuffer
 import java.nio.channels.SocketChannel
 
+
+/**
+ * A zero-overhead abstraction over a byte array.  This offers much of the same
+ * functionally offered by Akka's ByteString, but without any additional
+ * overhead.  While this makes DataBlock a bit less powerful and flexible, it
+ * also makes it considerably faster.
+ *
+ * Note - unfortunately implementing `take` and `drop` here conflicted with the
+ * methods in arrayOps that is imported by default.  It's possible to "unimport"
+ * things defined in predef, but that seems like too much of a hassle for anyone
+ * who wants to use this.
+ */
+object DataBlock {
+  type DataBlock = Array[Byte]
+
+  implicit class DataBlockOps(val data: DataBlock) extends AnyVal {
+    
+    //def size = data.length
+    def byteString: ByteString = ByteString(data)
+    def utf8String = byteString.utf8String
+
+    /**
+     * Create a new DataBlock containing the data from `block` appended to the
+     * data in this block
+     */
+    def +:+(block: DataBlock): DataBlock = {
+      val concat = DataBlock(data.length + block.length)
+      System.arraycopy(data, 0, concat, 0, data.length)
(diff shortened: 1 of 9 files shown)


colossus/src/main/scala/colossus/core/DataBuffer.scala
```diff
@@ -5,6 +5,69 @@ import akka.util.ByteString
...
+ * also makes it considerably faster.
+ *
+ * TODO: This is possibly a contender for using in all places where Array[Byte]
+ * is used, but thorough benchmarking is needed.  It is also possible that the
+ * performance of ByteString has improved in later versions of Akka, so that
+ * should also be tested before expanding on this any more.
+ *
+ */
+
+case class DataBlock(data: Array[Byte]) {
```
nsauro: Tests for the few functions in here?

colossus/src/main/scala/colossus/protocols/websocket/Websocket.scala
```diff
@@ -196,11 +191,78 @@ abstract class WebsocketHandler[P <: Protocol](context: Context)
...
+  /**
+   * Start a Websocket server on the specified port.  Since Websocket
+   * connections are upgraded from HTTP connections, this will actually start an
+   * HTTP server and react to Websocket upgrade requests on the path
+   * `upgradePath`, all other paths will 404.
+   */
+  def start(name: String, port: Int, upgradePath: String = "/")(init: WorkerRef => WebsocketInitializer)(implicit io: IOSystem) = {
+    Server.start(name, port){worker => new Initializer(worker) {
+    
+      val websockinit = init(worker)
```
nsauro: Can we put a type annotation on this, not sure what it is from following the code

colossus/src/main/scala/colossus/protocols/websocket/Websocket.scala
```diff
@@ -196,11 +191,78 @@ abstract class WebsocketHandler[P <: Protocol](context: Context)
...
+
+      def onConnect = ctx => new HttpService(ServiceConfig.Default,ctx) {
+        def handle = {
+          case req @ UpgradeRequest(resp) if (req.head.path == upgradePath) => {
```
nsauro: An UpgradeRequest is only a valid UpgradeRequest if the path matches the upgradePath, right?  I have mixed feelings about doing stuff like this, but in this case what if UpgradeRequest was an instance object, that has the upradePath on it, and an unapply, which can then use that reference that upgradePath.  It gets rid of this guard case, and makes UpgradeRequest unapply more complete.

PR #388, opened 2016-05-11
ClientConfig now can be sourced from typesafe Config

nsauro:
First pass of making ClientConfig classes sourceable from Typesafe Config.

The idea was to give ClientConfig `load` functions, similar to `ServiceClient`.  These functions can be invoked directly or indirectly by way of `ClientFactory`

Client definitions are expected to appear in `colossus.client`, and are sourced by name.  Defaults are provided in `colossus.client.defaults`.  The only caveat here, is that, for obvious reasons, the defaults don't have an `address` or `name` field.

### DIFF
--- a/colossus/src/main/scala/colossus/service/ServiceDSL.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceDSL.scala
@@ -1,11 +1,12 @@
 package colossus
 package service
 
+import com.typesafe.config.{Config, ConfigFactory}
 import core._
 
 import akka.actor.ActorRef
 import scala.concurrent.duration._
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.Future
 import scala.language.higherKinds
 
 import java.net.InetSocketAddress
@@ -13,8 +14,6 @@ import metrics.MetricAddress
 
 import Codec._
 
-
-//TODO : Rename to Protocol
 trait Protocol {self =>
   type Input
   type Output
@@ -36,11 +35,13 @@ import Protocol._
 
 /**
  * Provide a Codec as well as some convenience functions for usage within in a Service.
- * @tparam C the type of codec this provider will supply
+  *
+  * @tparam C the type of codec this provider will supply
  */
 trait CodecProvider[C <: Protocol] {
   /**
     * The Codec which will be used.
+    *
     * @return
     */
   def provideCodec: ServerCodec[C#Input, C#Output]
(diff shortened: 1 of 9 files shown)


colossus/src/main/scala/colossus/service/ServiceDSL.scala
```diff
@@ -185,9 +151,30 @@ trait ClientLifter[C <: Protocol, T[M[_]] <: Sender[C,M]] {
...
 
 
+  /**
+    * Load a ServiceClient definition from a Config.  Looks into `colossus.clients.$clientName` and falls back onto
+    * `colossus.client-defaults`
+    * @param clientName The name of the client definition to load
+    * @param config A config object which contains at the least a `colossus.clients.$clientName` and a `colossus.client-defaults`
+    * @return
+    */
+  def apply(clientName : String, config : Config = ConfigFactory.load())(implicit provider: ClientCodecProvider[C], env: E) : T = {
```
DanSimon: I think it might be better to create a `ClientConfig.Default` and have `config`default to that, so that we're not loading config all the time.  That's what we do for `ServiceConfig`.

DanSimon: Oh right that's not going to work here since the config is path based and you need the whole config.  Still, I think we can probably do a `Config.load`once in the `ServiceClient` object and have some kind of default function that reads from there.

colossus/src/main/scala/colossus/service/ServiceDSL.scala
```diff
@@ -185,9 +151,30 @@ trait ClientLifter[C <: Protocol, T[M[_]] <: Sender[C,M]] {
...
+    * @return
+    */
+  def apply(config : Config)(implicit provider: ClientCodecProvider[C], env: E) : T = {
+    apply(ClientConfig.load(config))
+  }
+
   def apply(config: ClientConfig)(implicit provider: ClientCodecProvider[C], env: E): T
 
-  def apply(host: String, port: Int, requestTimeout: Duration = 1.second)(implicit provider: ClientCodecProvider[C], env: E): T = {
+  def apply(host: String, port: Int, requestTimeout: Duration)(implicit provider: ClientCodecProvider[C], env: E): T = {
```
DanSimon: I think I'd be more in favor of just dropping the `requestTimeout` param instead of making it mandatory.  I think in most cases if you're going to be configuring a timeout you're gonna want to pass in a whole config and set a name as well.  This constructor is really just for demos and hello-world-style stuff.

nsauro:
addressed comments

DanSimon:
👍

colossus/src/main/scala/colossus/service/ServiceDSL.scala
nsauro: Should I remove this parameter?

PR #389, opened 2016-05-12
fix open up spot when connection closes test

aliyakamercan:
@DanSimon

### DIFF
--- a/colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
@@ -243,8 +243,7 @@ class ServerSpec extends ColossusSpec {
         }
       }
 
-      //TODO : fix this test
-      "open up spot when connection closes" ignore {
+      "open up spot when connection closes" in {
         val settings = ServerSettings(
           port = TEST_PORT,
           maxConnections = 1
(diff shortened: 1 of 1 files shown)


DanSimon:
👍

colossus-tests/src/test/scala/colossus/core/ServerSpec.scala
```diff
@@ -255,16 +254,9 @@ class ServerSpec extends ColossusSpec {
           withServer(server) {
             val c1 = TestClient(server.system, TEST_PORT)
             expectConnections(server, 1)
-            val c2 = TestClient(server.system, TEST_PORT, waitForConnected = false, connectRetry = NoRetry)
```
nsauro: Shouldn't this test still be maintaining that the Server rejects connections when it is at capacity?

nsauro: Hrm, maybe not, I just saw the test above this tests for that.

nsauro:
👍

DanSimon:
Can you add a brief description to the PR as well as a link to the issue.

I'm not entirely sure this fixes the test, as the test should have worked before as well, but this at least simplifies it so if it starts having problems again it should be easier to diagnose.

PR #390, opened 2016-05-17
WorkerManager IdleCheck now waits for acks 

### LINKED ISSUE #265: Don't use scheduler.schedule for worker idle checks
Right now the worker manager periodically broadcasts `CheckIdleConnections` messages to all workers to trigger the periodic idle check.  However, if the workers are under heavy load and are unable to process the messages fast enough, the manager will keep spamming them with more messages.

We should switch this to use `scheduler.scheduleOnce` and also have the worker manager not send the next message until it receives acks from all the workers.

nsauro:
This is to fix #265 .  

WorkerManager now doesn't use `context.schedule` it uses `context.scheduleOnce`.  After Workers process the `CheckIdleConnections`, they respond with a 'IdleCheckExecuted`.  Once the WorkerManager receives those acks from all Workers, it will only then schedule another idle check.

Note, this is branched off of the service_client_config branch.  So, I can merge this directly into that, or wait until that branch is approved and merged into develop, then rebase this off of develop.  Either or.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/core/WorkerManagerSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/core/WorkerManagerSpec.scala
@@ -1,0 +1,66 @@
+package colossus.core
+
+import akka.actor.{ActorContext, ActorRef, Props}
+import akka.agent.Agent
+import akka.testkit.TestProbe
+import colossus.IOSystem
+import colossus.testkit.ColossusSpec
+import org.scalatest.concurrent.Eventually
+
+import scala.collection.immutable.IndexedSeq
+import scala.concurrent.duration._
+
+class WorkerManagerSpec extends ColossusSpec with Eventually{
+
+  "WorkerManager" must {
+
+    "send an idle only after all workers ack the previous idle check" in {
+      withIOSystem{ io =>
+        //we need to create a separate manager.
+        implicit val ec = system.dispatcher
+        //need to use the same number of ioSystem workers, since WorkerManager uses this to control the # of workers created
+        val probes: IndexedSeq[TestProbe] = (1 to io.numWorkers).map(_ => TestProbe())
+        val probesIter = probes.iterator
+        var workerRefs : Seq[WorkerRef] = Seq()
+
+        //create a "Worker" which is just a probe.
+        val workerFact = new WorkerFactory {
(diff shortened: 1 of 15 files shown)


colossus-tests/src/test/scala/colossus/core/WorkerManagerSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+import colossus.IOSystem
+import colossus.testkit.ColossusSpec
+import org.scalatest.concurrent.Eventually
+
+import scala.collection.immutable.IndexedSeq
+import scala.concurrent.duration._
+
+class WorkerManagerSpec extends ColossusSpec with Eventually{
+
+  "WorkerManager" must {
```
nsauro: Sorry for the tediousness, but this actually was a bit of a pain to test :)

colossus/src/main/scala/colossus/core/WorkerManager.scala
```diff
@@ -24,7 +24,7 @@ import scala.util.{Failure, Success}
  * @param workerAgent WorkerRefs that this WorkerManager manages
  * @param ioSystem Containing IOSystem
  */
-private[colossus] class WorkerManager(workerAgent: Agent[IndexedSeq[WorkerRef]], ioSystem: IOSystem)
+private[colossus] class WorkerManager(workerAgent: Agent[IndexedSeq[WorkerRef]], ioSystem: IOSystem, workerFactory : WorkerFactory )
```
nsauro: workerFactory had to be injected now, since tests needed a test impl which allowed for Workers to be modeled by TestProbes.

colossus/src/main/scala/colossus/core/WorkerManager.scala
```diff
@@ -219,6 +219,22 @@ extends Actor with ActorLogging with Stash {
   }
 }
 
+private[colossus] trait WorkerFactory {
+  def createWorker(id: Int, ioSystem: IOSystem, context: ActorContext): ActorRef
+}
+
+private[colossus] object DefaultWorkerFactory extends WorkerFactory{
```
nsauro: zee original code which was in WorkerManager for creating Workers.  Only moved, not changed.

DanSimon:
👍

nsauro:
reopening this in a second, against develop

PR #391, opened 2016-05-20
WorkerManager IdleCheck now waits for acks 

### LINKED ISSUE #265: Don't use scheduler.schedule for worker idle checks

nsauro:
This is to fix #265 .  

WorkerManager now doesn't use `context.schedule` it uses `context.scheduleOnce`.  After Workers process the `CheckIdleConnections`, they respond with a 'IdleCheckExecuted`.  Once the WorkerManager receives those acks from all Workers, it will only then schedule another idle check.

Note, this is branched off of the service_client_config branch.  So, I can merge this directly into that, or wait until that branch is approved and merged into develop, then rebase this off of develop.  Either or.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/core/WorkerItemSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/core/WorkerItemSpec.scala
@@ -12,7 +12,7 @@ class WorkerItemSpec extends ColossusSpec {
 
   "A WorkerItem" must {
     "bind to a worker" in {
-      withIOSystem{io => 
+      withIOSystem{io =>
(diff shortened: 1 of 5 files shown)


DanSimon:
👍

Issue #373, opened 2016-04-20
AsyncServiceClient appears to be trying to bind twice


DanSimon:
Seeing this a bunch in tests: `attempted to bind worker colossus.service.AsyncHandlerGenerator$AsyncHandler@157bf742 that was already bound`

nsauro:
close this?

PR #393, opened 2016-05-23
bind to worker when worker constructor is used

aliyakamercan:
Fixes https://github.com/tumblr/colossus/issues/373

Bind to worker only if proper constructor is used as notice says

@DanSimon


DanSimon:
👍

nsauro:
👍

Issue #379, opened 2016-05-03
Add global tags in metric namespaces


DanSimon:
Pre 0.7, you used to be able to define an implicit val of global tags that would be automatically added to all metrics in scope.  This was removed in 0.7 when we overhauled the metrics, but we should add equivalent functionality back in with 0.8.x.

This should probably be done by simply adding a `tags` field to `MetricNamespace`.  If users want to set additional tags, they can just create a new namespace and use that one implicitly.

These tags need to be only added to metrics when snapshots are created, which means we might have to change the relationship between collectors and namespaces.

PR #394, opened 2016-05-24
Collection trie

aliyakamercan:
solves https://github.com/tumblr/colossus/issues/379

Collection as a trie of collectors keyed with metric address. 
MetricContexts can also be inserted, tags of the metric contexts would be applied to whole trie below that context.

@DanSimon

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
@@ -99,7 +99,8 @@ class DuplicateMetricException(message: String) extends Exception(message)
 
 class Collection(val config: CollectorConfig) {
 
-  val collectors : ConcurrentHashMap[MetricAddress, Collector] = new ConcurrentHashMap[MetricAddress, Collector]()
+  import Collection.TaggedCollector
+  val collectors : ConcurrentHashMap[MetricAddress, TaggedCollector] = new ConcurrentHashMap[MetricAddress, TaggedCollector]()
 
   /**
    * Retrieve a collector of a specific type by address, creating a new one if
@@ -108,22 +109,21 @@ class Collection(val config: CollectorConfig) {
    * @param address Address meant to be relative to this MetricNamespace's namespace
    * @param f Function which takes in an absolutely pathed MetricAddress, and a [[CollectorConfig]] and returns an instance of a [[Collector]]
     */
-  def getOrAdd[T <: Collector : ClassTag](address : MetricAddress)(f : (MetricAddress, CollectorConfig) => T): T = {
+  def getOrAdd[T <: Collector : ClassTag](address : MetricAddress, tags: TagMap)(f : (MetricAddress, CollectorConfig) => T): T = {
     def cast(retrieved: Collector): T = retrieved match {
       case t : T => t
-      case other => {
+      case other =>
         throw new DuplicateMetricException(
           s"An event collector with address $address of type ${other.getClass.getSimpleName} already exists"
         )
-      }
     }
     if (collectors.containsKey(address)) {
-      cast(collectors.get(address))
+      cast(collectors.get(address).collector)
     } else {
       val c = f(address, config)
-      collectors.putIfAbsent(address, c) match {
+      collectors.putIfAbsent(address, TaggedCollector(c, tags)) match {
         case null => c
-        case other => cast(other)
+        case other => cast(other.collector)
       }
     }
   }
@@ -134,7 +134,9 @@ class Collection(val config: CollectorConfig) {
     while (keys.hasMoreElements) {
       val key = keys.nextElement
       Option(collectors.get(key)) foreach { c =>
-        build = build ++ c.tick(interval)
+        build = build ++ c.collector.tick(interval).mapValues { valueMap =>
+          valueMap.map { case (t, tValue) => (t ++ c.tagMap, tValue) }
+        }
       }
     }
     build
@@ -148,4 +150,6 @@ object Collection{
   def withReferenceConf(intervals : Seq[FiniteDuration]) : Collection = {
     new Collection(CollectorConfig(intervals, ConfigFactory.defaultReference().getConfig(MetricSystem.ConfigRoot)))
   }
+
+  case class TaggedCollector(collector: Collector, tagMap: TagMap)
 }
--- a/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
@@ -30,9 +30,14 @@ class MetricInterval private[metrics](val namespace : MetricAddress,
 
 
 /**
- * A MetricNamespace is essentially just an address prefix and is needed when
- * getting or creating collectors.  The `namespace` address is prefixed onto the
- * given address for the collector to create the full address.
+ * A MetricNamespace is essentially just an address prefix and set of tags. It is needed when
+ * getting or creating collectors. The `namespace` address is prefixed onto the
+ * given address for the collector to create the full address. Tags are added to each collector under under this
+ * context.
+ *
+ * {{{
+ * val subnameSpace: MetricContext = namespace / "foo" * ("a" -> "b")
+ * }}}
  */
 trait MetricNamespace {
 
@@ -42,8 +47,22 @@ trait MetricNamespace {
     */
   def namespace: MetricAddress
 
+  /**
+    * Tags applied to all collectors in this namespace
+    * @return
+    */
+  def tags: TagMap
 
   protected def collection : Collection
+
+  /**
+    * Add tags to context
+    * @param tags TagMap
+    * @return Returns a new context with supplied tags added
+    */
+  def *(tags: (String, String)*): MetricContext = MetricContext(namespace, collection, this.tags ++ tags)
+
+  def withTags(tags: (String, String)*): MetricContext = *(tags:_*)
 
   /**
     * Retrieve a [[Collector]] of a specific type by address, creating a new one if
@@ -56,17 +75,17 @@ trait MetricNamespace {
     */
   def getOrAdd[T <: Collector : ClassTag](address : MetricAddress)(f : (MetricAddress, CollectorConfig) => T) : T = {
     val fullAddress =  namespace / address
-    collection.getOrAdd(fullAddress)(f)
+    collection.getOrAdd(fullAddress, tags)(f)
   }
 
   /**
    * Get a new namespace by appending `subpath` to the namespace
    */
-  def /(subpath: MetricAddress): MetricNamespace = MetricContext(namespace / subpath, collection)
+  def /(subpath: MetricAddress): MetricContext = MetricContext(namespace / subpath, collection, tags)
 
 }
 
-case class MetricContext(namespace: MetricAddress, collection: Collection) extends MetricNamespace
+case class MetricContext(namespace: MetricAddress, collection: Collection, tags: TagMap = TagMap.Empty) extends MetricNamespace
 
 /**
   * The MetricSystem is a set of actors which handle the background operations of dealing with metrics. In most cases,
@@ -103,6 +122,9 @@ case class MetricContext(namespace: MetricAddress, collection: Collection) exten
  */
 case class MetricSystem private[metrics] (namespace: MetricAddress, collectionIntervals : Map[FiniteDuration, MetricInterval],
                                           collectionSystemMetrics : Boolean, config : Config) extends MetricNamespace {
+
+  private val localHostname = java.net.InetAddress.getLocalHost.getHostName
+  val tags: TagMap = Map("host" -> localHostname)
 
   val collection = new Collection(CollectorConfig(collectionIntervals.keys.toSeq, config))
   registerCollection(collection)
@@ -165,7 +187,6 @@ object MetricSystem {
     * Create a new MetricSystem, using the specified configuration
     *
     * @param config Config object expected to be in the shape of the reference.conf's `colossus.metrics` definition.
-    * @param system
     * @return
     */
   def apply(config : Config = loadDefaultConfig())(implicit system : ActorSystem) : MetricSystem = {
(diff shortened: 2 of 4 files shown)


nsauro:
tests ? :)

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -97,18 +102,21 @@ private[metrics] class CollectionMap[T] {
...
 class DuplicateMetricException(message: String) extends Exception(message)
 
-class Collection(val config: CollectorConfig) {
+class Collection(val config: CollectorConfig, var tags: TagMap = TagMap.Empty)
+  extends mutable.Map[MetricAddress, Collector]
+  with mutable.MapLike[MetricAddress, Collector, Collection] {
 
-  val collectors : ConcurrentHashMap[MetricAddress, Collector] = new ConcurrentHashMap[MetricAddress, Collector]()
+  val suffixes: ConcurrentHashMap[MetricAddress, Collection] = new ConcurrentHashMap[MetricAddress, Collection]()
+  var collector: Option[Collector] = None
```
nsauro: What's this for?

colossus-metrics/src/test/scala/colossus/metrics/CollectionSpec.scala
```diff
@@ -9,14 +9,14 @@ class CollectionSpec extends WordSpec with MustMatchers{
...
-    implicit val namespace = MetricContext("/", collection)
+    implicit val namespace: MetricContext = MetricContext("/", collection)
 
     "getOrAdd should add a new Collector" in {
       //note, this constructor implicitly adds its constructed object into this collection
       val foo = MetricAddress("foo")
       val rate = Rate(foo, false, true)
 
-      val x = namespace.collection.collectors.get(foo)
+      val x = namespace.collection.get(foo).get
```
nsauro: For this,and other instances of where you are calling `get` directly, ScalaTest does have at `OptionValues` trait which provides a  `value` function. This behaves like `get' but actually provides a more appropriate error message if you are `get`ing a None (vs NPE)

colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
```diff
@@ -104,6 +106,8 @@ case class MetricContext(namespace: MetricAddress, collection: Collection) exten
 case class MetricSystem private[metrics] (namespace: MetricAddress, collectionIntervals : Map[FiniteDuration, MetricInterval],
                                           collectionSystemMetrics : Boolean, config : Config) extends MetricNamespace {
 
+  private val localHostname = java.net.InetAddress.getLocalHost.getHostName
+  val extraTags: TagMap = Map("host" -> localHostname)
```
DanSimon: This needs to be optional.  In the StatsReporter there's a `includeHostnameInGlobalTags` option.  I'm ok with merging this PR without this change, but an issue needs to be created first.

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -97,18 +97,22 @@ private[metrics] class CollectionMap[T] {
...
 
-class Collection(val config: CollectorConfig) {
-
-  val collectors : ConcurrentHashMap[MetricAddress, Collector] = new ConcurrentHashMap[MetricAddress, Collector]()
+class Collection(val config: CollectorConfig)
+  extends mutable.Map[MetricAddress, Collector]
+  with mutable.MapLike[MetricAddress, Collector, Collection] {
 
+  val suffixes: ConcurrentHashMap[MetricAddress, Collection] = new ConcurrentHashMap[MetricAddress, Collection]()
+  var collector: Option[Collector] = None
```
DanSimon: If I understand correctly, this needs to be a var because you could create a metric called `/a/b/c`, which creates a 3-node trie.  Then you create a metric `/a/b` so we need to update the `b`node with a collector.  This...seems ok to me except that this has to be an `AtomicReference` since right now it's not threadsafe.  Also this should definitely not be a public var.

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -117,29 +121,72 @@ class Collection(val config: CollectorConfig) {
         )
       }
     }
-    if (collectors.containsKey(address)) {
-      cast(collectors.get(address))
-    } else {
+    addContext(ns)
```
DanSimon: If a metric already exists at the address, does this end up overwriting whatever tags are set without changing the collector?

It might make more sense to have a single var (which will be an atomic reference) that stores a tuple of a collector and a tagmap, and both are set at the same time.

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -97,18 +97,22 @@ private[metrics] class CollectionMap[T] {
 
 class DuplicateMetricException(message: String) extends Exception(message)
 
-class Collection(val config: CollectorConfig) {
-
-  val collectors : ConcurrentHashMap[MetricAddress, Collector] = new ConcurrentHashMap[MetricAddress, Collector]()
+class Collection(val config: CollectorConfig)
```
DanSimon: This class definitely needs some docs now explaining what it is, how it works, where it's used, and also why we've chosen to use a trie structure.

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -117,29 +121,72 @@ class Collection(val config: CollectorConfig) {
...
+  def withPrefix(s: MetricAddress): Collection =
+    if (s.size == 0) {
+      this
+    } else {
+      val leading = MetricAddress(s.head)
+      suffixes.putIfAbsent(leading, empty)
+      suffixes.get(leading).withPrefix(s.tail)
+    }
+  
+  def addContext(ns: MetricNamespace) = withPrefix(ns.namespace).tags = ns.extraTags
```
nsauro: If a user has 2 namespaces, with different tags, then the last one in wins?

DanSimon: Well I think that would only happen if a user tries to create two collectors with the same address, which is related to my comment above.

nsauro: Ah..duh, yea same thing

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -117,29 +121,72 @@ class Collection(val config: CollectorConfig) {
...
+    } else {
+      if(suffixes.containsKey(MetricAddress(key.head))) {
+        suffixes.get(MetricAddress(key.head)).get(key.tail)
+      } else {
+        None
+      }
+    }
+  }
+
+  override def iterator: Iterator[(MetricAddress, Collector)] = (for (v <- collector.iterator) yield (MetricAddress.Root, v)) ++
```
nsauro: is MetricAddress.Root right here?  Shouldn't it be the "last" element in the Collector's MetricAddress?

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -117,29 +121,72 @@ class Collection(val config: CollectorConfig) {
...
+    (for (
+      (component, m) <- suffixes.iterator;
+      (s, v) <- m.iterator
+    ) yield (component / s, v))
+
+  def iteratorWithTags: Iterator[(MetricAddress, TagMap, Collector)] = (for (v <- collector.iterator) yield (MetricAddress.Root, tags, v)) ++
+    (for (
+      (component, m) <- suffixes.iterator;
+      (s, t, v) <- m.iteratorWithTags
+    ) yield (component / s, t ++ tags, v))
```
nsauro: If we are always using `Root`, then this won't actually rebuild the address, right?  I think `Collection` needs to either store the tail element of the Collector's `MetricAddress` or pull that out of the Collector every time.   That's IF I understand what's going on here

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -117,29 +121,72 @@ class Collection(val config: CollectorConfig) {
...
   }
 
   def tick(interval: FiniteDuration): MetricMap = {
-    val keys = collectors.keys
     var build: MetricMap = Map()
-    while (keys.hasMoreElements) {
-      val key = keys.nextElement
-      Option(collectors.get(key)) foreach { c =>
-        build = build ++ c.tick(interval)
+    iteratorWithTags.foreach{ case (address, extraTags, c) =>
```
nsauro: Also, it looks like address isn't even really used, so why are we bothering to generate it in the iterate* functions below?  Would simplify them,and nullify my comments about using `Root`

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
aliyakamercan: `MetricAddress.Root` is correct here, what is returned here is appended to end of previous call creating the whole metric address.

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
aliyakamercan: same as above.

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
aliyakamercan: good point :)

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
aliyakamercan: This would only happen if you create a metric context with an address of existing collector, I am not sure how that should be handled. I think we should throw an exception in that case instead of updating tags.

colossus-metrics/src/main/scala/colossus/metrics/MetricSystem.scala
aliyakamercan: created https://github.com/tumblr/colossus/issues/397

aliyakamercan:
👊

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
nsauro: Should we just remove that address building code then?

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
nsauro: At the very least, log something, but yea it is a bit of an edgy case.   I'm not strongly opinionated on any 3(unless we overwrite, then we should log something at debug? level)

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
DanSimon: I agree if it's not being used we can just remove the address building and have the iterator return a 2-tuple.

colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -117,29 +134,89 @@ class Collection(val config: CollectorConfig) {
...
+  final def withPrefix(s: MetricAddress): Collection =
+    if (s.size == 0) {
+      this
+    } else {
+      val leading = MetricAddress(s.head)
+      suffixes.putIfAbsent(leading, empty)
+      suffixes.get(leading).withPrefix(s.tail)
+    }
+  
+  private def addContext(ns: MetricNamespace) = {
```
DanSimon: This method should be type annotated, or just remove the =

DanSimon:
So after spending some time playing around with this branch, I'm now kind of convinced we don't need a collection trie.

First, let's restate the overall goals we want out of this change:
* We should be able to easily apply tags to groups of collectors based on address.  So I should  be able to add tags to the address `/foo/bar`, and every collector with address `/foo/bar/...` will have those tags applied to all their metrics
* Tags accumulate down a metric address.  So if I add tags to `/foo` and to `/foo/bar/`, any collectors under `/foo/bar` get both sets of tags added.
* The full path and extra tags of a collector are decided based on the implicit `MetricNamespace` in scope.  So if a namespace with path `/foo/bar` and tags `biz -> buz` is in scope when I do `val rate = Rate("/baz")`, then the end result is a rate `/foo/bar/baz` with the extra tags applied.  Tags from the namespace's parents also get applied, all the way back to the root `MetricSystem`.
* We still want to prevent two collectors from having the same full address, but there shouldn't be any problem with having two different metric namespaces with the same path and different tags.

The thing is now we actually have two trie structures.  `Collection` is a trie, but also so is `MetricNamespace`.  It seems to me now that we're storing the "effictive" tagmap with the collector that the trie structure of `Collection` is redundant.  So I think what we should do is:

* Revert `Collection` back to a single hashmap, but of type `MetricAddress -> V`, though maybe make the name of that case class a little more descriptive.
* Give `MetricNamespace` a `parent` field of type `Option[MetricNamespace]`
* Have the `/` operator's return type be `MetricContext` instead of `MetricNamespace` and add a `withTags` method, so users can do `val namespace = system / "foo" withTags ("tag" -> "value") / "bar"`
* the `getOrAdd` method on `MetricNamespace` will build the set of extra tags by traversing itself back to the root using `parent` to accumulate the extra tags of all its parents, so the `TagMap` that gets put in the `V` is the full set of extra tags

I think this gets us the end result we want, the ability for the user to do something like:

```scala
val system = MetricSystem()

val ns1 = system / "foo" withTags ("a" -> "b")
val ns2 = system / "foo" withTags ("a" -> "c") / "bar"

def makeRate(ns: MetricNamespace): Rate = {
  implicit val mynamespace = ns withTags ("b" -> "c")
  Rate("baz")
}

val r1 = makeRate(ns1) //rate with address "/foo/baz" and tags (a -> b, b -> c)
val r2 = makeRate(ns2) //rate with address "/foo/bar/baz" and tags (a -> c, b -> c)
```

Does this make sense?  Is there something I'm missing?

aliyakamercan:
That would break integrity, you can tag same cascading subsections differently with a structure like that, I am not sure if that would create a problem or not.

aliyakamercan:
Closing this one, I will open a new new one with different approach

PR #395, opened 2016-05-24
fixing TagDecorator not being applied to ServiceClient metrics

### LINKED ISSUE #290: ServiceClient does not use the provided TagDecorator when reporting metrics
Internally, it formulates `hTags` and `hpTags` and uses those instead. 

```scala
private val hpTags: TagMap = Map("client_host" -> address.getHostName, "client_port" -> address.getPort.toString)
private val hTags: TagMap = Map("client_host" -> address.getHostName)
```

nsauro:
fixes #290


DanSimon:
👍

PR #399, opened 2016-06-08
tag all collecotrs under metric context

aliyakamercan:
@DanSimon 

solves https://github.com/tumblr/colossus/issues/379

Apply tags on metric context to collectors created with it. Metric context inherit tags from their parents, extra tags can be added on contexts using `*` or `withTags`.

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
@@ -112,10 +112,11 @@ class Collection(val config: CollectorConfig) {
   def getOrAdd[T <: Collector : ClassTag](address : MetricAddress, tags: TagMap)(f : (MetricAddress, CollectorConfig) => T): T = {
     def cast(retrieved: Collector): T = retrieved match {
       case t : T => t
-      case other =>
+      case other => {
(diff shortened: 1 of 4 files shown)


colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
```diff
@@ -108,22 +109,21 @@ class Collection(val config: CollectorConfig) {
    * @param address Address meant to be relative to this MetricNamespace's namespace
    * @param f Function which takes in an absolutely pathed MetricAddress, and a [[CollectorConfig]] and returns an instance of a [[Collector]]
     */
-  def getOrAdd[T <: Collector : ClassTag](address : MetricAddress)(f : (MetricAddress, CollectorConfig) => T): T = {
+  def getOrAdd[T <: Collector : ClassTag](address : MetricAddress, tags: TagMap)(f : (MetricAddress, CollectorConfig) => T): T = {
     def cast(retrieved: Collector): T = retrieved match {
       case t : T => t
-      case other => {
```
DanSimon: let's add the surrounding braces back here, keep it consistent with the rest of the codebase.

DanSimon:
👍

PR #400, opened 2016-06-09
ServiceConfig now based on server name, caching

DanSimon:
This introduces two changes to how loading config for service server connection handlers works:

1.  Configuration is now path-based, where for a server named `foo`, the config is checked in `colossus.service.foo` falling back to defaults in `colossus.service.default`.  While it is still possible to directly pass a config object to load a service config, by default it will now use the default loaded config and check these paths.  So in practice users should almost never have to create `ServiceConfig` objects.
2.  Since each instantiation of a connection handler attempts to load config, basic caching of `ServiceConfig`based on service name has been added.  Again this is only used when the path-based config is used, so users are still free if (for some odd reason) they want to differently configure different connection handlers for the same server.

### DIFF
--- a/colossus-examples/src/main/resources/application.conf
+++ b/colossus-examples/src/main/resources/application.conf
@@ -1,0 +1,3 @@
+colossus.service.benchmark {
+  request-metrics = false
+}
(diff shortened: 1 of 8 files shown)


aliyakamercan:
👍

PR #402, opened 2016-06-13
add a new line at the end of metrics ref conf

aliyakamercan:
with out this `reference.conf` end up being 
```
   ....
    # path.to.my.metric{
    #
    #}
  }
}colossus{

  iosystem {
    #If n....
``` 
causing 
> Exception in thread "main" com.typesafe.config.ConfigException$Parse: reference.conf: 36: Expecting end of input or a comma, got 'colossus' (if you intended 'colossus' to be part of the value for 'colossus', try enclosing the value in double quotes)

@DanSimon


benblack86:
👍

DanSimon:
Odd that I haven't run into this, but it makes sense. 👍

PR #403, opened 2016-06-13
Several changes to metrics

DanSimon:
This PR makes a whole bunch of necessary changes to get metrics properly working:

* `MetricSystem` no longer has a configurable root, it's namespace is always `/`.  Now that we have proper metric namespaces there's no need to have a preset root.
* `MetricSystem` now has a `name`, however this is only used for naming actors so in most cases it can stay as the default
* A new `SystemMetricsConfig` is used to configure the system metrics.  
* `IOSystem` also no longer has a configurable root.  Its name is not used in its metrics namespace
* Servers now create their own namespace using the server's name.  Workers use the IOSystem's name.
* Created a new `MetricSystemConfig` that loads config settings for metric systems.  This simplified the constructors a bit
* `IOSystem` now has a constructor that can take in an `IOSystemConfig` and a `MetricSystem`
* Fixed a bug where histograms and counters were not being created with the full address
* Fixed a bug where system metrics did not have the hostname tag
* `ControllerConfig`now has a new `metricsEnabled` parameter, which `ServiceServer` propagates based on its own `requestMetrics` config value.  This is used for the `input_size` metric.
* Changed `MetricInterval` to `CollectionInterval` to keep naming consistent

### DIFF
--- a/colossus-metrics/src/main/resources/reference.conf
+++ b/colossus-metrics/src/main/resources/reference.conf
@@ -1,9 +1,17 @@
 colossus {
   metrics {
     system {
-      collect-system-metrics: true
+      # name is only used when naming internal actors or logging, generally this
+      # only needs to be changed in the rare event multiple metric systems are
+      # required.  Name is not used at all in metric namespaces or addresses
+      name = "metrics"
+
+      #system metrics include stats like GC time and memory usage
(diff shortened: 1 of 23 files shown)


aliyakamercan:
👍

PR #404, opened 2016-06-16
Minor Metrics config refactoring

DanSimon:
This makes a few changes to how `MetricSystem` is created and configured:

* Name has been brought back as a constructor parameter.  Now by default it will look for config at the path `colossus.metrics.<name>.system`, falling back to defaults in `colossus.metrics.system`.  Name is still not used in metric namespaces, but only for config and actor naming.
* By default, the system metrics will have a namespace using just the metric system's name, although this can be easily changed.
* `IOSystem` will now create a `MetricSystem` with the same name as it, so in most cases `MetricSystem` name is taken care of automatically.
* did a bit of refactoring around `CollectorConfig`.


aliyakamercan:
👍

PR #411, opened 2016-06-20
server settings is now path-based with defaults

DanSimon:
Similar to everything else, configuring server's now first looks for the configuration at the path `colossus.server.<name>`, falling back to just `colossus.server` defined in the reference.conf


aliyakamercan:
:+1

Issue #413, opened 2016-06-22
AsyncServiceClient doesn't respect graceful disconnect


DanSimon:
This is certainly going to be fixed as part of #412 but this was discovered separately so we'll track it in its own issue.

Right now calling `disconnect` on an `AsyncServiceClient` essentially triggers a forced-disconnect instead of a graceful disconnect.  In other words, writing code like : 

```scala
val client: HttpClient[Future] = Http.futureClient("host", 2345)
val response: Future[HttpResponse] = client.get("/foo/bar")
client.disconnect()
```

is supposed to allow the pending request complete before actually closing the connection, but right now it ends up sending a `Disconnect(id)` message to the worker, bypassing the graceful shutdown process and immediately killing the connection.  This does work correctly for in-thread (callback) clients, so we just have to fix the logic around what message is sent to the worker and the underlying connection.

PR #415, opened 2016-06-22
Refactoring AsyncServiceClient into FutureClient

### LINKED ISSUE #412: Cleanup AsyncServiceClient
`AsyncServiceClient` is the basis of how we can create Future clients.  It is a combination of an extended `ServiceClient`, a proxy actor, and an interface that extends `Sender` that allows us to tunnel request objects through the proxy actor to the actual client.

Two problems with the current implementation:

1.  This will cause problems with #406 and #405 , since the base class of the extended client will have to change per protocol
2.  Most of this was written before we added the `ProxyActor`mixin, so there's more weirdness than there needs to be to hook up an actor to the internal client.

The plan to fix this will be to drop the extended class, create a new WorkerItem that simply sets up the service client itself and uses it's own Proxy actor as the actor for the interface.

DanSimon:
Fixes #412  and #413 

Farewell `AsyncServiceClient`, long live `FutureClient`.  

* `shutdownRequest` has been moved from `ConnectionHandler` to `WorkerItem`, so all worker items can now have a shutdown procedure
* made several improvements to `ProxyActor` trait including fully handling the linking between itself and the proxy actor
* made improvements to the `WorkerItemProxy`, the actual proxy actor, so it now properly handles shutdown logic and correctly buffers messages before the worker item is bound to the worker.
* `AsyncServiceClient` and it's custom proxy actor are gone.  The new implementation has the exact same public interface, but aside from being considerably simpler it now uses the standard proxy actor.

As a side note, since 0.7.x `AsyncServiceClient` has been hidden by protocol specific future clients, so while this is technically not fully backwards compatible, any code using something like `Http.futureClient` will be unaffected.

### DIFF
--- a/colossus-testkit/src/main/scala/colossus-testkit/ServiceSpec.scala
+++ b/colossus-testkit/src/main/scala/colossus-testkit/ServiceSpec.scala
@@ -35,9 +35,9 @@ abstract class ServiceSpec[C <: Protocol](implicit provider: ServiceCodecProvide
     requestTimeout = timeout
   )
 
-  def client(timeout: FiniteDuration = requestTimeout) = AsyncServiceClient[C](clientConfig(timeout))//, clientProvider.clientCodec)
+  def client(timeout: FiniteDuration = requestTimeout) = FutureClient[C](clientConfig(timeout))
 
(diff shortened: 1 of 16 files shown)


aliyakamercan:
👍

PR #419, opened 2016-06-29
fixing bug with client graceful disconnect

### LINKED ISSUE #418: graceful disconnect on ServiceClient doesn't work sometimes for buffered requests
This is a race condition that can occur in certain conditions, particularly if a client is opened, one request is sent and then `disconnect()` is immediately called.  The disconnect process is supposed to allow all pending requests to complete before it actually terminates the connection, however in this case the termination happens before some requests have completed.

`ServiceClient` checks `sentBuffer.length` to see if it has any outstanding requests.  However since it requires 1 event loop iteration for the underlying controller layer to actually send the data and call the callback that adds the request to the sent buffer, there is a brief window where the `ServiceClient` is in the process of sending the request but doesn't "know" about it.  If the sent buffer is otherwise empty and `disconnect()` is called during that time, the client will just kill the connection.

This can be fixed by also checking if the controller's pending buffer is empty, and delay shutdown if so.

DanSimon:
Fixes #418, which has all the details about this.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/service/ServiceClientSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/service/ServiceClientSpec.scala
@@ -268,10 +268,10 @@ class ServiceClientSpec extends ColossusSpec {
         case Failure(nope) => throw nope
         case _ => throw new Exception("Bad Response")
       }
+      client.disconnect()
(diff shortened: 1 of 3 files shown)


benblack86:
👍

PR #420, opened 2016-06-29
Fix websocket server

DanSimon:
Well it turns out I accidentally _totally_ broke the websocket server when I refactored it.  So I've fixed it and added some tests to ensure the whole upgrade process works correctly.


DanSimon:
I've added fixes for the issues mentioned in #424 .  Turns out the payload-size byte order was backwards for payloads larger than 125 bytes.  I also added constructor parameter for a controller config object with a default using a much larger output buffer size.  Lastly the `sendMessage` function fully reflects the underlying send operation and I included an example in the websocket example of how to use it to prevent overloading the output buffer.

aliyakamercan:
👍

Issue #407, opened 2016-06-20
Separate Connection Handlers from Request Handlers


DanSimon:
Right now when a user creates an instance of `Service[Protocol]`, they are effectively creating the complete connection handler.  This has historically presented a few problems.  In particular, if a protocol wants to add extra functionality to the connection handler, it needs to provide a sub-class of `Service`, and the user has to know to construct that subclass instead of the base `Service`.

This is currently the case with Http.  It provides a `HttpService` class that adds some extra logic for automatically dealing with the http version and `Connection` header, but it's not immediately clear that this is the class the user should implement to provide that functionality.  

Furthermore, as we implement #406 , the protocol must now make decisions about which implementation of a particular layer to mixin.  The user should not need to know that using the `StreamingHttp` protocol requires the `StreamingController` layer.  

Therefore is makes sense to split a service connection handler into two parts.  The actual connection handler will now be constructed by the protocol itself, and the user will now be responsible for simply providing a request handler, the actual mapping of requests to responses.  The overall interface won't change all that much, and the request handler will still have access to methods like `disconnect`, but now the protocol is free to automatically provide its own implementation of the underlying logic and leave just the higher-level request processing to the user.

In particular, I'm envisioning we can eliminate the `ServiceCodecProvider` and instead have a protocol-based typeclass that more-or-less implements the `Initializer.onConnect` method.  This also means creating a new `RequestHandler[P <: Protocol]` type.

The interface for creating a service would thus change to something like:

```scala
Server.start("myserver", 8080){worker => new Initializer {
  def onConnect = new RequestHandler[Http](_) {
    def handle = //...
  }
}
```

This means that somehow, we will have an implicit `ConnectionHandlerProvider[Http]` with a method like `provideHandler(context: ServerContext, handler: RequestHandler[Http]`, which will be invoked to create the complete connection handler.

This also means that in the rare event the user needs to directly override some of the lower-level connection handler logic, they'll have to create their own `ConnectionHandlerProvider` that create their own custom connection handler.  But this will not be necessary for like 99% of use cases.

PR #421, opened 2016-06-29
Major Protocol, Connection handler Refactoring

DanSimon:
This PR got way larger than I anticipated but I've given up trying to split it up after a couple failed attempts.  This is the first step towards a fairly major shift in the role of Protocols and a fundamental change in the architecture of connection handlers.  

Overall, the goal here is to give protocols more power and responsibility in the API for services, both servers and clients.  We have already been moving in this direction for quite some time, but this PR signals a shift where users will now rarely interact with core components of colossus.

This PR is the first of several, and introduces a lot of new code that is intended to replace existing code.  There's a bunch of commented out stuff and some tests were disabled.  All of this will be fixed in subsequent PR's.  

I'll break down the major changes :

### WorkerItem, CoreHandler, Controller, ServiceServer, are now traits

This is the first step for making connection handlers more modular and allowing various layers to have multiple implementations.  This is not fully complete, as for example `ServiceServer` still extends `Controller`.  However new traits `ControllerIface`and `ControllerImpl`now define the interface for the controller layer, such that eventually the service layer will be able to be mixed in with anything implementing the `ControllerIface` trait.

### Redesigned Protocol trait and new Encoding trait

Originally types like `ServiceClient`required two type parameters for the request and response types.  Then we introduced the `Protocol`phantom trait, such that we could pair `HttpRequest`and `HttpResponse` into just `Http`, so we could just do `ServiceClient[Http]`.  However this didn't work at the controller level, since `Controller` is used in both servers and clients, and the input/output types are reversed (for example `HttpRequest` is the input type for a http server, but the output type for a http client), so we were stuck using two type parameters.

Now there is a new `Encoding` phantom trait which has `Input`and `Output` type members.  `Protocol` has `Request` and `Response` type members, and then creates two encodings, one for servers and one for clients.  While the service layer takes in a `Protocol`as a type parameter, the controller layer now uses the `Encoding` trait, and we can use the protocol's `ServerEncoding` and `ClientEncoding` to setup the controller for a server or client.

### New StaticController and StaticCodec

This is the beginning of the next step of cleaning up codecs and the controller layer.  The new Static Controller removes all of the streaming logic from the controller layer.  Going forward we will have total separation between static and streaming protocols, so eventually we'll have a `StreamingController` layer.

Likewise, there is a new `StaticCodec` that redefines the interface for codecs to remove the streaming aspects.

The existing `Controller` is now deprecated and will be totally removed in a subsequent PR.  

### New RequestHandler and separation from connection Handler

Currently when users create a new `Service[Protocol]` they are creating the entire connection handler as one trait.  The problems with this are explained more in #407 .

Now there is a separation between the connection handler, which is defined and instantiated by the protocol itself, and the request handler, which is what the user provides.

There are a few remaining issues, such as how we deal with actor messages and a few other things, but it's pretty straightforward to fix them.

### Protocols now have their own Server DSL's

This is major paradigm shift that gives each protocol responsibility for creating its own DSL for starting servers.  No longer will users use the core `Server.start`, but instead each protocol defines its own server, initializer, and request handler types to use.  The whole point is to allow protocols to add custom functionality into each component while still allowing users to easily override anything.  For example, the Http protocol uses this to define default headers added to every server response, and to add logic to check the `connection` header.

This change is also needed when we begin to have multiple implementations of various layers.  No longer can we construct just a `ServiceServer` for all protocols.  For example, the `StreamingHttp` protocol needs to create a connection handler that will mixin the `StreamingController` instead of the `StaticController`.

Of course these all inherit from generic versions and the `BasicServiceDSL` provides a complete implementation for protocols that don't need to hook any custom functionality into any component.

From a user's point of view, not much effectively changes aside from what packages to import.  Every protocol now has a `server` sub-package which contains the necessary types.


DanSimon:
ugh, opened against wrong base branch, closing

PR #427, opened 2016-07-12
break out error logging into overridable method

DanSimon:
This simply makes it easier to take more control over how request errors are logged.  Users can override this to, for example, log different types of errors differently, or perhaps not at all.  This does not apply to metrics, so even if log errors are totally suppressed, error metrics will still be ticked.


benblack86:
👍

PR #429, opened 2016-07-13
Rename things

DanSimon:
This was going to be a larger PR, but I think we'll take it in smaller steps.  Since we've now removed all the streaming code, there's not need to have things prefixed with "Static", so just rename `StaticCodec` and `StaticController`


benblack86:
👍

PR #430, opened 2016-07-13
New Http streaming API part 1, basic server API

DanSimon:
This is the beginning of a new API for creating streaming Http servers.  Rather than try to build this on the service layer, this builds directly on top of the controller layer.  This means that there is no built-in support for concurrency (you have to ensure you send back data in the right order for multiple requests), but much greater flexibility and a much simpler interface for streaming large or chunked messages.  

I think overall this is a fair tradeoff since usually when writing streaming services most connections only send one request and then receive a very long, possibly infinite, response.  It's also possible we can create higher-level abstractions on top of what's here.

This PR is just for the server-side.  I don't think it will be too much work to get the client-side done but this is part of my continued effort to keep PR's small.  I also want to get a basic API out there as fast as possible as I'm fairly certain it's going to take some time to refine and make it more intuitive.


DanSimon:
I made a few additional changes to make the codec work for both servers and clients.  I've also added a barebones client handler, although I'll be significantly improving it in coming pr's.

colossus/src/main/scala/colossus/controller/Controller.scala
```diff
@@ -32,13 +32,20 @@ trait ControllerIface[E <: Encoding] {
   protected def controllerConfig: ControllerConfig
   implicit val namespace: MetricNamespace
 
-  protected def onFatalError(reason: Throwable): Option[E#Output] = None
+  protected def onFatalError(reason: Throwable): Option[E#Output] = {
+    //TODO: Logging
```
DanSimon: I will be fixing this in my next PR

colossus/src/main/scala/colossus/protocols/http/StreamingHttp.scala
```diff
@@ -0,0 +1,231 @@
...
+      case Some(End()) => {
+        state = HeadState
+        Some(End())
+      }
+      case other => other
+    }
+  }
+
+
+  def resetDecoder() {
```
benblack86: For clarity I like to see all methods provide a return type. I'm guessing in this case it is Unit?

benblack86:
Looks good to my untrained eyes.

PR #433, opened 2016-07-21
array building cleanup and optimizations

DanSimon:
**tl;dr** - fixes a small performance regression involving parsing lines, cleans up code

These are some small optimizations that seem to improve overall parsing performance rather significantly.  Overall in benchmarks it improves throughput by about 10%, or rather I should say it fixes a performance regression that decreased throughput by about 10%.  I traced it down to the fact that in `FastArrayBuilder` the code for growing the array was moved into its own function.  Inlining the function fixes the issue.  Not entirely sure why that works, I spent some time looking at generated bytecode and don't see an obvious reason, but if it works, it works.

I also made some other changes that cleaned up the code a bit.  Furthermore, the line parser now correctly verifies that a `\r` is followed by a `\n`where before it just assumed the byte following a `\r`was correct.


benblack86:
👍

aliyakamercan:
👍

PR #434, opened 2016-07-22
moving idleCheck to WorkerItem mixin trait

### LINKED ISSUE #432: Request timeouts sometimes fail to occur on clients
If a service client is not able to connect to its remote host and is in the process of reconnecting, and if `failFast`is turned off, then any queued requests will not have their request timeouts respected.

This occurs because the periodic idle check is only called on active connections, and while a service client is active when it begins the connection process, if the connection fails and the client backs off to wait to try again, during that backoff time there is no active `Connection`object and the idle check is not performed.

There are two possible solutions:

1.  Don't destroy and re-create a `Connection` each time a client attempts to connect.  Leaving the connection object in the `connections` map will ensure idle check continues to happen.  However, currently we strongly associate a `Connection`with a `SelectionKey` and `SocketChannel`, which do need to be recreated each time.

2.  Move `idle check` from `ConnectionHandler`to `WorkerItem`, so the method is called on all bound worker items and not just active connection handlers.  Since every connection handler is a worker item, this would also solve the issue.  This would then mean each idle check would loop over both `WorkerItems` and `Connections`, since `Connection`must also perform the idle connection timeout check.  This might make idle checks slightly slower, but in general that hasn't caused much issue.

So far it seems like option 2 is the best.  It might actually be better to make this a mixin trait so worker items can then choose if they want to be part of the idle check.

DanSimon:
Fixes #432 

The `idleCheck` method has been moved out of `ConnectionHandler` and into a separate mixin trait for `WorkerItem`.  The main reason we're doing this is so that the idle check on `ServiceClient` can occur even when the client is waiting to reconnect and is not recognized as a connection handler.  

I also changed the parameter passed into `idleCheck` from `Duration` to `FiniteDuration`, since there's no reason for it to ever be infinite.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/controller/OutputControllerSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/controller/OutputControllerSpec.scala
@@ -110,7 +110,7 @@ class OutputControllerSpec extends ColossusSpec {
       val endpoint = static()
       val p = endpoint.typedHandler.pPush(ByteString("wat"))
       Thread.sleep(300)
-      endpoint.handler.idleCheck(1.millisecond)
+      endpoint.typedHandler.idleCheck(1.millisecond)
       p.expectCancelled()
     }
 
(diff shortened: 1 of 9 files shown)


aliyakamercan:
👍

PR #436, opened 2016-07-26
Metrics getters

DanSimon:
This adds some methods to `Rate` and `Histogram` to make it easier to retrieve values as they're being collected.  In particular  they should help make it easier in tests to verify that metrics are properly being collected with the right values.

Also did a little refactoring in `Histogram` to get rid of the obsolete `Snapshot`.


DanSimon:
Also adding a fix for #437

dxuhuang:
👍

PR #439, opened 2016-07-26
Make clients compatible with Scalamock

DanSimon:
So right now mocking service clients doesn't really work.  Mockito just doesn't like the higher-kinded type parameter and scalamock doesn't seem to like that we have a trait extending a class without a default constructor.

This solves the issue with scalamock.  `LiftedClient`, which was previously a class, is now a trait, and the new `BasicLiftedClient` is a class extending it.  this way, for example, `HttpClient` can still extends `LiftedClient` but also work with scalamock.


benblack86:
👍

PR #440, opened 2016-08-03
Server slow start

DanSimon:
This is a new feature that allows servers to gradually increase the number of allowable connections when first starting.  This is useful to help mitigate the effects of JVM warmup and help stop thundering herd problems when restarting a running service under heavy load.


benblack86:
👍

PR #441, opened 2016-08-04
Server cleanup

DanSimon:
* Adding a proper `onShutdown()` hook for `Initializer`
* Made a bunch of internal classes and objects private
* `ServerRef` now has a `info()` method instead of users sending the underlying actor a message
* fixed up a bunch of scaladocs for Server-related stuff,


benblack86:
👍

PR #450, opened 2016-09-22
Resurrect Pipes

DanSimon:
So when I deleted all the streaming code in #426, I was probably a little overzealous and deleted all the `Pipe` code.  However with the new streaming API pipes are going to make a comeback and will probably work a whole lot better than they did before.

All in all, pipes will serve a slightly different purpose than before.  One of the goals behind the previous iteration of pipes was to facilitate true zero-copy data proxying.  The idea was you could take a raw `DataBuffer` received on a server-side connection and pipe it directly through to a client-side connection without ever copying it.  However, not only has doing that been impossible since 0.7.0 (with the implementation of pull-based writing), but we never actually implemented that, and I believe it was adding way too much complexity for a fairly specific use-case.

Removing that goal considerably cleans up the logic around pipes and I think will make them much more useful overall.

[This comparison](https://github.com/tumblr/colossus/compare/12916d4...132407c) shows the changes I made to the resurrected code.  The main differences are:

* `FiniteBytePipe` is dead and gone
* `InfinitePipe` is now `BufferedPipe` and now has an internal fixed-size buffer.  When the size is set to 0 it behaves as `InfinitePipe` did.
* No action on pipes will cause an exception to be thrown, which I think was previously a major source of instability.  All exceptions are now propagated inside callback functions.
* `BufferedPipe` also now has high and low watermarks for pushing items.  When the internal buffer hits the high watermark, calls to `push` will start returning `Filling` instead of `Ok`.  Likewise, the size of the low watermark determines when the trigger callback is executed, which signals to the producer that its ok to push again.  They should be useful in the future to help smooth out backpressure.

This is probably the first of several PR's that will improve and integrate pipes into the new streaming API.  Both the Streaming Http and upcoming Http/2 API's will be using this extensively.

### DIFF
--- a/colossus/src/main/scala/colossus/controller/Pipe.scala
+++ b/colossus/src/main/scala/colossus/controller/Pipe.scala
@@ -3,9 +3,17 @@ package controller
 
 import core.DataBuffer
 import akka.util.ByteString
+import java.util.LinkedList
 import scala.util.{Try, Success, Failure}
 
 import service.{Callback, UnmappedCallback}
+
+
+trait StreamState[T] {
+
+  def terminus: T
+
+}
 
 trait Transport {
   //can be triggered by either a source or sink, this will immediately cause
@@ -29,11 +37,12 @@ trait Sink[T] extends Transport {
   def isFull: Boolean
 
   //after this is called, data can no longer be written, but can still be read until EOS
-  def complete()
+  def complete(): Try[Unit]
 
   def feed(from: Source[T], linkState: Boolean) {
     def tryPush(item: T): Unit = push(item) match {
       case PushResult.Ok            => feed(from, linkState)
+      case PushResult.Filling            => feed(from, linkState)
       case PushResult.Closed        => from.terminate(new Exception("This is probably not what we want to do"))
       case PushResult.Complete      => from.terminate(new Exception("This is probably not what we want to do"))
       case PushResult.Full(trig)    => trig.fill(() => tryPush(item))
@@ -141,9 +150,9 @@ object Source {
  */
 trait Pipe[T, U] extends Sink[T] with Source[U] {
 
-  def join[V, W](pipe : Pipe[V, W])(f : U => Seq[V]) : Pipe[T, W] = PipeCombinator.join(this, pipe)(f)
+  //def join[V, W](pipe : Pipe[V, W])(f : U => Seq[V]) : Pipe[T, W] = PipeCombinator.join(this, pipe)(f)
 
-  def ->>[V, W](pipe : Pipe[V, W])(f : U => Seq[V]) : Pipe[T, W] = join(pipe)(f)
+  //def ->>[V, W](pipe : Pipe[V, W])(f : U => Seq[V]) : Pipe[T, W] = join(pipe)(f)
 }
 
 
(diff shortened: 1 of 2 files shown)


colossus/src/main/scala/colossus/controller/Pipe.scala
```diff
@@ -0,0 +1,411 @@
...
+class PipeStateException(message: String) extends Exception(message) with PipeException
+
+/**
+ * This is a special exception that Input/Output controllers look for when
+ * error handling pipes.  In most cases they will log the error that terminated
+ * the pipe, but for this one exception, the failure will be silent.  This is
+ * basically for situations where a certain amount of data is expected but for
+ * some reason the receiver decides to cancel for some business-logic reason.
+ */
+class PipeCancelledException extends Exception("Pipe Cancelled") with PipeException
```
jlbelmonte: Are this exceptions intended to be regular classes or should they be case classes. Maybe it could be helpful  for exhaustiveness.

colossus/src/main/scala/colossus/controller/Pipe.scala
```diff
@@ -0,0 +1,411 @@
...
+
+/**
+ * A Source is the read side of a pipe.  You provide a handler for when an item
+ * is ready and the Source will call it.  Note that if the underlying pipe has
+ * multiple items ready, onReady will only be called once.  This is so that the
+ * consumer of the sink can implicitly apply backpressure by only pulling when
+ * it is able to
+ */
+trait Source[T] extends Transport {
+  def pull(onReady: Try[Option[T]] => Unit)
```
jlbelmonte: nitpick. 
Does pull evaluate to Unit?  return type will make the trait more understandable at first sight.

colossus/src/main/scala/colossus/controller/Pipe.scala
```diff
@@ -0,0 +1,411 @@
...
+  sealed trait State
+  sealed trait PushableState extends State
+  case class Full(trigger: Trigger) extends State
+  case class Pulling(callback: Try[Option[T]] => Unit) extends PushableState
+  case object Closed extends State
+  case object Idle extends PushableState
+  case class Dead(reason: Throwable) extends State
+
+  //Full is the default state because we can only push once we've received a callback from pull
+  private var state: State = Idle
```
jlbelmonte: Seems that comment and default value of state differ.

DanSimon: Ah yeah the initial state used to be `Full` because without a buffer the producer could only push an item if the consumer had already asked for one, but that's no longer the case so `Idle` is the correct initial state.

colossus/src/main/scala/colossus/controller/Pipe.scala
DanSimon: I'm going to leave this as-is for now since I already have another branch that changes the logic of `Pull` so it returns an ADT similar to `push`

colossus/src/main/scala/colossus/controller/Pipe.scala
DanSimon: Yeah it's kind of arbitrary now.  I made it a regular class since you don't want an object as an exception and a case class with no parameters seems weird.  I do think this does bring forth an issue that exception for pipes probably need to be a bit more structured and unified.

jlbelmonte: Oh yes if there is no parameter you don't want a case object as an exception. 👍

PR #451, opened 2016-09-28
Streaming Http Service, part 1

DanSimon:
**Be Aware - This is the first of Several PR's**.  In an effort to keep my PR's small, I'll be breaking all of this work into several smaller PR's, so in this PR things are a little disorganized and lacking tests.  Notice that the base branch is not develop.  All PR's will go into `develop-streaming-http` and then we'll do one last big PR into develop.

There are a few significant changes in this PR:

### New Streaming Http Service API

Oh boy, it's back!!  The stream service API provides a way to write services that stream in/out requests/responses.  So instead of having the body of a message be a `HttpBody` (which is just a wrapper around a byte array), it is a `Source[BodyData]` which allows you to read in chunks of a message as they arrive (in a non-blocking manner).  

While this functionality has been around since Colossus 0.6.0, this new version works quite differently.  The `StreamService` API is built on top of the streaming http API, so we now have two layers of message processing:

On the input side:

1.  Raw Data is received on the connection and passed to the streaming http API.
2.  The data is decoded into a series of `HttpStream` objects, which can either be a `Head`, `Data` or `End`.  So for example, a http request with no body would just be a `Head` and `End`, but a request with a body would have one or more `Data` objects as well.
3.  These messages are passed to the `StreamService` layer, which will create and manage `StreamHttpMessage` objects.  Each message contains a `Source[Data]`.

On the output side:

1.  The `StreamService` layer receives output `StreamingHttpMessage` objects which contain a header and a `Source[Data]`.  These messages are buffered and written out one at a time.
2.  The head of the queue is "collapsed" into a `Source[HttpStream]`.  The message objects read from the source are pushed to the streaming http layer.
3.  When the collapsed source is complete, the message is dequeued and the process begins again with the next one.

So all in all, we now have 3 different Http API's:

1.  The plain old http service API, which is unchanged
2.  The streaming http API, which has an actor-like interface and sends/receives pieces of http messages
3.  The stream service API, which sits on top of the streaming API and presents a service-like interface for incoming/outgoing streams.

### Other Supporting Changes

1.  Renamed `StreamHttpMessage` to `HttpStream` to eliminate confusion with the new `StreamingHttpMessage` type.  Also renamed `BodyData` to just `Data`
2.  `HttpMessageHead` is no longer a recursive type.  This was initially done so we could have generic versions of the `withHeader` method that returned the correct type.  Now this functionality has been moved to a typeclass, which overall makes things a bit easier to follow and more straightforward
3.  Replacing the `Protocol#ServerEncoding` and `Protocol#ClientEncoding`type members with type constructors `Encoding.Server` and `Encoding.Client`.  I ran into a whole bunch of bizarre compile errors with the members, so switching to type constructors basically offers the same functionality

### DIFF
--- a/colossus-examples/src/main/scala/colossus-examples/StreamExample.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/StreamExample.scala
@@ -5,6 +5,7 @@ import colossus._
 import protocols.http._
 import stream._
 import core.{DataBlock, ServerContext}
+import controller._
 
 import scala.util.{Failure, Success, Try}
 
--- a/colossus-examples/src/main/scala/colossus-examples/StreamExample.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/StreamExample.scala
@@ -18,7 +18,7 @@ object StreamExample {
           case Success(num) => {
             upstream.push (Head(HttpResponseHead(head.version, HttpCodes.OK, HttpHeaders.fromString("transfer-encoding" -> "chunked")))){_ => ()}
             def sendNumbers(num: Int): Unit = num match {
-              case 0 => upstream.push(End()){_ => ()}
+              case 0 => upstream.push(End){_ => ()}
               case n => {
                 upstream.push (BodyData(DataBlock(s"$n\r\n"))){_ => sendNumbers(n - 1)}
               }
--- a/colossus-examples/src/main/scala/colossus-examples/StreamExample.scala
+++ b/colossus-examples/src/main/scala/colossus-examples/StreamExample.scala
@@ -12,7 +12,7 @@ import scala.util.{Failure, Success, Try}
 object StreamExample {
 
   class Handler(ctx: ServerContext) extends StreamServerHandler(ctx) {
-    def handle(message: StreamHttpMessage[HttpRequestHead]) {
+    def handle(message: HttpStream[HttpRequestHead]) {
       message match {
         case Head(head) if (head.path == "/zop") => head.parameters.getFirstAs[Int]("num") match {
           case Success(num) => {
@@ -20,7 +20,7 @@ object StreamExample {
             def sendNumbers(num: Int): Unit = num match {
               case 0 => upstream.push(End){_ => ()}
               case n => {
-                upstream.push (BodyData(DataBlock(s"$n\r\n"))){_ => sendNumbers(n - 1)}
+                upstream.push (Data(DataBlock(s"$n\r\n"))){_ => sendNumbers(n - 1)}
               }
             }
             sendNumbers(num)
--- a/colossus/src/main/scala/colossus/protocols/http/StreamService.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/StreamService.scala
@@ -13,13 +13,13 @@ import scala.util.{Try, Success, Failure}
 trait StreamingHttpMessage[T <: HttpMessageHead] {
 
   def head: T
-  def body: Source[BodyData]
+  def body: Source[Data]
 
-  def collapse: Source[StreamHttpMessage[T]] = Source.one[StreamHttpMessage[T]](Head(head)) ++ body
+  def collapse: Source[HttpStream[T]] = Source.one[HttpStream[T]](Head(head)) ++ body
 }
 
-case class StreamingHttpRequest(head: HttpRequestHead, body: Source[BodyData]) extends StreamingHttpMessage[HttpRequestHead]
-case class StreamingHttpResponse(head: HttpResponseHead, body: Source[BodyData]) extends StreamingHttpMessage[HttpResponseHead]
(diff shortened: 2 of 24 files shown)


colossus-examples/src/main/scala/colossus-examples/StreamExample.scala
```diff
@@ -5,21 +5,22 @@ import colossus._
...
       message match {
         case Head(head) if (head.path == "/zop") => head.parameters.getFirstAs[Int]("num") match {
           case Success(num) => {
             upstream.push (Head(HttpResponseHead(head.version, HttpCodes.OK, HttpHeaders.fromString("transfer-encoding" -> "chunked")))){_ => ()}
             def sendNumbers(num: Int): Unit = num match {
-              case 0 => upstream.push(End()){_ => ()}
+              case 0 => upstream.push(End){_ => ()}
               case n => {
-                upstream.push (BodyData(DataBlock(s"$n\r\n"))){_ => sendNumbers(n - 1)}
+                upstream.push (Data(DataBlock(s"$n\r\n"))){_ => sendNumbers(n - 1)}
```
benblack86: How is back pressure handled? Does `push` block if we send too many Data packets?

colossus/src/main/scala/colossus/protocols/http/StreamService.scala
```diff
@@ -0,0 +1,222 @@
+package colossus
```
benblack86: Any particular reason for this style instead of `package colossus.protocols.http.stream`?

colossus/src/main/scala/colossus/protocols/http/StreamService.scala
```diff
@@ -0,0 +1,222 @@
...
+
+  type InputHead = E#Input
+  type OutputHead = E#Output
+
+  private var currentInputStream: Option[Sink[Data]] = None
+
+  private val outputStreams = new MessageQueue[Source[HttpStream[OutputHead]]](100)
+
+  protected def fatal(message: String) {
+    println(s"FATAL ERROR: $message")
```
benblack86: fatal instead of println?

colossus/src/main/scala/colossus/protocols/http/StreamService.scala
```diff
@@ -0,0 +1,222 @@
...
+      }
+      case b @ Data(_, _) => currentInputStream match {
+        case Some(sink) => sink.push(b) match {
+          case PushResult.Filled(signal) => {
+            upstream.pauseReads()
+            signal.react{
+              upstream.resumeReads()
+            }
+          }
+          case bad: PushResult.NotPushed => {
```
benblack86: Don't need to capture `bad: ` since it isn't being used

colossus/src/main/scala/colossus/protocols/http/StreamService.scala
```diff
@@ -0,0 +1,222 @@
...
+            // :(
+            fatal("failed to push message to stream")
+          }
+          case other => {}
```
benblack86: fatal(s"unknown push result $other")?

colossus/src/main/scala/colossus/protocols/http/StreamService.scala
```diff
@@ -0,0 +1,222 @@
...
+case class StreamingHttpResponse(head: HttpResponseHead, body: Source[Data]) extends StreamingHttpMessage[HttpResponseHead]
+
+
+trait StreamingHttp extends Protocol {
+  type Request = StreamingHttpMessage[HttpRequestHead]
+  type Response = StreamingHttpMessage[HttpResponseHead]
+}
+
+
+object GenEncoding {
```
benblack86: Not sure what "Gen" means.

colossus-examples/src/main/scala/colossus-examples/StreamExample.scala
DanSimon: `push` will simply return false if the data is not successfully buffered.  However in this case that won't happen since only one data block is pushed at a time.  The callback function that does the recursion is only called when the data has been sent (not just buffered).

That said, dealing with backpressure is a little tricky here and I'm experimenting with using a pipe here instead of having the controller layer doing its own backpressure management.

colossus/src/main/scala/colossus/protocols/http/StreamService.scala
DanSimon: It lets you import the whole `colossus` package at the same time.

colossus/src/main/scala/colossus/protocols/http/StreamService.scala
DanSimon: Supposed to mean "generic".  I will probably find a better name and also try to figure out if this needs to be public

colossus/src/main/scala/colossus/protocols/http/StreamService.scala
DanSimon: actually in this case there is no error and no further action needs to be taken.  I'll add some comments to make that more clear

benblack86:
👍

PR #452, opened 2016-10-04
Pipe multiplexing

DanSimon:
This PR has some substantial changes and improvements to pipes, as well as a first run of multiplexing support for pipes, a crucial prerequisite for http/2 support.

_Notice that, as with my other PR's, the base branch is `develop-streaming-http`, so as a whole this is all still a work in progress._

### What is Multiplexing

In short, it's merging several independent sub-streams into a single multiplexed stream.  Demultiplexing is the process of taking a multiplexed stream and splitting it into its constituent substreams.  

In other words, a multiplexing pipe has the type `Pipe[Source[T], T]` and a demultiplexing pipe has the type `Pipe[T, Source[T]]`.

In order for demultiplexing to work, every message in the multiplexed stream must have a stream id that the demultiplexer can use to forward the message to the correct sub-stream.  Furthermore, it needs to know which messages start and end a sub-stream.  All of these are handled by the `MultiStream[K,T]` typeclass, which must have an instance in scope for messages of type `T` with stream ids of type `K`.

### Changes to Pipes

oh boy here we go:

As a quick recap, the whole point of a `Pipe[I,O]` is to provide a buffer that can mediate back-pressure.  Unlike a regular queue, pipes provide the ability to gracefully block overflow attempts and can provide callback signals to both producers waiting for the buffer to empty and consumers waiting for the buffer to fill.

#### Pipe.pull now returns an ADT

Previously this method required a function `Try[Option[T]] => Unit` and would either immediately call this function if it had a buffered item, or would hold onto it until it did.  This had two major problems:

1.  Repeatedly pulling from a pipe was very inefficient.  Basically you'd have to call `pull` again inside the callback function, resulting in some nasty recursion.
2.  It was impossible to "attempt" to pull.  In some cases you want to try pulling an item out of a pipe, but if nothing is there you _don't_ want to give the pipe a callback that is eventually called.  Sometimes you just want nothing and move on.

It also made testing a lot weirder.  So now `pull()` takes no parameters and returns a `PullResult` ADT.  

#### Pipe Fast-Tracking

So as I mentioned you can pull multiple items from a pipe more efficiently simply by looping until something besides `Item` is returned.

However there is an even faster way to do this.  The consumer can give the pipe a function of type `NEPullResult[T] => Boolean`  (`NEPullResult` is a subset of `PullResult` without the `Empty` case) and the pipe will simply hold onto this function and completely bypass buffering or backpressure logic until the function returns `false` (or until the pipe is closed or terminated).

This is now used in the `Pipe.pullWhile` method.  

#### Signals can either notify or react

**This was removed in a subsequent commit**

~~A `Signal[T]` is the trait returned to either a producer attempting to push to a full pipe, or to a consumer trying to pull from an empty pipe.  Previously you could "listen" on a signal with function of type `() => Unit` that would be called when the pipe is ready for action.~~

~~Now signals also support listening with a callback of type `T => Unit`.  This is not used in signals on the producer side, but on the consumer side this is used to signal a consumer immediately with a `NEPullResult[T]`.  This is primarily needed to be able to support pipes with 0-sized buffers.~~

#### Signals now support multiple listeners

This was needed for multiplexing, particularly so a pipe can have multiple producers, but this will also allow them to support multiple consumers.  This effectively makes pipes more like simple pub-sub channels.

Previously a signal would simply hold onto a single callback, but now it contains a queue.  So, for example, if multiple producers are trying to push to a single pipe but the pipe is full, each producer can add a listener to the signal.  When the pipe finally drains and can accept more items, each listener is signaled one-by-one until the pipe is full again.

This also then means that whereas before a new `Signal` was created each time the pipe filled, a single persistent signal is used.  This way, if the pipe fills up again while the signals are being triggered such that some listeners are still waiting, we can hold onto them and be ready to signal them once the pipe drains again.

#### New combinator methods

Lots of fun stuff here.  In particular, I'm now using a `Functor` typeclass to correctly implement `map` for pipes and sources.  `Pipe` inherits from `Source`, but we want `source.map` to return a source and `pipe.map` to return a pipe.  There are other ways to accomplish that (like how scala collections work), but the typeclass approach seems the simplest.

It is a bit weird that this one function is a typeclass and the others are methods, so this might need some more work.

#### Removed low/high watermarks

This was an interesting idea, but ultimately I realized it's not actually going to work in practice very well.  The idea behind it was to give a producer a heads-up that the pipe was filling up, but not entirely full.  This way the producer could initiate some kind async back-off but not get immediately rejected in case it had a bunch of items it was trying to push at once.

This won't work for a few reasons, but mainly because the producer can never know for sure that they won't try to overflow the pipe.  So even with the watermarks, we still need all the logic to handle an overflow attempt...which in the end isn't actually a whole lot of logic

#### Code organization

Right after this PR, I'll will be opening another that will move all the pipe code into the new `streaming` package and correctly breaking things up into separate files.  

#### Test Coverage

I wrote a bunch of tests but I'm assuming coverage is pretty low right now.  Will continue to work on that as things solidify more.  

#### Feature re-evaluation

**This was addressed in a subsquent commit.  Pipes must have >0 buffer size now**

~~Currently pipes are allowed to have a buffer size of 0, meaning a producer can only push if a consumer is waiting to pull or fast-tracking.  In some cases there is some weird logic that could be solved by having a minimum buffer size of 1.  I initially had some performance concerns around this, but the new pull fast-tracking probably eliminates this.~~

#### Integration with Http2

Stream multiplexing is a central feature of Http2, and while everything here definitely gets us close to achieving that, I'm fairly certain more changes will be needed to actually integrate all this logic into a functional Http2 API.  In particular, Http2 has a bunch of flow-control and prioritization features that we will need to somehow plug-in to multiplexing.

### DIFF
--- a/colossus/src/main/scala/colossus/controller/Pipe.scala
+++ b/colossus/src/main/scala/colossus/controller/Pipe.scala
@@ -56,6 +56,7 @@ trait Sink[T] extends Transport {
     }
   }
 
+
     
 
 }
@@ -66,7 +67,7 @@ object PullResult {
   case class Item[T](item: T) extends NEPullResult[T]
   case class Empty[T](whenReady: Signal[NEPullResult[T]]) extends PullResult[T]
   case object Closed extends NEPullResult[Nothing]
-  case class Terminated(reason: Throwable) extends NEPullResult[Nothing]
+  case class Error(reason: Throwable) extends NEPullResult[Nothing]
 }
 
 /**
@@ -84,11 +85,11 @@ trait Source[+T] extends Transport {
 
   def pull(whenReady: Try[Option[T]] => Unit): Unit = pull() match {
     case PullResult.Item(item)      => whenReady(Success(Some(item)))
-    case PullResult.Terminated(err) => whenReady(Failure(err))
+    case PullResult.Error(err) => whenReady(Failure(err))
     case PullResult.Closed          => whenReady(Success(None))
     case PullResult.Empty(trig)     => trig.react{  
       case PullResult.Item(item)      => whenReady(Success(Some(item)))
-      case PullResult.Terminated(err) => whenReady(Failure(err))
+      case PullResult.Error(err) => whenReady(Failure(err))
       case PullResult.Closed          => whenReady(Success(None))
     }
   }
@@ -132,7 +133,75 @@ trait Source[+T] extends Transport {
 
   def collected: Callback[Iterator[T]] = fold(new collection.mutable.ArrayBuffer[T]){ case (next, buf) => buf append next ; buf } map {_.toIterator}
 
+  /**
+   * Link this source to a sink.  Items will be pulled from the source and
+   * pushed to the sink, respecting backpressure, until either the source is
+   * closed or an error occurs.  The sink will be closed when this source is
+   * closed.  If the sink is closed before this source, this source will be
+   * terminated.  Other terminations are propagated in both directions.
+   */
+  def into[U >: T] (sink: Sink[U]) {
+    def tryPush(item: T): Boolean = sink.push(item) match {
+      case PushResult.Filled(sig) => {
+        sig.react{_ => into(sink)}
+        false
+      }
+      case PushResult.Full(sig) => {
+        sig.react{_ => if (tryPush(item)) into(sink)}
+        false
+      }
+      case PushResult.Closed => {
+        terminate(new PipeStateException("downstream sink unexpectedly closed"))
+        false
+      }
+      case PushResult.Error(err) => {
+        terminate(err)
+        false
+      }
+      case ok => true
+    }
+    def handlePull(r: PullResult[T]): Boolean =  r match {
+      case PullResult.Item(item) => {
+        tryPush(item)
+      }
+      case PullResult.Closed => {
+        sink.complete()
+        false
+      }
+      case PullResult.Error(err) => {
+        sink.terminate(err)
+        false
+      }
+      case PullResult.Empty(sig) => {
+        sig.react{i =>
+          if (handlePull(i)) {
+            into(sink)
+          }
+        }
+        false
+      }
+    }
+    while (handlePull(pull())) {}
+  }
+
+  def map[B](f: T => B): Source[B] = {
+    val me = this
+    new Source[B] {
+      def pull(): PullResult[B] = me.pull match {
+        case PullResult.Item(i) => PullResult.Item(f(i))
+        case other => other.asInstanceOf[PullResult[B]]
+      }
+
+      def outputState = me.outputState
+      def terminate(err: Throwable) {
+        me.terminate(err)
+      }
+    }
+  }
+
+
 }
+
 
 object Source {
   def one[T](data: T) = new Source[T] {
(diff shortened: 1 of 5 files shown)


colossus/src/main/scala/colossus/controller/Pipe.scala
```diff
@@ -99,60 +189,132 @@ trait Source[+T] extends Transport {
...
+  /**
+   * Link this source to a sink.  Items will be pulled from the source and
+   * pushed to the sink, respecting backpressure, until either the source is
+   * closed or an error occurs.  The sink will be closed when this source is
+   * closed.  If the sink is closed before this source, this source will be
+   * terminated.  Other terminations are propagated in both directions.
+   */
+  def into[U >: T] (sink: Sink[U]) {
+    def tryPush(item: T): Boolean = sink.push(item) match {
+      case PushResult.Filled(sig) => {
```
benblack86: Nitpick: I'm guessing `sig` means `signal`? Being more explicit with variable names makes it easier to read.

colossus/src/main/scala/colossus/streaming/PipeMultiplexer.scala
```diff
@@ -0,0 +1,130 @@
...
+  def streamId(t: T): K
+}
+
+case class SubSource[K,T](id: K, stream: Source[T])
+case class SubSink[K,T](id: K, stream: Sink[T])
+
+object Multiplexing {
+
+  def multiplex[K,T](base: Sink[T])(implicit ms: MultiStream[K,T]): Sink[SubSource[K,T]] = {
+    val sources = new BufferedPipe[SubSource[K,T]](10)
```
benblack86: Any particular reason to use 10?

colossus/src/main/scala/colossus/streaming/PipeMultiplexer.scala
```diff
@@ -0,0 +1,130 @@
...
+            case PushResult.Full(signal) => {
+              signal.notify{
+                if (tryPush(close)) {
+                  doPull()
+                }
+              }
+              false
+            }
+            case PushResult.Closed => fatal(s"sub-pipe $id unexpectedly closed")
+            case PushResult.Error(error) => fatal(s"sub-pipe error: $error")
```
benblack86: Include $id in error msg?

colossus/src/main/scala/colossus/streaming/PipeMultiplexer.scala
DanSimon: Ah I made this configurable on the demultiplexing side but forgot to do it for multiplexing.  I think that means I also forgot to test backpressure on the sources pipe so I'll fix that as well.

colossus/src/main/scala/colossus/streaming/PipeMultiplexer.scala
DanSimon: Yeah good idea

colossus/src/main/scala/colossus/controller/Pipe.scala
DanSimon: yeah I'll change this everywhere

colossus/src/main/scala/colossus/streaming/PipeMultiplexer.scala
DanSimon: Actually I take that back, this doesn't need to be configurable and can have a size of 1.  Items pushed to that pipe are always immediately pulled and backpressure is never applied.

PR #455, opened 2016-10-07
Streaming cleanup, part 1

DanSimon:
Few smallish changes in this PR to cleanup some of the new pipe/streaming code:

#### Removing PushResult.Filled

The idea behind `Filled` was that it would be returned when the last item pushed to a pipe filled it, indicating that any subsequent pushes would fail until the pipe freed up space.  This could in theory allow a producer to avoid getting stuck with a "hanging" item, e.g. an item it attempted to push but failed because the pipe was full.

While technically this worked, it doesn't actually make things easier.  In reality, a producer can really never know for sure that they will be able to push an item, so every producer is still forced to deal with the `PushResult.Full` case even if they respect the semantics of `Filled`.

#### New ability to "peek" at a source

This allows a consumer to basically find out if a `Source` has items ready to pull.  The return type is `PullResult[Unit]`, so the structure is the same as `Source.pull` but you can't actually get any items by calling `peek`.

#### New Source.flatten

This takes a `Source[Source[T]]` and "flattens" it to a `Source[T]`.  Basically sources are pulled out one at a time and streamed into the returned source.  This is going to be used to help clean up some of the streaming service logic in an upcoming PR.

#### New options for Source.into

`Source.into` allows you to stream a `Source[T]` into a `Sink[T]`.  Previously when the source closed, the sink would also be closed.  Now this is configurable via the `linkClosed: Boolean` parameter.  This means you can stream multiple sources into a single sink, either sequentially or at the same time.

Also, the new `onComplete: NonOpenTransportState => Any` callback parameter allows you to be notified when the source either completes or terminates.

#### Other minor cleanup

* Removed `Sink.feed(iterator)` as it's the same as `Source.fromIterator(iterator).into(sink)`
* `Source.one` just uses a pipe, doesn't create it's own class.
* Various other small cleanup stuff

### DIFF
--- a/colossus/src/main/scala/colossus/streaming/Pipe.scala
+++ b/colossus/src/main/scala/colossus/streaming/Pipe.scala
@@ -159,13 +159,20 @@ class BufferedPipe[T](size: Int) extends Pipe[T, T] {
   }
 
   override def pullWhile(fn: NEPullResult[T] => Boolean) {
-    state = PullFastTrack(fn)
-    var continue = true
-    while (continue && buffer.size > 0) {
-      continue = fn(PullResult.Item(buffer.remove()))
-    }
-    if (!continue) {
-      state = Active
+    state match {
+      case Dead(reason) => fn(PullResult.Error(reason))
+      case Closed if (buffer.size == 0) => fn(PullResult.Closed)
+      case _ => {
+        val oldstate = state
+        state = PullFastTrack(fn)
+        var continue = true
+        while (continue && buffer.size > 0) {
+          continue = fn(PullResult.Item(buffer.remove()))
+        }
+        if (!continue) {
+          state = oldstate
+        }
+      }
     }
   }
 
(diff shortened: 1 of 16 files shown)


colossus/src/main/scala/colossus/streaming/Pipe.scala
```diff
@@ -131,6 +147,21 @@ class BufferedPipe[T](size: Int) extends Pipe[T, T] {
     }
   }
 
+  /**
+   * Iterate through all buffered items, removing any where the provided
+   * function returns true
+   */
+  def filterScan(f: T => Boolean) {
+    var i = 0
```
benblack86: I think you should be able to use an iterator instead of tracking the index yourself.

colossus/src/main/scala/colossus/streaming/Source.scala
```diff
@@ -203,6 +253,8 @@ class DualSource[T](a: Source[T], b: Source[T]) extends Source[T] {
     }
   }
 
+  def peek = if (a_empty) b.peek else a.peek
```
benblack86: `aEmpty` 🐫

PR #460, opened 2016-10-24
Replace all connection handler backpressure logic with pipes

DanSimon:
Once in a blue moon you really end up killing two birds with one stone.  In a nutshell, this PR replaces all of the connection handler buffering and backpressure logic with the new reincarnation of pipes.  The end result is much cleaner code and the generalization of a pattern found in several places (buffering and backpressure) into a single solution.

### Preface

`Pipe[I,O]` was originally created for streaming services: the idea being that a request or response could be streamed in/out without having to load the entire thing in memory before using it.  `Pipe` was the representation of the streaming data, where pieces of the message could be "pushed" into and "pulled" out of.  Originally created in version 0.6.0, `Pipe`has been heavily reworked and dramatically improved in the upcoming 0.9.0 release.  It has essentially transformed into a buffer with a monad-like interface that can efficiently handle backpressure.

Meanwhile, buffering and backpressure has been a central concern in the controller and service layers of the connection handler architecture.  To recap, the `Controller` layer is the part of a connection handler responsible for reading in raw data and decoding it into message objects like Http requests, and vise versa for encoding messages into raw data.  The `Service` layer takes the incoming request messages and processes them into response messages, dealing with all the logic to make sure responses are sent back in the right order.

Up until now buffering and backpressure between these layers have been handled somewhat clumsily.  For instance, `Controller` has a `push`method that would return `true`or `false` depending on whether the pushed message was successfully buffered or not.  It was then up to the service layer to deal with the result manually.  This also led to `Controller`having a lot of exposed methods like `pauseReads()` and `pendingBufferSize` that led to a lot of weird logic, and probably a lot of weird edge cases.

### Using Pipes in Controller and ServiceServer/ServiceClient

Well now all that is gone.  Instead now pipes are used as the interface between layers.  So now, for example, the service layer can simply access an `outgoing` `Sink` from the controller and push responses there.  Likewise, the controller uses an `incoming` `Sink` to push decoded messages.

On the client side, the biggest change is `ServiceClient` now maintains its own buffer for pending requests.  Previously it used the controller's buffer to do this, but again that led to a lot of logic bleedthrough.  But now with pipes the client can simply route it's pending pipe into the controller's outgoing pipe.

### Additional Changes to Pipes

A few changes were made to pipes to get this all working.

The biggest change is to `Source.pullWhile`.  Instead of taking a function of type `T => Boolean`, it now takes a `T => PullAction`.  This gives the looping function greater control of the iteration process.  Most importantly, it can now decide to pull an item after receiving it by returning `PullAction.PullContinue`, which was very important when feeding a `Source` into a `Sink`.  

A new method `pullUntilNull` has been created which is similar to `pullWhile` but is intended to be used to pull out any items currently buffered and not wait for more items afterwards.  It is mainly used by `OutputController` to pull out items to write to the output buffer.

Another new method `Sink.valve`was created which essentially creates a new sink that splits into two sinks.  An item is only pushed if it can be pushed to both sub-sinks.  This is mainly used by `ServiceClient` to properly handle the `sentBufferSize` to limit the number of concurrently in-flight requests.  

### Performance

After some basic benchmarking, there appears to be no change in performance for a basic service.  I haven't done extensive benchmarking on the client yet, but initial results don't show anything drastic.

### What's Next

This PR clearly the first of several, as there's definitely some cleanup work to do, but this PR is already large enough as is.


colossus-examples/src/main/scala/colossus-examples/WebsocketExample.scala
```diff
@@ -82,13 +83,17 @@ object WebsocketExample {
           case "MANY" => {
```
benblack86: Is there one proxy actor per connection? It seems that handling "MANY" will be blocking until count equals zero so if the client tries to send another message (like "EXIT) it wouldn't be handled immediately. At least in this example, the logic is the same in OK and FULL state. Is there a scenario where we would want a different behaviour if full?

DanSimon: Yes, every connection gets a unique proxy actor.  Also in this case "MANY" will block only until the output buffer fills up (which I believe defaults to 50).  When it comes to blocking, it's true that while the output buffer is being filled, nothing else can occur.  In this situation it doesn't matter since it happens very quickly.  I'd imagine if generating the messages to send required significant CPU usage, it would be done on a different thread.

I have thought about ways to have more control over the "greediness" of pipes and situations like this.  One potential is to create a "deferred" wrapper for a pipe which will force the return of a "Full" result after a certain number of items have been pulled and fire off the returned signal at the beginning of the next event loop iteration (using a CallbackExecutor).  This would let you, for example, force the "MANY" situation to only send one item per event loop iteration.

I think in most cases when it comes to dealing with a "Full" vs "Ok" result, you just want to wait until the underlying buffer can accept more items and continue from there.  In fact, the code here can be cleaned up a lot by just wrapping this in an iterator and forwarding it into the output buffer directly using `Source.into`.  I may update or add to this example to show that.

DanSimon:
Added some more tests and fixed a few bugs.  Test coverage is not quite up to par, but I'm a bit hesitant to add more tests until more cleanup has been done and all the interfaces are more stable.

PR #461, opened 2016-10-26
Minor Pipe cleanup and some small optimizations

DanSimon:
After doing a bit more benchmarking and profiling, found a few small things to cleanup.

The largest change is that `pipe.pullWhile` now requires two functions instead of one.  The first function is used whenever an item is ready to be pulled, and the second is used when the pipe is closed/terminated.  Not only does this (slightly) improve performance by vastly reducing how often an item needs to be wrapped in `PullResult.Item`, but it cleans up the logic since a `PullAction` isn't actually needed when the pipe is closed or terminated.

The other changes are small optimizations and some updated docs.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -197,7 +197,7 @@ class ParsedHttpHeaders(
     connection: Option[Connection]
   ) = this(headers.headers, transferEncodingOpt, contentLength, connection)
 
-  override def transferEncoding = transferEncodingOpt.getOrElse(TransferEncoding.Identity)
+  override def transferEncoding = if (transferEncodingOpt.isDefined) transferEncodingOpt.get else TransferEncoding.Identity
 
 }
 
--- a/colossus/src/main/scala/colossus/service/ServiceServer.scala
+++ b/colossus/src/main/scala/colossus/service/ServiceServer.scala
@@ -248,8 +248,8 @@ with UpstreamEventHandler[ControllerUpstream[Encoding.Server[P]]]
   }
 
   def processMessages() {
-    incoming.pullWhile {
-      case PullResult.Item(request) => {
+    incoming.pullWhile (
+      request => {
         numRequests += 1
         val promise = new SyncPromise(request)
         requestBuffer.add(promise)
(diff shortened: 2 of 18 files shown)


colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -197,7 +197,7 @@ class ParsedHttpHeaders(
     connection: Option[Connection]
   ) = this(headers.headers, transferEncodingOpt, contentLength, connection)
 
-  override def transferEncoding = transferEncodingOpt.getOrElse(TransferEncoding.Identity)
+  override def transferEncoding = if (transferEncodingOpt.isDefined) transferEncodingOpt.get else TransferEncoding.Identity
```
amotamed: assuming this is for performance reasons, right?  the first version is a bit more readable imho but, I'm cool with it if the benchmarks improve with it

colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -211,18 +211,22 @@ with UpstreamEventHandler[ControllerUpstream[Encoding.Server[P]]]
...
         case PushResult.Ok => {
-          val done = requestBuffer.remove()
           if (requestMetrics) {
             concurrentRequests.decrement()
             val tags = tagDecorator.tagsFor(done.request, done.response)
             requests.hit(tags = tags)
             latency.add(tags = tags, value = (System.currentTimeMillis - done.creationTime).toInt)
           }
-          upstream.outgoing.push(done.response)
+          //upstream.outgoing.push(done.response)
```
amotamed: I'd nuke that line

colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -248,8 +252,8 @@ with UpstreamEventHandler[ControllerUpstream[Encoding.Server[P]]]
   }
 
   def processMessages() {
```
amotamed: minor issue, mind adding the return type to these functions?  it's a bit hard to follow

amotamed:
👍  looks good, could you do a team talk about colossus though?  there's nuances I'm sure I've missed

PR #462, opened 2016-11-01
fixing bug in retrypolicy with duration overflow

DanSimon:
This fixes an odd bug that occurred on exponential retry policies.  Even though there is a max value, it continued to calculate values larger than the max, eventually resulting in an exception being thrown by `FiniteDuration` once the number of attempts got high enough.  This ensures that once the max value is hit, it doesn't attempt to further recalculate an exponential value.


amotamed:
👍

benblack86:
👍

PR #463, opened 2016-11-04
fixing a bug where fatal errors failed to initiate reconnect logic

DanSimon:
Because `ServiceClient` didn't override `fatalInputError`, it would end up permanently closing the connection instead of going through the proper reconnection process whenever a parsing error occurred.  This would include situations where a parsed response exceeded the configured `maxResponseSize` of the client.


amotamed:
👍

benblack86:
👍

PR #464, opened 2016-11-08
Docs: Worker updates

benblack86:
Apart from minor changes (orange_file.gif doesn't exist, and slash on metric names is unnecessary) I have added a section about how workers and their request handlers/clients can get updates. I thought I'd add a new page since it didn't seem appropriate for the quick start page where workers/initializers are discussed.


DanSimon:
This is definitely a good start.  I want to add some docs about architecture in general so this will probably get added to that.

Issue #466, opened 2016-11-17
Http client should automatically include a host header


DanSimon:
As per the [http spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23), Http 1.1 requires this header.  We should have the client generate this header automatically.

This can probably be accomplished by adding a filter on top of the base client that adds the header to each request, before it gets wrapped in the `HttpClient` interface.  Alternatively, we could do this at the codec level, where the host header is set when creating the codec and is automatically added when encoding the request, similar to how content-length is encoded.

PR #468, opened 2016-11-20
Include the host header in HTTP requests

izeigerman:
This is a proposed fix for the Host header issue (https://github.com/tumblr/colossus/issues/466). It was a little bit tricky to forward the destination address information from the lower level (`ServiceClient`) to a protocol-specific implementation. That's why I had to update the `ClientLifter` and `LiftedClient` interfaces. I'm pretty sure this update can potentially be useful for other implementations in future. So far I tested it manually, but if my updates make sense I can also introduce some unit tests. The only detail that bothers me here is [apply](https://github.com/izeigerman/colossus/blob/96f1643b08282d5461169070dd567f9865d4ad69/colossus/src/main/scala/colossus/service/ServiceDSL.scala#L222-L224) method. It's definition allows user to specify a client configuration that differs from the sender's one. I'm not sure how to fix this properly (perhaps declare it `private[colossus]` since it's only used internally).

### DIFF
--- a/colossus-tests/src/test/scala/colossus/Util.scala
+++ b/colossus-tests/src/test/scala/colossus/Util.scala
@@ -44,7 +44,7 @@ object RawProtocol {
 
   implicit object RawClientLifter extends ClientLifter[Raw, RawClient] {
     
-    def lift[M[_]](client: Sender[Raw,M], clientConfig: ClientConfig)(implicit async: Async[M]) = {
+    def lift[M[_]](client: Sender[Raw,M], clientConfig: Option[ClientConfig])(implicit async: Async[M]) = {
       new BasicLiftedClient(client, clientConfig) with RawClient[M]
     }
   }
--- a/colossus-tests/src/test/scala/colossus/Util.scala
+++ b/colossus-tests/src/test/scala/colossus/Util.scala
@@ -44,7 +44,9 @@ object RawProtocol {
 
   implicit object RawClientLifter extends ClientLifter[Raw, RawClient] {
     
-    def lift[M[_]](client: Sender[Raw,M])(implicit async: Async[M]) = new BasicLiftedClient(client) with RawClient[M]
+    def lift[M[_]](client: Sender[Raw,M], clientConfig: ClientConfig)(implicit async: Async[M]) = {
+      new BasicLiftedClient(client, clientConfig) with RawClient[M]
+    }
   }
 
   object Raw extends ClientFactories[Raw, RawClient]{
--- a/colossus-tests/src/test/scala/colossus/protocols/http/HttpClientSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/protocols/http/HttpClientSpec.scala
@@ -1,0 +1,50 @@
+package colossus
+package protocols.http
+
+import java.net.InetSocketAddress
+
+import colossus.metrics.MetricAddress
+import colossus.service.{ClientConfig, Sender}
+import colossus.testkit.ColossusSpec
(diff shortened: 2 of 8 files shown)


DanSimon:
This is an interesting approach and one I hadn't thought of.  You're absolutely right that this touches on a larger issue of being able to easily add protocol-specific behavior and configuration to clients.  I think some of the changes we've made in version 0.9.0 (in the develop branch) will allow us to provide a slightly cleaner solution, but I think this approach will work as 

The main issue I see right now is that `Sender` is intended to be a very generic interface for anything that can turn a request into a response, and "lifting" a sender allows protocols to wrap a sender in a protocol-specific interface.  So for example a `LoadBalancingClient` can be lifted into a `HttpClient` using the constructor `Http.client(loadbalancer)` , but in that case it doesn't really make sense to pass a `ClientConfig` as a second parameter when doing that.  It also means we can't really make that `apply` version private.  Lastly in this case it would mean the header-adding logic could happen twice, once on the load-balancer itself and once on the underlying connection the request is routed to.  Now luckily your change accounts for a pre-existing header, but it would be nice to have it execute only once.

I think the best way to handle this for now is to make the new `ClientConfig` parameter in `lift` an `Option[ClientConfig]`, and have versions of the `apply` method like:

```scala
def apply(config: ClientConfig)(implicit provider: ClientCodecProvider[C], env: E): T[M] =  {
  apply(baseFactory(config), config)
}

def apply(sender: Sender[C,M], clientConfig: ClientConfig)(implicit env: E): T[M] = {
  lifter.lift(sender, Some(clientConfig))(builder.build(env))
}
def apply(sender: Sender[C,M])(implicit env: E): T[M] = {
  lifter.lift(sender, None)(builder.build(env))
}
```

And likewise have the config in `HttpClient` be an option.  Now we can properly have the header added when the client is created directly from a `ClientConfig`, but not when the client is going to wrap a pre-existing client (that should have been created from a config).  This will also allow us to maintain backwards compatibility and allow users to opt-out of adding the header if they want.  

Take a shot at making these changes and we'll see if it works.  Also see if you can rework things so you don't have to construct the header each time (you can create the `HttpHeader` once and use `request.withHeader(header)`, since it'll be identical for every request going through the connection).

izeigerman:
@DanSimon thank you for such a detailed response, appreciate it. I didn't completely understand: have you fixed this issue in 0.9.0 branch (or updated the interface so that this issue can be fixed easily)? If yes I won't continue working on this, because I believe your solution is going to be better. But if no - I'll be glad to fix this.

DanSimon:
Ah sorry for the confusion, the issue is definitely not fixed in 0.9.0.  What I meant was in that branch we're making a big effort to make things more modular and composable, and I think those changes might make it easier to implement a solution in a more generalized way.  But I think the approach you're taking here is perfectly fine, especially for the next 0.8.x release.

DanSimon:
> This also means that clientConfig in LiftedClient should be an Option as well. Are you OK with that?

Yeah I think that's fine.

izeigerman:
@DanSimon ok, I guess I addressed all the comments.

DanSimon:
This looks good to me 👍  Tests look good too

I'll need you to sign our [CLA](http://static.tumblr.com/zyubucd/GaTngbrpr/tumblr_corporate_contributor_license_agreement_v1__10-7-14.pdf) (standard procedure for all our public repos) before I can merge this branch, but once that's done we can get it in for the next release 0.8.3.

izeigerman:
@DanSimon the CLA has been sent.

PR #470, opened 2016-12-02
LoadbalancingClient supports multiple connections to the same host

DanSimon:
Previously a `LoadBalancingClient` could only have one open connection per host, due to the fact that it was backed by a hash map using the address as the key.  This could be a problem, for example, if you're trying to load balance requests across multiple connections through a proxy.  This has now been changed to use an `ArrayBuffer` instead.

This is not a perfect solution, since although new connections to a host can be added one at a time, they can only be removed all at once.  This ensures backwards compatibility and does solve the use-case of creating a load-balancer with multiple connections to a single host, but further work will probably be needed down the road to support other use cases.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/service/LoadBalancingClientSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/service/LoadBalancingClientSpec.scala
@@ -27,7 +27,11 @@ class LoadBalancingClientSpec extends ColossusSpec with MockitoSugar{
 
   implicit val executor = FakeIOSystem.testExecutor
 
-  type C = ServiceClient[PR]
+  type C = Sender[PR, Callback]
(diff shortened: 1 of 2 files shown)


benblack86:
I don't think this is going to work for the way we are using update. For example, in pollscala we are polling config every x seconds and calling update on each of the load balancers with the assumption is wouldn't add new clients if the address already exists (see PollscalaService). With this design it looks like the list will will grow over time.

DanSimon:
Ah good point.  I have added a new optional `allowDuplicates` param to `update` that defaults to `false`.  So now the default behavior is as we'd expect in that situation.  Also added another test for that use case.

benblack86:
As a temporary solution 👍

PR #472, opened 2016-12-06
Move client Handler creation outside of ServiceClient

DanSimon:
This is a minor refactoring that moves some initialization logic out of `ServiceClient` and into `ServiceClientFactory`.  There will probably be more changes like this, but this one's been sitting unmerged for a couple weeks so I'll get it in now before I forget about it.

The change follows suit with the general ideas that 1) `ServiceClient`is now only one component of a client connection handler vs the whole thing, 2) that the structure of a client handler can change from codec to codec, and 3) that users no longer directly create or interact with `ServiceClient` but rather do so through a codec-specific interface like `HttpClient`.

This also means that `ServiceClient` itself is no longer responsible for binding the full connection handler to a worker.


benblack86:
👍

PR #473, opened 2016-12-07
You need to import the ActorSystem

tom-martin:


DanSimon:
Thanks for pointing this out, I'll regen the docs after merging. 👍

PR #474, opened 2016-12-17
fixing bug that caused clients to fail connection completion on Windows

DanSimon:
Yeah you read that right.  I've been messing around with the windows subsystem for Linux and ran into this issue pretty quickly.  Apparently this whole time client connections (probably only to localhost) have not worked on Windows because we weren't properly handling the result of `SocketChannel.connect()`.  

At least on Linux and MacOs, connecting to localhost is asynchronous, so `SocketChannel.connect()` returns false and we must subscribe the selection key to `OP_CONNECT`.  However on windows it appears that the method returns true and the connection is immediately established.  This meant not only were we never calling `Connection.handleConnected()`, but because the `OP_CONNECT` interest is apparently the same as `OP_WRITE`, it would cause the worker to go nuts constantly selecting the connection over and over with nothing to do, pegging the CPU.

Best I can tell this totally fixes the issue.  All tests pass locally and I can run all the examples.  I also had to tweak timeouts on a few tests, as it appears there is a delay when a connection on windows to localhost is refused.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/core/ConnectionHandlerSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/core/ConnectionHandlerSpec.scala
@@ -89,7 +89,7 @@ class ConnectionHandlerSpec extends ColossusSpec {
       }
     }
 
-    "automatically unbind on manual disconnect" in {
+    "automatically unbind on manual disconnect"  in {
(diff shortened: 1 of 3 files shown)


benblack86:
👍

amotamed:
👍

PR #475, opened 2016-12-19
histogram buckets can be configured through typesafe config

DanSimon:
This adds a config option for histograms that we missed when moving config to typesafe-config.  Histograms by default use 100 buckets on a log scale.  This works pretty well for latency when the average is around 100ms or lower.  But in many other cases the buckets will not give accurate percentiles, either because something with a non-log distribution is being measured or there are too few buckets to give enough resolution.

So now there are 3 possible ways to configure histogram buckets:
* `manual` - you provide list of bucket lower-bounds yourself
* `linearscale` - you provide `num_buckets` and `infinity` and the range below the infinity value is evenly divided amongst `num` buckets
* `logscale` - same as linear except the natual-log scale is used

the default is set to log-scale with 100 buckets and `Int.MaxValue` as infinity, which matches the existing defaults.

### DIFF
--- a/colossus-metrics/src/main/resources/reference.conf
+++ b/colossus-metrics/src/main/resources/reference.conf
@@ -22,6 +22,11 @@ colossus {
           percentiles: [0.75, 0.9, 0.99, 0.999, 0.9999]
           sample-rate: 1.0
           prune-empty: false
+	  buckets : {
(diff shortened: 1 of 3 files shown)


benblack86:
👍

PR #480, opened 2016-12-21
WIP - removing delegator and switching to initializer

amotamed:
This is the initial swap of initializer to the delegator, I'd like to sync up and see if we're seeing eye to eye on this change


colossus/src/main/scala/colossus/core/ServerDSL.scala
```diff
@@ -50,20 +52,18 @@ abstract class Initializer(context: InitContext) {
...
-
+object Initializer {
+  type Factory = (ServerRef, WorkerRef) => Initializer
+}
 
-  def acceptNewConnection: Option[ServerConnectionHandler] = {
-    Some(initializer.onConnect(ServerContext(server, worker.generateContext)))
-  }
+//almost seems like we don't need delegator anymore
+class DSLInitializer(server : ServerRef, _worker : WorkerRef, initializer: Initializer) extends Initializer(InitContext(server, _worker)) {
```
DanSimon: This class should be removed as well.  The worker can just create a `ServerContext` itself when it calls this method.

colossus/src/main/scala/colossus/core/Server.scala
```diff
@@ -299,7 +299,7 @@ private[colossus] class Server(io: IOSystem, serverConfig: ServerConfig,
 
 
   def binding(router: ActorRef): Receive = alwaysHandle orElse {
-    case DelegatorBroadcast(message) => router ! akka.routing.Broadcast(Worker.DelegatorMessage(me, message))
+    case DelegatorBroadcast(message) => router ! akka.routing.Broadcast(Worker.InitializerMessage(me, message))
```
DanSimon: This case class needs to be changed.  There's also a method in `ServerRef` called `delegatorBroadcast`, that method should _not_ be removed, but marked as deprecated, and then also create it's replacement `initializerBroadcast`.

colossus/src/main/scala/colossus/core/ServerDSL.scala
```diff
@@ -50,20 +52,18 @@ abstract class Initializer(context: InitContext) {
 
 }
 
-//almost seems like we don't need delegator anymore
-class DSLDelegator(server : ServerRef, _worker : WorkerRef, initializer: Initializer) extends Delegator(server, _worker) {
-
+object Initializer {
+  type Factory = (ServerRef, WorkerRef) => Initializer
```
DanSimon: This type should just be `InitContext => Initializer`

colossus/src/main/scala/colossus/core/Worker.scala
```diff
@@ -238,25 +238,25 @@ private[colossus] class Worker(config: WorkerConfig) extends Actor with ActorLog
...
         .getOrElse{
-          log.error(s"delegator message $message for unknown server ${server.name}")
+          log.error(s"initializer message $message for unknown server ${server.name}")
         }
     }
-    case NewConnection(sc, attempt) => delegators.get(sender()).map{delegator =>
-      delegator.createConnectionHandler.map{handler =>
-        registerConnection(sc, delegator.server, handler)
+    case NewConnection(sc, attempt) => initializers.get(sender()).map{initializer =>
+      Try(initializer.onConnect.apply(null)).map{handler =>
```
DanSimon: 1.  This obviously needs to be fixed (this is where the `ServerContext` should be generated.
2.  It's a little disturbing that all tests pass even with this code.

PR #483, opened 2017-01-10
fix quickstart Hello World! url

modalsoul:


DanSimon:
thanks for catching this 👍

PR #486, opened 2017-01-11
metrics example

dre-hh:


dre-hh:
not consistent to examples setup

PR #487, opened 2017-01-11
added doc for metrics config and fixed counter.set args order

dre-hh:
counter.set() has different arguments order now
meetrics system allways expects a typesafe config on initialization


DanSimon:
looks good 👍

PR #490, opened 2017-01-20
moving some http streaming files around, reorganizing package

DanSimon:
No actual code changes here, just doing some house-keeping on some of the http streaming files.


DanSimon:
Going to ignore the coverage check on this PR since I don't want to mix moving code around with actual changes.

PR #491, opened 2017-01-20
code cleanup

DanSimon:
This is mostly just removing some old code and cleaning up warnings.

The only significant change is the addition of a new `kill` method to `ConnectionManager`, which gives connection handler the ability to kill their attached connection with an error.  This functionality has already existed but required directly sending a `WorkerCommand` message to the handler's worker.


benblack86:
👍

PR #497, opened 2017-02-13
updating docs to include metrics information

amotamed:
@DanSimon @benblack86 @aliyakamercan 

Adding more documentation for rates.  Rate types and default rates are now listed.  Added an example of metric ticks inside a request.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -12,6 +12,7 @@ per second, which can easily translate to millions of recordable events per
 second.  The Metrics library provides a way to work with metrics with as little
 overhead as possible. It is configurable via a [typesafe config](https://github.com/typesafehub/config).
 See [reference config](https://github.com/tumblr/colossus/blob/master/colossus-metrics/src/main/resources/reference.conf)
+The Colossus server and clients define several metrics be default.
 
 ### Collection Intervals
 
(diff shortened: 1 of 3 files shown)


docs/metrics.md
```diff
@@ -24,6 +24,48 @@ of the last second of activity, whereas the 1 minute intervals are used for
 reporting values to an external database in which case the values reported are
 all reflective of the last minute of activity.
 
+### Default Metrics
+
+Colossus makes several metrics available in a default service, they are:
+
+* *service name*/system/fd_count
```
benblack86: It would be good to explain what each of these means and which part of colossus they come from (and maybe tags that can be found on them?)

docs/metrics.md
```diff
@@ -10,7 +10,7 @@ Colossus uses the (currently nameless) metrics library.
 High-throughput Colossus services can serve hundreds of thousands of requests
 per second, which can easily translate to millions of recordable events per
 second.  The Metrics library provides a way to work with metrics with as little
-overhead as possible.
+overhead as possible.  Colossus services define several metrics be default.
```
DanSimon: _by_ default

docs/metrics.md
DanSimon: Also these are only the metrics for a server, clients have their own set of metrics not included here.

Also, this is a nitpick, but remember that `colossus-metrics`is a separate library from colossus itself, so this section doesn't apply for people using colossus-metrics but not colossus.  So maybe change the title to "Default Metrics in Colossus"

docs/metrics.md
```diff
@@ -24,6 +24,48 @@ of the last second of activity, whereas the 1 minute intervals are used for
...
+* *service name*/worker/connections
+* *service name*/worker/rejected_connections
+* *service name*/connections
+* *service name*/refused_connections
+* *service name*/connects
+* *service name*/closed
+* *service name*/highwaters
+
+
+### Metric Types
```
DanSimon: I think this section is kinda redundant with the "Available Collectors" section below.

amotamed:
bump, updated per comments

DanSimon:
Looks good to me but there appear to be merge conflicts.  Maybe you didn't do a git pull before branching?  Anyway once that's fixed this is 👍

benblack86:
👍

PR #500, opened 2017-02-17
adding LoadBalancingClient documentation

amotamed:

Adding documentation about failed calls and load balancing. I tried to make it informative as well as terse.

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -4,7 +4,7 @@ title: Clients
 ---
 
 Clients are used to open connections to external systems.  Colossus currently
-has built-in support for Http, Memcached, and Redis, but similar to servers, you
+has built-in support for Http, Memcached, and Redis. Similar to servers, you
 can add support for any protocol by implementing the necessary traits and typeclasses.
 
 ## Client Behavior in a Nutshell
@@ -68,6 +68,44 @@ Therefore these two rules may help when choosing what to use:
 
 * When opening a client within a service, use local clients.  Usually you'll want to open one client per worker.
 * When opening a client from outside a service, such as from inside an actor or some other general use case, use a future client.
+
+## Failure Management
+ 
+### Client Behavior 
+ 
+If any failure occurs (example: connection closed, external system isn't 
+responding) the response is sent back as a failure and NOT retried.  The 
+default behavior is to continue trying subsequent requests.  If failfast is
+enabled then all queued requests in the ServiceClient are immediately failed
+and removed.  
+
+### Retrying Requests
+
+To enable retries, the LoadBalancingClient (LBC) should be used.  The LBC takes
+in a generator that builds a ServiceClient from an InetSocketAddress and a list
+of InetSocketAddresses.  The LBC defaults to using an unused host for a request
+or the max retry number, whichever is lower.  An example: if the retry count is
+set to four but there are only two hosts, it'll only attempt a request twice.
+
+Any failure occurring before the client receives the response (example: 
+connection closed, host not responding) is treated as an attempt and is 
+retried on the next connection.  If the maximum number of tries are exhausted, 
+a SendFailedException is returned with the last exception.
+
+### Example
+
+
+Below is an example of retrying three times to the same host
+
+{% highlight scala %}
+val generator = (address: InetSocketAddress) => {
+  Redis.client(address.toString, 6379)
+}
+val clients = List.fill(3)("myredisserver")
+val lbc = new LoadBalancingClient[Redis](worker, generator, 3, clients)
+
+Redis.client(lbc).zadd(ByteString("key"), ByteString("value"))
+{% endhighlight %}
 
 ## Using clients generically
 


docs/clients.md
```diff
@@ -69,6 +69,43 @@ Therefore these two rules may help when choosing what to use:
...
+is returned with the last exception.
+
+### Example
+
+
+Below is an example of retrying three times to the same host
+
+{% highlight scala %}
+val generator = (address: InetSocketAddress) => {
+  Redis.futureClient(address.toString, 6379)
```
DanSimon: you can't use `FutureClient` here, it has to be `Redis.client`.

docs/clients.md
```diff
@@ -69,6 +69,43 @@ Therefore these two rules may help when choosing what to use:
...
+}
+val clients = List.fill(3)("myredisserver")
+val lbc = new LoadBalancingClient[Redis](worker, generator, 3, clients)
+
+lbc.send(Command.c(CMD_ZADD, ByteString("key"), ByteString("value")))
```
DanSimon: You can actually wrap a LBC inside a redis client, so by doing `val client = Redis.client(lbc)` you can have an LBC but still have the redis client interface and not have to construct commands like this.

docs/clients.md
```diff
@@ -69,6 +69,43 @@ Therefore these two rules may help when choosing what to use:
 * When opening a client within a service, use local clients.  Usually you'll want to open one client per worker.
 * When opening a client from outside a service, such as from inside an actor or some other general use case, use a future client.
 
+## Failure Management
+ 
+### Client Behavior 
+ 
+A Client handles makes a request to an external system.  If any failure occurs
```
DanSimon: lol I have no idea what this first sentence is trying to say.

docs/clients.md
```diff
@@ -69,6 +69,43 @@ Therefore these two rules may help when choosing what to use:
...
+(example: connection closed, external system isn't responding) the response is 
+sent back as a failure and NOT retried.  The default behavior is to continue
+trying subsequent requests.  If failfast is enabled then all subsequent 
+requests that are queued up in the ServiceClient are removed.  
+
+### Retrying Requests
+
+To enable retries, the LoadBalancingClient (LBC) should be used.  The LBC takes
```
DanSimon: I think it's worth pointing out that a LBC can open multiple connections, so it can be used for more than just reties....like actual load balancing.

docs/clients.md
```diff
@@ -69,6 +69,43 @@ Therefore these two rules may help when choosing what to use:
...
 
+## Failure Management
+ 
+### Client Behavior 
+ 
+If any failure occurs (example: connection closed, external system isn't 
+responding) the response is sent back as a failure and NOT retried.  The 
+default behavior is to continue trying subsequent requests.  If failfast is
+enabled then all subsequent requests that are queued up in the ServiceClient
+are removed.
```
benblack86: I would be more explicit:

If failfast is enabled then all queued requests in the ServiceClient are immediately failed.

docs/clients.md
```diff
@@ -69,6 +69,43 @@ Therefore these two rules may help when choosing what to use:
...
+are removed.  
+
+### Retrying Requests
+
+To enable retries, the LoadBalancingClient (LBC) should be used.  The LBC takes
+in a generator that builds a ServiceClient from an InetSocketAddress and a list
+of InetSocketAddresses.  The LBC defaults to using the Integer max retry as the
+default connection attempts.
+
+Any failure occurring before the the client receives the response (example:
```
benblack86: the the

docs/clients.md
```diff
@@ -69,6 +69,43 @@ Therefore these two rules may help when choosing what to use:
...
+### Retrying Requests
+
+To enable retries, the LoadBalancingClient (LBC) should be used.  The LBC takes
+in a generator that builds a ServiceClient from an InetSocketAddress and a list
+of InetSocketAddresses.  The LBC defaults to using the Integer max retry as the
+default connection attempts.
+
+Any failure occurring before the the client receives the response (example: 
+connection closed, host not responding) is treated as an attempt and is 
+retried.  If the maximum number of tries are exhausted, a SendFailedException
```
benblack86: retried on the next connection?

amotamed:
bump

docs/clients.md
```diff
@@ -69,6 +69,42 @@ Therefore these two rules may help when choosing what to use:
...
 * When opening a client from outside a service, such as from inside an actor or some other general use case, use a future client.
 
+## Failure Management
+ 
+### Client Behavior 
+ 
+If any failure occurs (example: connection closed, external system isn't 
+responding) the response is sent back as a failure and NOT retried.  The 
+default behavior is to continue trying subsequent requests.  If failfast is
+enabled then all queued requests in the ServiceClient are immediately removed.
```
benblack86: "immediately failed and removed" would be better. Just "removed" suggests that no response is sent.

docs/clients.md
```diff
@@ -69,6 +69,42 @@ Therefore these two rules may help when choosing what to use:
...
+If any failure occurs (example: connection closed, external system isn't 
+responding) the response is sent back as a failure and NOT retried.  The 
+default behavior is to continue trying subsequent requests.  If failfast is
+enabled then all queued requests in the ServiceClient are immediately removed.  
+
+### Retrying Requests
+
+To enable retries, the LoadBalancingClient (LBC) should be used.  The LBC takes
+in a generator that builds a ServiceClient from an InetSocketAddress and a list
+of InetSocketAddresses.  The LBC defaults to using the Integer max retry as the
```
benblack86: "Integer max retry" - that's true, but I don't think it tries 2,147,483,647 times. I think it tries as many times as there are connections, so if there are three connections it will try three times. @DanSimon am I correct?

amotamed: yeah, it's a combination of the retries and available hosts, I'll clarify that

DanSimon:
I realized there appears to be nothing here about using a `RetryPolicy` to control how a client handles attempting to connect.  It's kind of implied in the connection management section, but it would be good to include an example or two to show how to control reconnect attempts.

Doesn't have to go into this PR, but definitely something we should add.

benblack86:
👍

PR #502, opened 2017-02-23
adding retry policy documentation

amotamed:

I've added a small section on the retry policy and retry types

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -72,12 +72,19 @@ Therefore these two rules may help when choosing what to use:
 ## Failure Management
  
 ### Client Behavior 
+
+Failures can be broken down into two categories: Connection Failures and 
+Request Failures.  Regardless of which type, if a failure occurs the response
+is sent back as a failure and NOT retried.  If failfast is enabled, then all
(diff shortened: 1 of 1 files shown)


docs/clients.md
```diff
@@ -107,6 +107,18 @@ val lbc = new LoadBalancingClient[Redis](worker, generator, 3, clients)
 Redis.client(lbc).zadd(ByteString("key"), ByteString("value"))
 {% endhighlight %}
 
+### Retry Policy
+
+Each service client takes in a retry policy. The client defaults to an
```
DanSimon: It should be a bit more clear in this section that the RetryPolicy is used for retrying to connect, not for retrying a request.  Also we should start linking more frequently to the scaladocs, for example, here are the docs for [RetryPolicy](https://tumblr.github.io/colossus/api/index.html#colossus.core.RetryPolicy) and [BackoffPolicy](https://tumblr.github.io/colossus/api/index.html#colossus.core.BackoffPolicy)

amotamed:
@DanSimon I broke out the error types into Request Failures and Connection Failures to be more explicit

amotamed:
bump

PR #503, opened 2017-02-24
fixing histogram values to not be higher than max

amotamed:

Since our histogram values are approximations, we occasionally run into times when the approximation calculation is greater than the max value.  Since there's no issue in the underlying approximation calculation (tradeoff between accuracy and data size to keep) I'm setting a hard limit on max values.


benblack86:
👍

DanSimon:
I don't know what the deal is with codecov, but all tests pass and I think this change makes a lot of sense so a big 👍  from me.

PR #504, opened 2017-02-27
Fixes to Pipe filterScan, CircuitBreaker Error suppression

DanSimon:
So this fixes to main issues:

1.  `Pipe.filterScan` was not correctly triggering push signals.  In a nutshell, when a pipe is full, agents trying to push to the pipe will get a `PushResult.Full` containing a signal that they can map on for when the pipe clears up and is again pushable.  `filterScan` is a method on `BufferedPipe` to traverse buffered items and remove any that don't pass a filter function.  This is used by `ServiceClient`'s pending and sent buffers.  The issue was that when `filterScan` opened up a buffer, the push signals were not triggered.  This lead to a bug where `ServiceClient` would not properly resume sending requests if it disconnected and reconnected while the buffer was full.

2.  `CircuitBreaker` now suppresses all closes/terminations.  This type mutably wraps a Pipe and suppresses its error states.  In particular the goal is for two circuitbreakers to link together, and even if the underlying pipes terminate/close, the link between the circuitbreakers is not severed.  This is used, for example, to connect `ServiceClient` to its underlying `Controller`.  In general this prevents things like disconnects from accidentally severing these links.  However we still need to detect when the underlying Pipe terminates/closes, so a new `onBreak` method has been added to deal with such states.  This is used, for example, in `StreamTranscodingController` to close the connection when a stream hits an error.

Issue 2 is leading me to believe that we need a different interface than pipes to handle the message passing between connection handler layers, something with the same back-pressure semantics but without the notion of termination/closed, which only makes sense for actual streams of data.  For now though these changes will get us the functionality we want.

### DIFF
--- a/colossus/src/main/scala/colossus/streaming/CircuitBreaker.scala
+++ b/colossus/src/main/scala/colossus/streaming/CircuitBreaker.scala
@@ -31,13 +31,25 @@ trait CircuitBreaker[T <: Transport] {
 
 trait SourceCircuitBreaker[A, T <: Source[A]] extends  Source.BasicMethods[A] { self: CircuitBreaker[T] => 
 
+  private def redirectFailure(p: PullResult[A]): PullResult[A] = {
+    def emptyUnset = {
+      unset()
+      PullResult.Empty(trigger)
+    }
+    p match {
+      case PullResult.Closed => emptyUnset
+      case PullResult.Error(_) => emptyUnset
+      case other => other
+    }
+  }
+
   def pull(): PullResult[A] = current match {
-    case Some(c) => c.pull
+    case Some(c) => redirectFailure(c.pull)
     case None => PullResult.Empty(trigger)
   }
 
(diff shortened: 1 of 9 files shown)


colossus/src/main/scala/colossus/streaming/CircuitBreaker.scala
```diff
@@ -48,14 +71,37 @@ trait SourceCircuitBreaker[A, T <: Source[A]] extends  Source.BasicMethods[A] {
   //inside BufferedPipe
 
   override def pullWhile(fn: A => PullAction, onc: TerminalPullResult => Any) {
+
+    //so we're totally ignoring the passed-in terminal result handler, since
+    //circuit breakers suppress closed/terminated
+    
+    def myonc(r: TerminalPullResult): Any = {
```
benblack86: What is a `myonc`?

PR #505, opened 2017-03-01
allow requests up to 20mb in size to be parsed

dxuhuang:
@DanSimon @benblack86 

Allow for HttpRequestParser to accept requests up to 20MB in size.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
@@ -16,7 +16,7 @@ object HttpRequestParser {
     head.headers.transferEncoding match {
       case TransferEncoding.Identity => head.headers.contentLength match {
         case Some(0) | None => const(HttpRequest(head, HttpBody.NoBody))
-        case Some(n) => bytes(n, 20.MB, 1.MB) >> {body => HttpRequest(head, new HttpBody(body))}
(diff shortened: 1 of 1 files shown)


benblack86:
As a quick fix 👍 

It would be nice if the service could handle the situation more gracefully. For example, the value for the http request parser should be the same as the service level:

```
colossus.service.imgreg.max-request-size = "20 MB"
```

Even after this change, if someone sends a 30MB file then a colossus exception will be thrown instead of gracefully rejecting the request with a nice error message.

dxuhuang:
bump @DanSimon

DanSimon:
After some thought, I think we should increase the max to something much higher, like 1GB.  Essentially this limit isn't needed at all since the connection handler is already tracking the size of messages and will prevent the parser from going over the limit.  The only important limit is the initial buffer size (the second parameter), which I actually think we should significantly _decrease_ to like 1KB.  The buffer is reused between requests on the same connection and once it increases it doesn't decrease (another thing we should change!!), so in cases where we only have small requests we're wasting a lot of unused memory.

I also agree that we need to unify this with the configurable max message limits.  That is a slightly more invasive change but I don't think it will be too difficult.

dxuhuang:
@DanSimon I pushed the change with the hardcoded limits (1 GB for request size, 1 KB for initial buffer). I tried to make the change for passing in the size limit from the service config as suggested, but I got stumped.

DanSimon:
Ok, this should be good for now 👍

PR #508, opened 2017-05-22
adding simple client specific examples

amotamed:
@benblack86 @aliyakamercan @DanSimon 

This PR adds very simple examples of how to use Redis, Memcached, and Http clients


benblack86:
Can you add a little more to the examples to demonstrate handling the asyncResult? Also you should point out these examples are using the callback version of the clients, so "requires an implicit workerRef" is not totally correct.

amotamed:
done!  bump

DanSimon:
looks fine to me 👍

PR #512, opened 2017-07-07
adding logging on failed insert

amotamed:
@DanSimon @benblack86 @aliyakamercan 

Logging the error when serviceclient's queue is full

### DIFF
--- a/colossus/src/main/scala/colossus/controller/OutputController.scala
+++ b/colossus/src/main/scala/colossus/controller/OutputController.scala
@@ -3,10 +3,12 @@ package controller
 
 import core._
 import java.util.LinkedList
+
+import akka.event.Logging
+
 import scala.annotation.tailrec
 import scala.concurrent.duration._
-import scala.util.{Success, Failure}
-
+import scala.util.{Failure, Success}
 import service.{NotConnectedException, RequestTimeoutException}
 
 /**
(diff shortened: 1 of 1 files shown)


benblack86:
Instead of logging, I think it would be better to return failure. Failure the request?

DanSimon:
Hmm, is there a specific reason or symptom why this is being added?  To me it looks like `isAdded` can never be false.  The `canPush` method will return false if `msgq`  is full, and being full is the only reason why `msgq.enqueue` would return false.

Also, `ServiceClient` does properly handle the return value of `push` [here](https://github.com/amotamed/colossus/blob/e75b3142c0922a87ed1bc0512242d000d754710f/colossus/src/main/scala/colossus/service/ServiceClient.scala#L236), which ends up failing the request with a `ClientOverloadedException`.

amotamed:
@benblack86 I'm not sure if we want to do a full error or warn.  I'm totally open to discussing it

@DanSimon If we have a full queue of items, enqueue doesn't add the item and returns false.  This is to fix an issue we saw in Refrog.  Adding a Memcached node (invalidating a fifth of the cache) caused a ton of backfill requests that resulted in the backfill not finishing.

In other news: I'm officially (still with a ton of boxes) moved

DanSimon:
Right, but if the queue is full than the if-statement this code is inside will also be false, so we know for sure that it will only enqueue when `msgq` is not full.  If you look at the [canPush method](https://github.com/amotamed/colossus/blob/e75b3142c0922a87ed1bc0512242d000d754710f/colossus/src/main/scala/colossus/controller/OutputController.scala#L246) it checks to see if the queue is full and returns false if it is.

amotamed:
I totally see your point Dan, I'm closing this since I haven't been able to put the time to reproduce what we were seeing on the redis node rotation

PR #517, opened 2017-07-12
Fixing a bug with handling unexpected exceptions in Controller

DanSimon:
This fixes a bug uncovered by another bug that will be in a separate PR against master.  Basically when `ServiceClient` threw an exception, it got caught by `InputController` as an unexpected error resulting in killing the connection (which it should do).  The problem was that it was calling `disconnect` instead of `kill`, which meant `ServiceClient` would not handle the disconnect as an error and would _not_ try to reconnect.

The solution was a little more involved because in the case of server connections, calling `disconnect()` is actually what we want to do, because when pipelining responses we want the server connection to be able to send back any responses for successfully processed requests before closing the connection.

So the solution is to basically punt the decision of which function to call to the downstream handler, aka `ServiceServer` or `ServiceClient`.  So now the `onFatalError` event handler method is required to return a `FatalErrorAction` which instructs the controller how to behave.

### DIFF
--- a/colossus-tests/src/test/scala/colossus/controller/Common.scala
+++ b/colossus-tests/src/test/scala/colossus/controller/Common.scala
@@ -27,9 +27,9 @@ trait ControllerMocks extends MockFactory {self: org.scalamock.scalatest.MockFac
 
     def namespace = colossus.metrics.MetricSystem.deadSystem
 
-    override def onFatalError(reason: Throwable): Option[E#Output] = {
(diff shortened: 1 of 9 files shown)


benblack86:
👍  

Should we wait for the mapTry fix before merging and cutting RC4?

DanSimon:
This can be merged now, but we should wait for the mapTry fix to do the release.  That one should be much more straight-forward so I should have a PR ready soon.

PR #522, opened 2017-07-27
Colossus 0.9.0

benblack86:


DanSimon:
Everything in here has already been PR'd and tested, and we've been in the release candidate stage for months so I think this is good to go.  I did a quick once-over of everything and didn't see anything important that needed to be changed, though I think there's a few smaller followup tasks to do.

If the websocket merge conflicts aren't straightforward I can take a look and open a PR into this PR.

PR #524, opened 2017-07-28
Add a test that applies a predicate to response body in HttpServiceSpec

dxuhuang:

Want to try out this whole process


jlbelmonte:
👍

PR #527, opened 2017-08-01
Docs

benblack86:
First cut of the docs. There are a few sections I need some advice on since the code snippets didn't work and I couldn't get them to work - I'll add comments to the problem areas. To generate the docs:

```
sbt
project colossus-docs
paradox
```

Then open colossus-docs/target/paradox/site/main/index.html.

Before I merge I want to add a script that will generate multiple versions of the doc. The plan is to have a config file that will list the versions we want to generate. For example, right now it would only contain 0.9.0, but as we add versions (i.e. 0.9.1) we would add to the config list. The script would then iterate over each version, check out the tag and generate the docs. It would then push all the docs for each version to gh-pages. Since it would generate the docs for each version, if the style changes, then the change will be made to all versions, and we can update each version of the docs with the list of available docs so the user can easily switch between versions (like akka does). Having a config file list the versions means we can control which versions are generated.

The only problem I see with this method is that the docs would be tied to the tag and therefore would be hard to update without first deleting the tag and then push a new tag with the doc changes. I don't think we will want to change old versions so this seems fine to me. For 0.9.0 I will have to delete the current tag and recreate once docs have been merge so we can start it correctly. 

Alternatively, instead of having a config file we would create a new tag, for example we would have a 0.9.0 tag and a 0.9.0-doc tag. If we no longer wanted to generate that version anymore, we would delete the 0.9.0-doc tag.

@aliyakamercan @jlbelmonte @dxuhuang @amotamed @DanSimon


colossus-docs/src/main/paradox/clients.md
```diff
@@ -0,0 +1,145 @@
...
+creation.  The client defaults to an exponential backoff starting at 50 
+milliseconds and with a maximum of 5 seconds.  The policy type can be either 
+[NoRetry](https://tumblr.github.io/colossus/api/index.html#colossus.core.NoRetry) 
+or [BackoffPolicy](https://tumblr.github.io/colossus/api/index.html#colossus.core.BackoffPolicy).
+
+A Backoff retry policy contains a multiplier which can be:
+ 
+ * Constant
+ * Linear
+ * Exponential
```
benblack86: I cut the bit about "Using clients generically" (http://tumblr.github.io/colossus/docs/clients/) since it didn't work and not sure how to make it work. Guessing high kinded type needed?

colossus-docs/src/main/paradox/index.md
```diff
@@ -0,0 +1,25 @@
+@@@ index
```
benblack86: There is an io-overview.png and iosystem.png images that aren't used. Should they? Or should I delete?

colossus-docs/src/test/scala/TestkitExampleSpec.scala
```diff
@@ -0,0 +1,18 @@
+import colossus.protocols.http.HttpRequest
+import colossus.testkit.CallbackAwait
+import colossus.testkit.MockConnection
+import org.scalatest.{MustMatchers, WordSpec}
+
+import scala.concurrent.duration._
+
+// #example
+/*class TestkitExampleSpec extends WordSpec with MustMatchers {
```
benblack86: I couldn't get this example on teskit (http://tumblr.github.io/colossus/docs/testkit/) to work. Help!

benblack86:
Added the script that will generate and publish, this is roughly how it works:

1. It reads docs.yml for the tags it should checkout.
2. It creates a new navigation template file (based on https://github.com/lightbend/paradox/blob/master/themes/generic/src/main/assets/navigation.st) that adds a dropbown box (html select) that allows you to change between versions.
3. It then iterates over each tag, checks out the code, and generates the docs (paradox and colossus/colossus-metrics/colossus-testkit scaladocs).
4. Finally it creates an index.html at the base that redirects to the latest version.
5. It then uploads all this to gh-pages (code taken from previous script).

If you want to see what it looks like, look at https://benblack86.github.io/colossus where I have created 0.9.0 and 0.9.1.

I should note that I'm taking the liberty of ignore 0.8 as I don't plan on retrofitting this on versions before 0.9.0. Going forward we will keep each version.

DanSimon:
Looks like a really good start!  I'll go over it in more detail later and probably open a PR or two.  

Only thing I noticed is it doesn't look like there's a link to github anywhere, should probably add that.

benblack86:
Added github link to front page

benblack86:
Bump. Would be good to get this merged soonish. I can cut out the testkit example and fix it later if that is quicker.

DanSimon:
Yeah we can look at the testkit example separately.  This looks good to me 👍

jlbelmonte:
LGTM 👍

PR #532, opened 2017-08-08
Remove akka agent from iosystem

benblack86:
Replaced Agent with an AtomicReference. Whilst I was there I tidied up some code.


colossus-tests/src/test/scala/colossus/core/WorkerManagerSpec.scala
```diff
@@ -33,8 +34,8 @@ class WorkerManagerSpec extends ColossusSpec with Eventually{
           }
         }
 
-        val agent = Agent(Seq[WorkerRef]())
-        val manager: ActorRef = system.actorOf(Props(classOf[WorkerManager], agent, io, workerFact), name = s"idle-check-manager")
+        val agent = new AtomicReference(Seq[WorkerRef]())
+        val manager: ActorRef = system.actorOf(WorkerManager.props(agent, io, workerFact), name = s"idle-check-manager")
```
benblack86: Changed from `classOf` (which is not type safe) to the recommend props method. http://doc.akka.io/docs/akka/current/scala/actors.html#recommended-practices

DanSimon:
👍

Issue #499, opened 2017-02-17
Request body content type is always None


benblack86:
`req.body.contentType` is always `None`. Have to use `req.head.headers.firstValue("content-type")` instead.

PR #534, opened 2017-08-09
Pass http content type header to body

dxuhuang:
@benblack86 @amotamed @jlbelmonte @aliyakamercan @DanSimon 

https://github.com/tumblr/colossus/issues/499

### DIFF
--- a/colossus-tests/src/test/scala/colossus/protocols/http/HttpParseSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/protocols/http/HttpParseSpec.scala
@@ -34,7 +34,7 @@ class HttpParserSuite extends WordSpec with MustMatchers{
         )
       ), HttpBody.NoBody)
 
-      parser.parse(DataBuffer(ByteString(req))).toList must equal(List(expected))
+      parser.parse(DataBuffer(ByteString(req))) must equal(Some(expected))
     }
 
     "correctly handle fragments" in {
@@ -57,8 +57,8 @@ class HttpParserSuite extends WordSpec with MustMatchers{
         val p1 = req.take(splitIndex)
         val p2 = req.drop(splitIndex)
         try {
-          parser.parse(DataBuffer(ByteString(p1))).toList must equal(Nil)
-          parser.parse(DataBuffer(ByteString(p2))).toList must equal(List(expected))
+          parser.parse(DataBuffer(ByteString(p1))) must equal(None)
+          parser.parse(DataBuffer(ByteString(p2))) must equal(Some(expected))
         } catch {
           case t: Throwable => throw new Exception(s"Failed with splitIndex $splitIndex: '$p1' : '$p2' : $t", t)
         }
@@ -76,7 +76,7 @@ class HttpParserSuite extends WordSpec with MustMatchers{
     "parse request with content" in {
       val body = ByteString("HELLO I AM A BODY")
       val len = body.size
-      val req = s"POST /hello/world HTTP/1.1\r\nHost: api.foo.bar:444\r\nAccept: */*\r\nContent-Length: $len\r\n\r\n${body.utf8String}"
+      val req = s"POST /hello/world HTTP/1.1\r\nHost: api.foo.bar:444\r\nAccept: */*\r\nContent-Type: text/plain\r\nContent-Length: $len\r\n\r\n${body.utf8String}"
       val expected = HttpRequest(HttpRequestHead(
         method = HttpMethod.Post,
         url = "/hello/world",
@@ -84,13 +84,16 @@ class HttpParserSuite extends WordSpec with MustMatchers{
         headers = List(
           "host" -> "api.foo.bar:444",
           "accept" ->  "*/*",
+          "content-type" -> "text/plain",
           "content-length" -> len.toString
         )
-      ), HttpBody(body))
+      ), HttpBody(body, "text/plain"))
 
       val parser = requestParser
+      val parsed = parser.parse(DataBuffer(ByteString(req)))
 
-      parser.parse(DataBuffer(ByteString(req))).toList must equal(List(expected))
+      parsed must equal(Some(expected))
+      parsed.get.body.contentType must equal(Some(HttpHeader("Content-Type", "text/plain")))
     }
 
     "handle request with content-length set to 0" in {
@@ -108,7 +111,7 @@ class HttpParserSuite extends WordSpec with MustMatchers{
 
       val parser = requestParser
 
-      parser.parse(DataBuffer(ByteString(req))).toList must equal(List(expected))
+      parser.parse(DataBuffer(ByteString(req))) must equal(Some(expected))
 
     }
 
@@ -126,7 +129,15 @@ class HttpParserSuite extends WordSpec with MustMatchers{
     "parse request with authorization header" taggedAs(Broke) in {
       val auth_info = "Basic YmI6Y29vbHBhc3N3b3JkYnJvNDI="// + Base64.encodeBase64String(StringUtils.getBytesUtf8("bb:coolpasswordbro42"))
       val url = "/foo"
-      val request = HttpRequest(HttpRequestHead(version = HttpVersion.`1.1`, method = HttpMethod.Get, url = url, headers = List("authorization" -> auth_info)), HttpBody.NoBody)
+      val request = HttpRequest(
+        HttpRequestHead(
+          version = HttpVersion.`1.1`,
+          method = HttpMethod.Get,
+          url = url,
+          headers = List("authorization" -> auth_info)
+        ),
+        HttpBody.NoBody
+      )
       val d = new DynamicOutBuffer(100, false)
       request.encode(d)
       val parser = requestParser
--- a/colossus-tests/src/test/scala/colossus/protocols/http/client/HttpResponseParserSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/protocols/http/client/HttpResponseParserSpec.scala
@@ -54,22 +54,42 @@ class HttpResponseParserSpec extends WordSpec with MustMatchers {
         HttpBody(body)
       )
       parser.parse(DataBuffer(ByteString(res))) must equal(None)
-      parser.endOfStream() must equal(Some(expected))
+      val parsed = parser.endOfStream()
+      // currently there's no easy way to pass the contentType into HttpResponse.apply()
+      parsed.map(response => response.copy(body = response.body.withContentType("text/plain"))) must equal(Some(expected))
     }
 
     "parse a response with a body" in {
       val content = "{some : json}"
       val size = content.getBytes.size
-      val res = s"HTTP/1.1 200 OK\r\nHost: api.foo.bar:444\r\nAccept: */*\r\nAuthorization: Basic XXX\r\nAccept-Encoding: gzip, deflate\r\nContent-Length: $size\r\n\r\n{some : json}"
+      val res = s"HTTP/1.1 200 OK" +
+        s"\r\nHost: api.foo.bar:444" +
+        s"\r\nAccept: */*" +
+        s"\r\nAuthorization: Basic XXX" +
+        s"\r\nAccept-Encoding: gzip, deflate" +
+        s"\r\nContent-Type: application/json" +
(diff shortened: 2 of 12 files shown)


benblack86:
Code looks good, can we get some tests?

colossus-tests/src/test/scala/colossus/protocols/http/HttpParseSpec.scala
```diff
@@ -76,21 +76,24 @@ class HttpParserSuite extends WordSpec with MustMatchers{
...
           "content-length" -> len.toString
         )
-      ), HttpBody(body))
+      ), HttpBody(body, HttpHeader("Content-Type", "text/plain")))
 
       val parser = requestParser
+      val parsed = parser.parse(DataBuffer(ByteString(req)))
 
-      parser.parse(DataBuffer(ByteString(req))).toList must equal(List(expected))
+      parsed must equal(Some(expected)) // what was the reason for calling toList on the Option?
```
dxuhuang: ?

colossus-tests/src/test/scala/colossus/protocols/http/client/HttpResponseParserSpec.scala
```diff
@@ -60,16 +60,32 @@ class HttpResponseParserSpec extends WordSpec with MustMatchers {
...
+          "accept-encoding"->"gzip, deflate",
+          "content-type" -> "application/json",
+          "content-length"->size.toString),
         ByteString("{some : json}")
       )
       val data = DataBuffer(ByteString(res))
-      parser.parse(data).toList must equal(List(expected))
+      val parsed = parser.parse(data)
+
+      parsed must equal(Some(expected)) // what was the reason for calling toList on the Option?
```
dxuhuang: ?

colossus-tests/src/test/scala/colossus/protocols/http/HttpParseSpec.scala
benblack86: I'd guess, no good reason.

colossus-tests/src/test/scala/colossus/protocols/http/HttpParseSpec.scala
```diff
@@ -76,21 +76,24 @@ class HttpParserSuite extends WordSpec with MustMatchers{
...
         url = "/hello/world",
         version = HttpVersion.`1.1`,
         headers = List(
           "host" -> "api.foo.bar:444",
           "accept" ->  "*/*",
+          "content-type" -> "text/plain",
           "content-length" -> len.toString
         )
-      ), HttpBody(body))
+      ), HttpBody(body, HttpHeader("Content-Type", "text/plain")))
```
dxuhuang: have to compare the header separately since for some reason the test works even without it. `equals()` overriding?

benblack86: Hmm that's interesting. I do some several equal methods being overridden, although it's not obvious how that affects the equality of HttpRequest. For now I'd suggest creating an issue to investigate further and link to this code.

colossus/src/main/scala/colossus/protocols/http/HttpRequestParser.scala
```diff
@@ -11,18 +11,20 @@ object HttpRequestParser {
   
   def apply() = httpRequest
 
-  protected def httpRequest: Parser[HttpRequest] = httpHead |> {head =>
+  protected def httpRequest: Parser[HttpRequest] = httpHead |> { head =>
+    val contentType = head.headers.firstValue("content-type")
```
DanSimon: So this is not the right place to do this logic.  `headers.firstValue` ends up scanning the list of headers and is not efficient.  Instead, you want to look for this header during the parsing itself, which uses some specialized tools to do this as efficiently as possible.  See [HeadersBuilder](https://github.com/tumblr/colossus/blob/3a296309c97adf6bf878f69dc2c82c6d6dcfaa61/colossus/src/main/scala/colossus/protocols/http/HttpParse.scala#L28) and [ParsedHttpHeaders](https://github.com/tumblr/colossus/blob/3a296309c97adf6bf878f69dc2c82c6d6dcfaa61/colossus/src/main/scala/colossus/protocols/http/Header.scala#L184).  It should be pretty clear how to add the extra case for content-type, since it should work exactly the same as the other headers, but don't hesitate to ask if you have questions.

This applies to parsing both requests and responses.

colossus-tests/src/test/scala/colossus/protocols/http/HttpParseSpec.scala
DanSimon: Yeah sounds like an issue with equality checking.  `HttpRequest` is not as simple as it looks, since there's some kinda-crazy stuff happening internally for optimizations.

dxuhuang: `HttpBody` is equal if the `body: Array[Byte]` is equal:
https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/protocols/http/HttpBody.scala#L22

dxuhuang:
About the equality of `HttpRequest`, `HttpBody` is equal if and only if the `body: Array[Byte]` is equal:
https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/protocols/http/HttpBody.scala#L21

I temporarily added `&& that.contentType == this.contentType` to the byte array comparison and was able to make the test fail as expected.

The question is whether this should be the case; `HttpBody` being equal if the byte arrays are the same but the content-types are (possibly) different.

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
```diff
@@ -5,7 +5,9 @@ import akka.util.ByteString
 import scala.util.Try
 import core.DataBlock
 
-class HttpBody(private val body: Array[Byte], val contentType : Option[HttpHeader] = None)  {
+class HttpBody(private val body: Array[Byte], contentTypeString: Option[String] = None) {
```
dxuhuang: Decided to be a little bold here and change the constructor, since we probably don't want to give the caller the freedom to pass in non-`ContentType` `HttpHeader`s.

colossus-tests/src/test/scala/colossus/protocols/http/HttpParseSpec.scala
```diff
@@ -76,21 +76,24 @@ class HttpParserSuite extends WordSpec with MustMatchers{
...
         url = "/hello/world",
         version = HttpVersion.`1.1`,
         headers = List(
           "host" -> "api.foo.bar:444",
           "accept" ->  "*/*",
+          "content-type" -> "text/plain",
           "content-length" -> len.toString
         )
-      ), HttpBody(body))
+      ), HttpBody(body, "text/plain"))
```
dxuhuang: For reference, here is where the `HttpRequest` / `HttpBody` equality checking was supposed to fail

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
```diff
@@ -27,9 +29,7 @@ class HttpBody(private val body: Array[Byte], val contentType : Option[HttpHeade
 
   override def toString = bytes.utf8String
 
-  def withContentType(contentType: String) = withContentTypeHeader(HttpHeader("Content-Type", contentType))
-
-  def withContentTypeHeader(header: HttpHeader) = new HttpBody(body, Some(header))
```
dxuhuang: Nowhere did I see `withContentTypeHeader()` getting called so I removed it.

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
```diff
@@ -89,16 +88,19 @@ trait HttpBodyEncoders {
...
+  val NoBody = new HttpBody(Array.emptyByteArray, Option.empty[String])
 
   def apply[T](data: T)(implicit encoder: HttpBodyEncoder[T]): HttpBody = encoder.encode(data)
 
-  def apply[T](data: T, contentType: String)(implicit encoder: HttpBodyEncoder[T]) = {
+  def apply[T](data: T, contentType: String)(implicit encoder: HttpBodyEncoder[T]): HttpBody = {
     encoder.encode(data).withContentType(contentType)
   }
 
-  def apply[T](data: T, contentTypeHeader: HttpHeader)(implicit encoder: HttpBodyEncoder[T]) = {
```
dxuhuang: Nowhere did I see this overload being called so I removed it

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
```diff
@@ -15,13 +15,14 @@ object HttpResponseParser  {
...
+  protected def staticBody(dechunk: Boolean): Parser[HttpResponse] = head |> { parsedHead =>
+    val contentType = parsedHead.headers.contentType
     parsedHead.headers.transferEncoding match {
       case TransferEncoding.Identity => parsedHead.headers.contentLength match {
         case Some(0)  => const(HttpResponse(parsedHead, HttpBody.NoBody))
-        case Some(n)  => bytes(n) >> {body => HttpResponse(parsedHead, new HttpBody(body))}
+        case Some(n)  => bytes(n) >> { body => HttpResponse(parsedHead, new HttpBody(body, contentType)) }
         case None if (parsedHead.code.isInstanceOf[NoBodyCode]) => const(HttpResponse(parsedHead, HttpBody.NoBody))
-        case None     => bytesUntilEOS >> {body => HttpResponse(parsedHead, body)}
+        case None     => bytesUntilEOS >> { body => HttpResponse(parsedHead, body) }
```
dxuhuang: TODO: pass `contentType` into `HttpResponse.apply`

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
```diff
@@ -15,13 +15,14 @@ object HttpResponseParser  {
...
       }
       case _  => chunkedBody >> {body => HttpResponse(parsedHead, body)}
```
dxuhuang: TODO: pass `contentType` into `HttpResponse.apply`

colossus-tests/src/test/scala/colossus/protocols/http/HttpParseSpec.scala
DanSimon: Yes I think content-type should be included in the equality check.

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
DanSimon: We should not be making API changes in minor releases as it will break binary compatibility.  If we do this (which seems pretty reasonable to me), it'll have to go in the 0.10 release.

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
DanSimon: Again, we shouldn't be making API changes in a minor release.  Also in general I don't think these should be removed.  Just because nothing _inside_ the framework uses this method doesn't necessarily mean that nobody _using_ the framework is using it.

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
DanSimon: Same as my other comments.

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
DanSimon: Actually, `bytesUntilEOS` should be returning an `Array[Byte]`, not a `ByteString`, in which case you should be able to create a `HttpBody` the same way you do in the other cases.  I'll create an issue for this, but def out of scope for this PR.

colossus/src/main/scala/colossus/protocols/http/HttpResponseParser.scala
DanSimon: Actually this might be a good place to use `HttpBody.withContentType`

dxuhuang:
bump

PR #535, opened 2017-08-09
Generic clients

benblack86:
Added back information on generic clients. I'm trying to learn more about higher kinded types so this now makes sense! Whilst I was there, I changed a few `flatMap`s to` map`s since nothing needed to be flatted.

### DIFF
--- a/colossus-docs/src/main/scala/GenericClient.scala
+++ b/colossus-docs/src/main/scala/GenericClient.scala
@@ -1,0 +1,50 @@
+import akka.actor.ActorSystem
+import colossus.IOSystem
+import colossus.protocols.http.HttpMethod.Get
+import colossus.protocols.http.UrlParsing.{Root, on}
+import colossus.protocols.http.server.{HttpServer, Initializer, RequestHandler}
+import colossus.protocols.http.{Http, HttpClient, HttpRequest}
+import colossus.service.GenRequestHandler.PartialHandler
+import colossus.service.{Async, CallbackAsync, FutureAsync}
+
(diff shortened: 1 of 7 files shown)


colossus-docs/src/main/scala/GenericClient.scala
```diff
@@ -0,0 +1,50 @@
...
+  implicit val ioSystem = IOSystem()
+
+  def doTheThing[A[_]](client: HttpClient[A])(implicit async: Async[A]): A[String] = {
+    async.map(client.send(HttpRequest.get("/"))) { response =>
+      response.body.bytes.utf8String
+    }
+  }
+
+  // using future client
+  implicit val futureAsync = new FutureAsync()
```
DanSimon: Hmm, you shouldn't have to be creating your own `FutureAsync`.  Well, at least that was my intention.  This stuff is currently used in [ClientFactory](https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/service/ServiceDSL.scala#L226), and notice that an `AsyncBuilder` is used.  

This is probably out of scope for this PR, but it should ideally be the case that all you need to do is import a sub-package or object containing some implicits, and let the builder do the work for you.

benblack86: I saw the AsyncBuilder, but I couldn't see how to use it and there was no obvious implicits I should be importing.

benblack86:
Added `codecov.yml`

benblack86:
Also bumping SBT version since the previous version was causing problems with IntelliJ.

PR #536, opened 2017-08-09
Refactor

benblack86:
Hopefully a minor PR.

+ I noticed that some protocol types used the letter `C`, whilst other places used `P`. I refactor so `P` is consistently used.
+ Removed some unused imports
+ Made the `lift` implementations more explicit


amotamed:
👍

PR #537, opened 2017-08-10
Remove implicits

benblack86:
I was trying to understand how the client setup logic works and in particular how the `CodecClientFactory` was creating the client. However, I was getting stuck understanding how it was all glued together due to `baseFactory`, `lifter`, and `builder` being implicits. I was like, where are these implicits coming from!?! After a while I realized that some of them there being defined in different files, but under the same namespace so where in scope.

To bring more clarity to the code base, I've turning the implicits for `CodecClientFactory` and `ClientFactories` into explicit parameters. Thumbs up, thumbs down?

I then went crazy and moved the lifter implementations to where they are used and cut out some default objects that weren't being used. In the case of http, I moved the error routing logic to where it is being used. I'm guessing this breaks the rule of no API changes so I should leave those for 0.10?

Thoughts?

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/package.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/package.scala
@@ -5,6 +5,7 @@ import colossus.metrics.TagMap
 
 import service._
 
+import scala.language.higherKinds
 
 package object http extends HttpBodyEncoders with HttpBodyDecoders {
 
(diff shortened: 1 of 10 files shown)


DanSimon:
Yeah most of that stuff isn't needed anymore now that users cannot directly create `ServiceClient` and `FutureClient`.  I will examine this more in depth later today, but at first glance it seems fine.  I want to make sure we have some tests around creating clients, including load balancers, but I'm pretty sure everything that I expect to work will still work.

As for API changes, yes this looks like it will break binary compatibility and should go in 0.10.  On that note we should probably spend some time to actually scope out what we want to go in 0.10 (I don't think it should be too much, want to start aiming for faster releases).

colossus/src/main/scala/colossus/protocols/redis/package.scala
```diff
@@ -12,22 +13,15 @@ package object redis {
...
 
-      implicit val redisServerDefaults = new ServiceCodecProvider[Redis] {
-        def provideCodec() = new RedisServerCodec
-        def errorResponse(error: ProcessingFailure[Command]) = ErrorReply(s"Error (${error.reason.getClass.getName}): ${error.reason.getMessage}")
-      }
-      */
+  object Redis extends ClientFactories[Redis, RedisClient](RedisClientLifter) {
 
-    }
+    implicit def clientFactory = ServiceClientFactory.basic("redis", () => new RedisClientCodec)
```
aliyakamercan: should this be a `lazy val` ?

benblack86: should it? what are you thinking?

aliyakamercan: Since implicits are gone, I think we can just use `val`

colossus/src/main/scala/colossus/protocols/http/package.scala
```diff
@@ -25,33 +26,16 @@ package object http extends HttpBodyEncoders with HttpBodyDecoders {
...
     }
+  }
 
-    object defaults  {
-      
-      implicit val httpServerDefaults = new ServerDefaults
+  object Http extends ClientFactories[Http, HttpClient](HttpClientLifter) {
 
-      implicit val httpClientDefaults = new ClientDefaults
+    implicit lazy val clientFactory = ServiceClientFactory.basic("http", () => new StaticHttpClientCodec)
```
aliyakamercan: same here as well

benblack86:
Going to split up and put into 0.10

PR #538, opened 2017-08-10
Benchmarking

benblack86:
+ Added instructions Dan sent me on how to do benchmarking along with the lua script that enables pipelining.
+ Updated to latest sbt-revolver (don't need to explicitly add to `Build.scala`)
+ Removed `scala.version` from `build.properties` since I'm pretty sure it isn't used.
+ Removed commented out code from colossus-example main apps, which I seem to have broken when I merge the docs PR 😱


DanSimon:
👍

Issue #531, opened 2017-08-03
Remove akka agent from MetricClock/MetricSystem


benblack86:

benblack86:
See https://github.com/tumblr/colossus/issues/529 for dan's approach

PR #539, opened 2017-08-10
Replace Akka Agent with AtomicReference in MetricClock and MetricSystem

dxuhuang:

https://github.com/tumblr/colossus/issues/531


benblack86:
Looks good. I think this is the last place where akka-agent is used so you can remove the library from Build.scala.

jlbelmonte:
👍

Issue #449, opened 2016-09-19
Add a way to intialize collectors for specific tagmaps


DanSimon:
Normally when a metric is created, it will not report anything until values are added to it.  We can't simply start reporting 0 since we don't know what tag key/values will be used, but in many cases we use metrics without any tags or with a known set of fixed tags.  In those cases it would be nice to be able to force collectors to start immediately reporting 0, so we don't have gaps in the resulting time-series data when services are restarted.

I see two possible ways to address this:
1.  Give every collector some kind of "touch" or "reset" function that accepts a `TagMap`.  This would essentially create a 0 value for the given tagmap.
2.  Add an `initialZeros` configuration setting that accepts a sequence of `TagMap`s and basically does the 1. itself.

Or we could do both.

benblack86:
I'd vote to close this until we have a better use case. If we have a metric that tends to be zero, then we use a counter and set it to zero on service startup. It is occasionally annoying when a new service doesn't report /errors, but it usually resolves itself :)

cc @aliyakamercan

PR #540, opened 2017-08-14
Initialize metrics collection map

dxuhuang:
Assign 0 to the empty `TagMap`. Wish I could do this in the `CollectionMap` constructor, but it only accepts params of type `T`.

https://github.com/tumblr/colossus/issues/449

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/Collection.scala
@@ -65,18 +65,18 @@ private[metrics] class CollectionMap[T] {
     Option(map.get(tags)) match {
       case Some(got) => op(got)(num)
       case None => {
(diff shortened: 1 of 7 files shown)


DanSimon:
@benblack86 I mostly agree with your comments.  I'm fairly certain the issue was created after a discussion about the errors metric and the fact that it can be confusing when sometimes the metric shows up as 0 vs not at all.

The only thing to keep in mind is `service/errors` is a built-in Colossus metric.  So in that case "initialize to zero on startup" would have to be something we hardcode into colossus itself.  Maybe that's what we want, but we have to be really careful that it's a viable solution for all use cases.

Also since it seems like we're not fully settled on what to do, maybe we should move discussion into the issue itself and make a final decision before we make any more changes here.

aliyakamercan:
I agree with @benblack86 on #449. I don't think this is an issue and we don't have a solution which covers all use cases.

PR #549, opened 2017-08-15
Pass ServiceConfig's maxRequestSize into HttpRequestParser

dxuhuang:

https://github.com/tumblr/colossus/issues/542

### DIFF
--- a/colossus/src/main/resources/reference.conf
+++ b/colossus/src/main/resources/reference.conf
@@ -48,7 +48,7 @@ colossus{
       request-buffer-size : 100
       log-errors : true
       request-metrics : true
-      max-request-size : "1000 MB"
+      max-request-size : "10 MB"
     }
   }
   
--- a/colossus/src/main/resources/reference.conf
+++ b/colossus/src/main/resources/reference.conf
@@ -48,7 +48,7 @@ colossus{
       request-buffer-size : 100
       log-errors : true
       request-metrics : true
-      max-request-size : "1 MB"
+      max-request-size : "1000 MB"
     }
   }
   
--- a/colossus-docs/src/main/paradox/services.md
+++ b/colossus-docs/src/main/paradox/services.md
@@ -34,3 +34,8 @@ On the incoming request, body and content type are on the `HttpBody` and headers
 `HttpHead`.
 
 @@snip [HttpService2.scala](../scala/HttpService2.scala) { #example3 }
+
+Services can be configured by specifying values in `application.conf` for the `ServiceConfig` object.
+The values are `request-timeout`, `request-buffer-size`, `log-errors`, `request-metrics`, and `max-request-size`.
+
+@@snip [application.conf](../resources/application.conf) { #example4 }
--- a/colossus-docs/src/main/resources/application.conf
+++ b/colossus-docs/src/main/resources/application.conf
@@ -1,0 +1,6 @@
+// #example4
+colossus.service.example-server {
+  max-request-size = "20 MB"
+  request-timeout = "5 seconds"
+}
+// #example4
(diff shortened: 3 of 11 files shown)


dxuhuang:
Wondering if the `maxInitBufferSize` parameter to `bytes()` should also come from `ServiceConfig.requestBufferSize`.

dxuhuang:
`ServiceConfig.maxRequestSize` is passed into `ServiceServer.controllerConfig.inputMaxSize`. Should we change `reference.conf` accordingly?

benblack86:
My initial thoughts are yes we should add to `reference.conf` and no to `maxInitBufferSize` since I think we should leave that until people ask for it (hopefully never?!). I think passing `ServiceConfig` is fine since it is passed by reference so there should be no performance difference between passing an `int` and a `ServiceConfig` object (unless cpu memory is affected?).

DanSimon:
So `ServiceConfig.requestBufferSize` is not the size of a byte buffer, but rather the maximum number of pending requests for a single connection before the connection starts automatically failing requests.  The Http parsers won't need that value.

But `maxInputBufferSize` in `bytes()` and `inputMaxSize` in `ControllerConfig` should both be set to `ServiceConfig.maxRequestSize`.  

So you should not have to pass in the entire `ServiceConfig` to the parsers.  I don't think that should be done anyway since the parser doesn't need to know about any of the other settings and we should always aim to keep things as focused as possible.  If in the future a parser needs multiple values from the config, we would group those into another case class inside of `ServiceConfig`.

dxuhuang:
@DanSimon In `Parsing.scala`:
```
def bytes(num: Int, maxSize: DataSize, maxInitBufferSize: DataSize): Parser[Array[Byte]] = { ... }
```
Which parameter did you mean by `maxInputBufferSize`? If it's `maxInitBufferSize`, I recall you said a while ago that this should be small, which is why it's currently hardcoded to 1 KB (https://github.com/tumblr/colossus/pull/505).

colossus/src/main/resources/reference.conf
```diff
@@ -48,7 +48,7 @@ colossus{
       request-buffer-size : 100
       log-errors : true
       request-metrics : true
-      max-request-size : "1 MB"
+      max-request-size : "1000 MB"
```
dxuhuang: Correct limit?

DanSimon:
Ah yeah I misread, I meant `maxSize` when talking about coming from config.  You're correct that `maxInitBufferSize` should be small and can stay hardcoded at 1KB.  It could potentially be an issue if a service is constantly parsing large (multi-MB) requests, but would need some experimentation to be really sure before we make it configurable.

dxuhuang:
bump

colossus/src/main/resources/reference.conf
DanSimon: hmm, I'd like to hear other opinions on this.  Personally I think it's probably better to keep this small but make it clear how to increase it.  Not a big deal since I think you should always tune this for each situation, but if I had to choose between accidentally rejecting a bunch of requests because this was defaulted to too low vs crashing a server because it ran out of memory trying to accept lots of large requests, I'd choose the former.

1MB was probably too small though, maybe 10MB?

dxuhuang: @benblack86 @aliyakamercan @jlbelmonte @amotamed

benblack86: 10.MB sounds good to me. Can you update the services docs to explain how to configure a service please. Also your branch needs to be compared to develop branch and not master. All development happens on develop now.

dxuhuang:
bump, added docs

Issue #372, opened 2016-04-19
Upgrade Akka


DanSimon:
Right now we're on Akka 2.3.9, we should upgrade to the latest, probably 2.4.3

hntd187:
It should be important to note akka 2.4.X is Scala 2.11+ and Java8+

DanSimon:
Ah that's a good point.  I suspect then this will get pushed off a bit.  How soon probably depends if we can easily do builds for both 2.3.x and 2.4.x

The good news is I did the upgrade myself just for fun and everything seemed to work fine.  The same goes for using Scala 2.12, with the exception that there appears to be no artifact of akka-agent published right now.  So we'll likely do both at the same time, probably in some kind of experimental branch.

hntd187:
I believe as long as your language dependencies are there it should be fine, the migration guide says it's gucci

http://doc.akka.io/docs/akka/2.4.1/project/migration-guide-2.3.x-2.4.x.html

hntd187:
Hey @DanSimon just thought I'd let you know, in some internal stuff I am working on I evicted akka 2.3 in favor of 2.4.9 and it's been running perfectly fine for awhile. So to test this further, I forked colossus and upgraded the dependency and other than some mild deprecation warnings all the tests passed for me. Just figured I'd let you know when you decide to upgrade or if you want to push another artifact the upgrade for it seems rather painless at face value.

mfirry:
(Slightly) related... Akka 2.11 has been ported to Scala 2.12. So a move over to akka 2.4.11/12 would also gain (possibly) a move to Scala 2.12.

PR #557, opened 2017-08-16
Drop 2.10.6 Support

aliyakamercan:
- removed `crossScalaVersions`
- removed scala version and version check in travis.yml
- added new lines in travis.yml

#554 

@benblack86 @DanSimon @jlbelmonte @dxuhuang

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,35 @@
 language: scala
+
 scala:
-- 2.10.6
 - 2.11.7
-jdk: oraclejdk7
+- 2.12.2
+
+jdk: oraclejdk8
+
 sudo: false
+
 services:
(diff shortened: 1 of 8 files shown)


benblack86:
I fine dropping 2.10, but don't we also want to support 2.12? Or does 2.11 not work with the latest akka?

aliyakamercan:
We need to update akka before that.

aliyakamercan:
@benblack86 updated akka and added 2.12. 
#548 #372

DanSimon:
LGTM 👍  

The last time I tried building colossus on 2.12 it was significantly (but not dramatically) slower in benchmarks.  I will be very interested to see how it does now.

PR #559, opened 2017-08-17
Document redis

benblack86:
Adding a redis example to the services page and providing an example of how to configure a service through code.


colossus-docs/src/main/scala/RedisServiceExample.scala
```diff
@@ -0,0 +1,46 @@
...
+  implicit val actorSystem = ActorSystem()
+  implicit val ioSystem = IOSystem()
+
+  val db = new ConcurrentHashMap[String, String]()
+
+  RedisServer.start("example-server", 9000) {
+    new Initializer(_) {
+      override def onConnect = new RequestHandler(_) {
+        override def handle: PartialHandler[Redis] = {
+          case Command("GET", args) =>
```
DanSimon: Just a comment, but I know at one point we used to have extractors for each command, which would have made this cleaner `case Get(key)`.  I think that was all removed when we created the `RedisClient` interface (which in general was a good move), but maybe we should think about bringing back the extractors.  I'd also not be opposed to a macro or code-generation tool for those since it is a pain to write/update stuff for each command in multiple places.

benblack86: You suggesting something similar to what we have for [memcache](https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala#L47)?

I didn't understand your point on macro/code-generation.

Issue #545, opened 2017-08-15
Setting content type via `withHeader` should work


benblack86:
```
req.ok(jsonized.toString()).withHeader(HttpHeaders.ContentType, ContentType.ApplicationJson)
```

Actually need to set on HttpBody for it to work.

PR #569, opened 2017-08-24
Remove Content Type from Http Body

dxuhuang:
@DanSimon  @benblack86  @aliyakamercan  @amotamed  @jlbelmonte 

Second stab at https://github.com/tumblr/colossus/issues/545

So what we'd like to be able to do is set the content-type on the `HttpResponse` without having to bother the internal `HttpBody`. I don't see an inherent reason why the content-type has to be part of `HttpBody` anyway; after all, the only substantial place it's used is in the `encode()` method that writes to the `DynamicOutBuffer`, and the body itself is just a byte array. If we add the content-type `Header` properly, `HttpResponseHead.encode()` will take care of it just the same.

The only issue I see is potential duplicate headers since `HttpHeaders` is just a wrapper around a linked list, but I think that should still be handled in `HttpHeaders` alone. In fact I'm thinking about going a bit further and making `HttpHeaders` wrap a `HashMap[String, HttpHeader]` instead.

Discuss.

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -260,7 +260,6 @@ class HttpHeaders(private[http] val headers: JList[HttpHeader]) {
     while (it.hasNext) {
       it.next.encode(buffer)
     }
-
   }
 
   /**
@@ -273,8 +272,10 @@ class HttpHeaders(private[http] val headers: JList[HttpHeader]) {
 
   override def equals(that: Any): Boolean = that match {
     case that: HttpHeaders => this.toSeq.toSet == that.toSeq.toSet
-    case other             => false
+    case _                 => false
   }
+
+  override def hashCode(): Int = this.toSeq.toSet.hashCode()
 
   override def toString = "[" + toSeq.map { _.toString }.mkString(" ") + "]"
 
--- a/colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
@@ -5,8 +5,7 @@ import colossus.core.{DataBlock, DataOutBuffer}
 
 import scala.util.Try
 
-//TODO: this constructor should take an Option[String] for the contentType and create the header internally
-class HttpBody(private val body: Array[Byte], val contentType: Option[HttpHeader] = None) {
+class HttpBody(private val body: Array[Byte]) {
 
   def size = body.length
 
@@ -20,19 +19,13 @@ class HttpBody(private val body: Array[Byte], val contentType: Option[HttpHeader
   def as[T](implicit decoder: HttpBodyDecoder[T]): Try[T] = decoder.decode(body)
 
   override def equals(that: Any) = that match {
-    case that: HttpBody => (that.bytes == this.bytes) && (that.contentType == this.contentType)
+    case that: HttpBody => that.bytes == this.bytes
     case _              => false
   }
 
   override def hashCode = body.hashCode
 
   override def toString = bytes.utf8String
-
-  def withContentType(contentType: String) = withContentTypeHeader(HttpHeader(HttpHeaders.ContentType, contentType))
-
-  // TODO: remove?
-  def withContentTypeHeader(header: HttpHeader) = new HttpBody(body, Some(header))
-
 }
 
 /**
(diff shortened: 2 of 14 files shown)


colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -86,9 +86,6 @@ case class HttpResponse(head: HttpResponseHead, body: HttpBody) extends Encoder
 
   def encode(buffer: DataOutBuffer, extraHeaders: HttpHeaders) {
     head.encode(buffer)
-    body.contentType.foreach { ctype =>
```
dxuhuang: removed from here

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
```diff
@@ -70,38 +63,31 @@ trait HttpBodyEncoder[T] {
...
 trait HttpBodyEncoders {
   implicit object ByteStringEncoder extends HttpBodyEncoder[ByteString] {
+    val contentType = None
     def encode(data: ByteString): HttpBody = new HttpBody(data.toArray)
   }
 
   implicit object StringEncoder extends HttpBodyEncoder[String] {
-    val contentTypeHeader              = HttpHeader(HttpHeaders.ContentType, "text/plain")
-    def encode(data: String): HttpBody = new HttpBody(data.getBytes("UTF-8"), Some(contentTypeHeader))
+    val contentType = Some(ContentType.TextPlain)
```
dxuhuang: Put these inside `Encoders`

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
```diff
@@ -70,38 +63,31 @@ trait HttpBodyEncoder[T] {
 
   def encode(data: T): HttpBody
 
-  def contentTypeHeader(contentType: String): HttpHeader = HttpHeader("Content-Type", contentType)
+  def contentType: Option[String]
```
dxuhuang: Put these inside Encoders

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -116,9 +115,12 @@ object HttpResponse extends HttpResponseBuilding {
   def initialVersion = HttpVersion.`1.1`
 
   def apply[T: HttpBodyEncoder](head: HttpResponseHead, body: T): HttpResponse = {
-    HttpResponse(head, HttpBody(implicitly[HttpBodyEncoder[T]].encode(body)))
+    val encoder = implicitly[HttpBodyEncoder[T]]
+    val response = HttpResponse(head, HttpBody(encoder.encode(body)))
+    encoder.contentType.fold(response) { ct => response.withContentType(ct) }
```
dxuhuang: pass these from `HttpBodyEncoder`s

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -273,9 +272,11 @@ class HttpHeaders(private[http] val headers: JList[HttpHeader]) {
 
   override def equals(that: Any): Boolean = that match {
     case that: HttpHeaders => this.toSeq.toSet == that.toSeq.toSet
-    case other             => false
+    case _                 => false
   }
 
+  override def hashCode(): Int = this.toSeq.toSet.hashCode()
```
benblack86: Nice, this is an example of the most common java interview question :)

colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -160,7 +160,6 @@ class ServiceClient[P <: Protocol](
     val context: Context
 )(implicit tagDecorator: TagDecorator[P] = TagDecorator.default[P])
     extends ControllerDownstream[Encoding.Client[P]]
-    with HasUpstream[ControllerUpstream[Encoding.Client[P]]]
```
benblack86: I take it this isn't needed?

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -116,9 +115,12 @@ object HttpResponse extends HttpResponseBuilding {
...
   }
 
+  // THESE ARE ONLY USED IN TESTS!
```
DanSimon: Not true, remember this is a framework, so just because a method isn't used by the framework itself doesn't mean a user of the framework isn't using it :)

colossus/src/main/scala/colossus/service/ServiceClient.scala
DanSimon: This seems ok since it's very small, but in the future I would like to avoid making changes that aren't relevant to the issue being addressed in the PR.

dxuhuang: https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/controller/Controller.scala#L43
`trait ControllerDownstream[E <: Encoding] extends HasUpstream[ControllerUpstream[E]]` so it's redundant

benblack86:
You are correct that you can set the content-type via `withHeader`, but that isn't very useful since you'll end up with two content-type headers when you actually want one (the one set `withHeader`). I know the problem has bitten me twice and at least one other person. I'm not sure what functionality we would lose/break if content type was removed from the body.

DanSimon:
Yeah that's definitely a good point.  I'm going to go ahead and approve this PR.  I'm still a little concerned that there's a better reason for keeping the header with the body that I'm just not thinking of right now, but since the PR still allows the `HttpBodyEncoder` to work properly, which was my main concern, I think it should be ok.

👍

dxuhuang:
bump @benblack86 updated docs

PR #571, opened 2017-08-28
Use a simple logger wrapper instead of Akka Logging

### LINKED ISSUE #409: Drop Akka logging in favor of slf4j
We should also create a wrapper using call-by-name for log messages.

aliyakamercan:
@benblack86 @DanSimon @dxuhuang 

fixes #409

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -38,7 +38,8 @@ lazy val GeneralSettings = Seq[Setting[_]](
     "org.scalatest"          %% "scalatest"                   % SCALATEST_VERSION % "test, it",
     "org.scalamock"          %% "scalamock-scalatest-support" % "3.6.0" % "test",
     "org.mockito"            % "mockito-all"                  % "1.9.5" % "test",
-    "com.github.nscala-time" %% "nscala-time"                 % "2.16.0"
+    "com.github.nscala-time" %% "nscala-time"                 % "2.16.0",
+    "org.slf4j"              % "slf4j-api"                    % "1.7.6"
   ),
   coverageExcludedPackages := "colossus\\.examples\\..*;.*\\.testkit\\.*"
 ) ++ Defaults.itSettings
@@ -85,7 +86,8 @@ lazy val MetricSettings = ColossusSettings
 
 lazy val ExamplesSettings = Seq(
   libraryDependencies ++= Seq(
-    "org.json4s" %% "json4s-jackson" % "3.5.3"
+    "org.json4s"     %% "json4s-jackson" % "3.5.3",
+    "ch.qos.logback" % "logback-classic" % "1.2.2"
   )
 )
 
--- a/colossus-metrics/src/main/scala/colossus/metrics/logging/ColossusLogging.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/logging/ColossusLogging.scala
@@ -1,0 +1,46 @@
+package colossus.metrics.logging
+
+import org.slf4j.{Logger, LoggerFactory}
+
+trait ColossusLogging {
+
+  private val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  def debug(s: => String): Unit = {
+    doLog(s, log.isDebugEnabled, log.debug)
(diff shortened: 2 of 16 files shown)


colossus-metrics/src/main/scala/colossus/metrics/logging/ColossusLogging.scala
```diff
@@ -0,0 +1,46 @@
+package colossus.metrics.logging
+
+import org.slf4j.{Logger, LoggerFactory}
+
+trait ColossusLogging {
```
DanSimon: Not sure we need to call it `ColossusLogging`, could just do `Logging`, but I'm ok either way.

build.sbt
```diff
@@ -85,7 +86,8 @@ lazy val MetricSettings = ColossusSettings
 
 lazy val ExamplesSettings = Seq(
   libraryDependencies ++= Seq(
-    "org.json4s" %% "json4s-jackson" % "3.5.3"
+    "org.json4s"     %% "json4s-jackson" % "3.5.3",
+    "ch.qos.logback" % "logback-classic" % "1.2.2"
```
DanSimon: I don't think we should be adding logback as a dependency.  We don't want to cause issues with anyone using a different logging backend.  

I realize that realistically most people probably don't care and just want logging to work out of the box.  Maybe we can create a separate `colossus-logback` artifact or something that just adds the dependency and XML file.

nsauro: +1.  I think this will cause conflicts to users if they are not using logback, or are using different version of logback.  If we stick to SLF4J, it allows colossus to provide logging output, but allows the users to configure how and when they see it.

colossus-metrics/src/main/scala/colossus/metrics/logging/ColossusLogging.scala
nsauro: FWIW: Not sure if anyone saw my previous comment, but there is already a thing that does this: 
https://github.com/typesafehub/scala-logging

build.sbt
benblack86: Isn't this just adding it to the examples project? Therefore, isn't part of the colossus build?

DanSimon: Ahhhh I missed that.  Well then everything looks fine to me. 💯

aliyakamercan: That's for examples project only, so it can log correctly. It's not pulled into colossus.

nsauro: duh. Sorry

colossus-metrics/src/main/scala/colossus/metrics/logging/ColossusLogging.scala
aliyakamercan: This is pretty simple, we don't really need to pull in another dependency.

benblack86: Also scala-logging does the magic using macros, which might cause issues with debugging if I read the bottom of the  scala-logging readme correctly.

nsauro: What kind of issues, do you think that will cause? It has a few other things in it too, such as lazy/non lazy implementations which is nice if you don't want to instantiate loggers which don't get used.

NBD either way, I figured we would avoid reinventing the wheel :)

benblack86: They have a section [Debugging Scala in Intellij](https://github.com/typesafehub/scala-logging#debugging-scala-in-intellij). I haven't tried it myself, so I don't know the details.

Luckily this wheel is pretty small, plus didn't you write this code? Can't get a better implementation than that!

nsauro: The code did look vaguely familiar, but I wasn't sure if it was me :-)

In either event it's not big deal as like you are saying, it's a small wheel.  But I felt like I should at least point out that the wheel does exist.

PR #581, opened 2017-09-01
Protect constant callback

benblack86:
I was playing around with callback, because I hear they are like the future 🥁 .

The docs say that `CallbackExecutionException` is thrown if there is an uncaught exception in the execution block of a Callback. This is true for unmapped and mapped callback, but not for constant callback. I added a try/catch!

@DanSimon @nsauro @dxuhuang @aliyakamercan


jlbelmonte:
👍

PR #583, opened 2017-09-01
adding mima

nsauro:
This is a no frills mima integration.

I wanted to do something a bit more fancy then just copy/pasting and changing the dep for the mima setting, but wasn't sure if it was super necessary.  I was going to use an implicit class on project that exposed a new function `withMima(...)` which would allow each project to opt into mima, and that function could generate the setting.  This is probably good for now, but if we want to start tracking more than just one pervious version, we might want that function at some point.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -99,6 +99,9 @@ lazy val RootProject = Project(id = "root", base = file("."))
 
 lazy val ColossusProject: Project = Project(id = "colossus", base = file("colossus"))
   .settings(ColossusSettings: _*)
+  .settings{
+    mimaPreviousArtifacts := Set("com.tumblr" % "colossus_2.11" % "0.9.1")
+  }
   .configs(IntegrationTest)
   .aggregate(ColossusTestsProject)
   .dependsOn(ColossusMetricsProject)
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -99,6 +99,9 @@ lazy val RootProject = Project(id = "root", base = file("."))
 
 lazy val ColossusProject: Project = Project(id = "colossus", base = file("colossus"))
   .settings(ColossusSettings: _*)
+  .settings{
+    mimaPreviousArtifacts := Set("com.tumblr" % "colossus_2.11" % "0.9.1")
```
DanSimon: maybe put the version in a constant so we only have to update it in one place next time.

nsauro:
I noticed that we have cross publishing for only 2.11 and 2.12, we dropped 2.10 now, right? It looks like we haven't yet published anything for 2.12, is that right?  If I try to run mima using 2.12 artifacts, I get failures.  I can, for now, change this a bit so that it only generates the setting for 2.11, and when we publish 2.12 artifacts, remove that restriction.


Something along the lines of:

```scala
def withMima(versions : String*) : Project = {
      project.settings{
        mimaPreviousArtifacts := (if(scalaBinaryVersion.value == "2.11") {
          versions.map{ v =>
            organization.value %% s"${moduleName.value}" % v
          }.toSet
        }else{
          Set.empty
        })
      }
    }
```

After we publish 2.12, we can then remove that if statement.

nsauro:
So, let me clarify my previous comment :).

Currently, the scala version of Colossus is set to 2.12, and we cross publish to 2.11.  Since we have not published anything yet(from what I can tell), mima fails since it cannot find the 2.12 artifacts.  However, if I run `+mimaReportBinaryIssues`, it will run that for all scala versions listed, and work for 2.11.  I just need to disable it for 2.12 until we publish artifacts

Issue #414, opened 2016-06-22
Make error reporting more configurable


DanSimon:
This is spawning from the fact that right now for exceptions that are built into colossus, such as `DroppedReplyException`, we currently print out the entire stack trace which is pretty much useless in those situations.  

We already have a `RequestFormatter` that is used to format requests when being output in an error message, but I think it makes more sense to allow even more flexibility by providing a way to completely control if/how errors are reported.

onesupercoder:
Hey @DanSimon this project is amazing, thank you for making it!  👍   I was wondering if the error reporting configuration is solved in the develop Branch with the new GenRequestHandler you made which provides an unhandledError function to capture errors.  I am new to Scala and am not entirely sure what I am saying.  Could i possibly help by testing the develop branch or some other version or adding tests somewhere?

DanSimon:
Hey @onesupercoder , you're correct that this is somewhat solved in the develop branch, although it's not quite as complete as I would like.  You are able to override the `unhandledError`function, but that only includes errors that occur while processing a request.  Other errors, including `DroppedReplyException`, can occur after request processing so they won't be caught there.  

I'm thinking more along the lines of creating a few mixin traits for various exception types that can instruct the logger on how much information to dump.  An exception like `DroppedReplyException` doesn't benefit from printing a stack trace (since it's always happening in the exact same spot), whereas it would be more useful for some exception thrown by your own code.

This probably won't make it into 0.9.0 since we're already in the release candidate phase and trying to avoid code changes that aren't direct bug fixes, but we'll try to work it into 0.9.1 which shouldn't be far behind.

onesupercoder:
Makes sense, thank you so much @DanSimon!

PR #589, opened 2017-09-13
Add method to format error

dxuhuang:
@benblack86 @DanSimon @aliyakamercan @jlbelmonte @amotamed 

My stab at https://github.com/tumblr/colossus/issues/414

### DIFF
--- a/colossus-docs/src/main/paradox/services.md
+++ b/colossus-docs/src/main/paradox/services.md
@@ -72,7 +72,7 @@ colossus {
       request-metrics : true
       max-request-size : "1000 MB"
       errors : {
-        do-not-log : ["NonsenseException"]
+        do-not-log : []
         log-only-name : ["DroppedReplyException"]
       }
     }
@@ -86,12 +86,8 @@ To configure via code, create a `ServiceConfig` object and pass it to the `Reque
 
 @@snip [ServiceConfigExample.conf](../scala/ServiceConfigExample.scala) { #example1 }
 
-`RequestHandler` allows you to configure how request errors are reported. The `ServiceConfig` case class contains the
-`logErrors` and `errorConfig` fields, populated from the `log-errors` and `errors` config fields respectively.
-The trait `RequestFormatter` contains methods stubs that determine format requests and errors (together if you so desire),
-and its implementation class `ConfigurableRequestFormatter` uses the `ErrorConfig` case class from the `ServiceConfig` to
-decide whether to format the specific error. This method, `logWithStackTrace()` returns a `Option[Boolean]` (chosen for
-the convenience of `map` and `foreach`). You can create your own subclass of `ConfigurableRequestFormatter` that overrides
-the `format()` method.
+`RequestHandler` allows for the configuration of how request errors are reported. By default, requests are directly
+converted to `String`s and logged with the complete stack trace. This can be overridden by either filling in the `errors`
+config, or by providing a custom implementation of the `RequestFormatter` trait in the `RequestHandler`.
 
 @@snip [ServiceConfigExample.conf](../scala/ServiceConfigExample.scala) { #example2 }
--- a/colossus-docs/src/main/paradox/services.md
+++ b/colossus-docs/src/main/paradox/services.md
@@ -87,7 +87,7 @@ To configure via code, create a `ServiceConfig` object and pass it to the `Reque
 @@snip [ServiceConfigExample.conf](../scala/ServiceConfigExample.scala) { #example1 }
 
 `RequestHandler` allows for the configuration of how request errors are reported. By default, requests are directly
-converted to `String`s and logged with the complete stack trace. This can be overridden by either filling in the `errors`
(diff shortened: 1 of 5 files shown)


colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -87,7 +87,8 @@ object ServiceConfig {
 }
 
 trait RequestFormatter[I] {
-  def format(request: I): String
+  def formatRequest(request: I): String
+  def formatError(error: Throwable): String
 }
```
dxuhuang: An alternative to this would be an `ErrorFormatter` trait that contains its own method for formatting `Throwable`s separate from the request formatter.

benblack86:
Just trying to think how it would be overriden:

```
def formatError(error: Throwable): String = {
  error match {
    case DroppedReplyException => "DroppedReplyException"
    case _ => // get stack trace
  }
}
```

The problem is that it is a bit of work to extract the stack trace since that functionality is contained within the logger. Maybe it would be more useful to be able to indicate whether an error should be logged. There could be three levels:

+ don't log
+ log name only
+ log name and stack trace

It would be great if this stuff can be configurable (like the metrics) so by default the `DroppedReplyException` would only log name, but everything else would be name and stack trace.

DanSimon:
I like the general idea.  What do you guys think about having `formatError` also take in the failed request as a parameter, and then using the returned string as the entire error message?  That way the formatter has complete control over the error message.

I also agree with what @benblack86 said.  Having an ADT like:

```scala
sealed trait LogAction
case object DoNotLog extends LogAction
case class Log(message: String, includeStackTrace: Boolean) extends LogAction
```

or something similar seems like a good idea.  Also #414 does mention doing something for exceptions like `DroppedReplyException`.  I'm ok if that's not addressed in this PR, but we should then create another issue for it.

dxuhuang:
bump. will add stuff in conf soon

benblack86:
Looks good so far.

dxuhuang:
bump @DanSimon

colossus/src/main/scala/colossus/service/ServiceServer.scala
```diff
@@ -82,12 +84,48 @@ object ServiceConfig {
...
+case class ErrorConfig(doNotLog: Set[String], logOnlyName: Set[String])
+
+object RequestFormatter {
+  case class LogMessage(message: String, includeStackTrace: Boolean)
+}
+
 trait RequestFormatter[I] {
-  def format(request: I): String
+  import RequestFormatter._
+  // None means do not log. Some(false) means log only the name. Some(true) means log with stack trace (default).
```
DanSimon: Let's format this comment with proper scaladoc formatting so it shows up in the docs.

Personally I'd be more in favor of using a custom 3-state ADT (eg `case object DoNotLog`) which would be more descriptive than a `Option[Boolean]`, but it's not a big deal since it's only for this one method.

benblack86:
Might be wrong, but I don't think the `RequestFormatter` is documented. Can you add/update the docs.

dxuhuang:
bump

colossus-docs/src/main/scala/ServiceConfigExample.scala
```diff
@@ -32,6 +33,13 @@ object ServiceConfigExample {
         override def handle: PartialHandler[Http] = {
           case request @ Get on Root => Callback.successful(request.ok("Hello world!"))
         }
+
+        // #example2
+        override def requestLogFormat: Option[RequestFormatter[Request]] = {
+          // you can create a custom subclass of `ConfigurableRequestFormatter` and override the format() method here
+          super.requestLogFormat
```
benblack86: Can you put an actual example of a request formatter.

colossus-docs/src/main/paradox/services.md
```diff
@@ -71,6 +71,10 @@ colossus {
       log-errors : true
       request-metrics : true
       max-request-size : "1000 MB"
+      errors : {
+        do-not-log : ["NonsenseException"]
```
benblack86: You can leave this as empty since `NonsenseException` is not a real exception.

colossus-docs/src/main/paradox/services.md
```diff
@@ -81,3 +85,13 @@ To configure via code, create a `ServiceConfig` object and pass it to the `Reque
 @@snip [ServiceConfigExample.conf](../scala/ServiceConfigExample.scala) { #example }
 
 @@snip [ServiceConfigExample.conf](../scala/ServiceConfigExample.scala) { #example1 }
+
+`RequestHandler` allows you to configure how request errors are reported. The `ServiceConfig` case class contains the
```
benblack86: This description is very implementation specific. For a user of colossus, they just need to know that they can suppress errors by adding the config, or override the whole thing by providing their own implementation of `RequestFormatter` in code.

dxuhuang:
bump

PR #591, opened 2017-09-15
Streaming docs

### LINKED ISSUE #553: Document http streaming
Cause I don't know how to use it

DanSimon:
Fixes #553 

First attempt at more thorough docs for the streaming API.  This focuses more on the fundamentals of pipes, sources, and sinks, since I think once you understand those using the StreamingHttp protocol is more straightforward.  I think there is still more to add but we can improve it incrementally.  I'd also be interested to get people's opinions on what else should be covered.

I tried adding some links to the public scaladocs hosted by `javadoc.io`.  Unfortunately it doesn't work with paradox's built-in `scaladoc` tag, but it seems to work nicely using an `extref` tag.

Writing up these docs uncovered some areas in the API that could be cleaned up and improved a bit, so I'll have another PR soon that'll address some of it.

### DIFF
--- a/colossus-docs/src/main/scala/PipeExamples.scala
+++ b/colossus-docs/src/main/scala/PipeExamples.scala
@@ -58,4 +58,19 @@ object PipeExamples extends App {
     
   }
 
+  def pipeTransform = {
+
+    // #pipe_compose
+    val pipe1: Pipe[Int, Int] = new BufferedPipe[Int](10)
+    val pipe2: Pipe[String, String] = new BufferedPipe[String](10)
+
+    val pipe3: Pipe[Int, String] = pipe1.map{i => (i * 2).toString}.weld(pipe2)
+
+    pipe3.push(2)
+    pipe3.pull() // PullResult.Item("4")
+    // #pipe_compose
+  }
+
+    
+
 }
--- a/colossus-docs/src/main/scala/PipeExamples.scala
+++ b/colossus-docs/src/main/scala/PipeExamples.scala
@@ -71,6 +71,4 @@ object PipeExamples extends App {
     // #pipe_compose
   }
 
-    
-
 }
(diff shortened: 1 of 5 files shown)


colossus-docs/src/main/scala/StreamingHttpExample.scala
```diff
@@ -0,0 +1,40 @@
...
+import protocols.http.streaming._
+import core.DataBlock
+import service._
+import colossus.streaming._
+
+object StreamServiceExample {
+
+  // #streaming_http
+
+  class MyRequestHandler extends GenRequestHandler[StreamingHttp](serverContext, config) {
```
benblack86: Compile problem ‼️

colossus-docs/src/main/scala/PipeExamples.scala
```diff
@@ -0,0 +1,76 @@
...
+  def fullPush = {
+    // #full_push
+    //create a pipe with a buffer size of 1
+    val pipe = new BufferedPipe[Int](1)
+
+    pipe.push(10) //PushResult.ok
+
+    //the pipe can only buffer one item, so the next push fails and returns a
+    //PushResult.Full
+    val fullResult = pipe.push(12).asInstanceOf[PushResult.Full]
```
benblack86: Using a match would be nicer and less frowned upon :)

colossus-docs/src/main/scala/PipeExamples.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    // #full_push  
+  }
+
+  def emptyPull = {
+    // #empty_pull
+    val pipe = new BufferedPipe[Int](10)
+
+    //the pipe is empty so it returns a PullResult.Empty
+    val result = pipe.pull().asInstanceOf[PullResult.Empty]
```
benblack86: Ditto

PR #592, opened 2017-09-15
More details on callbacks

benblack86:
Adding more details about the different callback implementations.

@DanSimon @aliyakamercan @dxuhuang @nsauro @jlbelmonte @amotamed


aliyakamercan:
👍

Issue #580, opened 2017-09-01
Fix random test failure


benblack86:
Seen this test randomly fail a few times now:

![screen shot 2017-09-01 at 10 27 57](https://user-images.githubusercontent.com/315836/29974239-46d94e78-8f00-11e7-9342-f4e094072cb0.png)

benblack86:
Also this one:

![screen shot 2017-09-15 at 09 42 07](https://user-images.githubusercontent.com/315836/30485047-35dc2c44-99fa-11e7-8090-e3fff5ee67ca.png)

I'm guessing we just need to increase some timeouts.

PR #597, opened 2017-09-21
Increase ColossusSpec Timeout

dxuhuang:
@benblack86 @DanSimon @aliyakamercan @jlbelmonte @amotamed 

https://github.com/tumblr/colossus/issues/580

### DIFF
--- a/colossus-tests/src/test/scala/colossus/core/WorkerManagerSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/core/WorkerManagerSpec.scala
@@ -10,8 +10,10 @@ import org.scalatest.concurrent.Eventually
 
 import scala.collection.immutable.IndexedSeq
 import scala.concurrent.duration._
+import org.scalatest.time.Span._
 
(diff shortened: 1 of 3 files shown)


colossus-tests/src/test/scala/colossus/core/WorkerManagerSpec.scala
```diff
@@ -10,8 +10,10 @@ import org.scalatest.concurrent.Eventually
 
 import scala.collection.immutable.IndexedSeq
 import scala.concurrent.duration._
+import org.scalatest.time.Span._
 
 class WorkerManagerSpec extends ColossusSpec with Eventually {
+  override implicit def patienceConfig = super.patienceConfig.copy(timeout = 1.second)
```
benblack86: What's the default value?

dxuhuang: 150 milliseconds
```
final case class PatienceConfig(timeout: Span = scaled(Span(150, Millis)), interval: Span = scaled(Span(15, Millis)))
```

PR #601, opened 2017-09-25
add scala version to publish command

aliyakamercan:
@benblack86 @jlbelmonte @amotamed


amotamed:
LGTM

publish.sh
```diff
@@ -10,7 +10,7 @@ if [[ $TRAVIS_BRANCH =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-RC[0-9]+|-M[0-9]+)?$ || ($TRAV
     openssl aes-256-cbc -K $encrypted_9df8dc1a95a3_key -iv $encrypted_9df8dc1a95a3_iv -in secrets.tar.enc -out secrets.tar -d
     tar xvf secrets.tar
     # publish, close & release
-    sbt publishSigned sonatypeReleaseAll
+    sbt ++$TRAVIS_SCALA_VERSION publishSigned sonatypeReleaseAll
```
benblack86: What is this doing? And what value is `TRAVIS_SCALA_VERSION `?

aliyakamercan:
```
scala:
- 2.11.7
- 2.12.2
```
one of these, it is setting the correct scala version. Without this both build only publish 2.12 because that's what's in `build.sbt`.

PR #606, opened 2017-10-02
correct date format

dxuhuang:
@benblack86 @DanSimon @jlbelmonte @aliyakamercan @amotamed

### DIFF
--- a/colossus-tests/src/test/scala/colossus/protocols/http/HttpHeadSpec.scala
+++ b/colossus-tests/src/test/scala/colossus/protocols/http/HttpHeadSpec.scala
@@ -2,7 +2,6 @@ package colossus.protocols.http
 
 import org.scalatest._
 import java.util.Date
-import java.text.SimpleDateFormat
 import scala.util.{Success, Failure}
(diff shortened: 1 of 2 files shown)


benblack86:
Can we get a test?

colossus-tests/src/test/scala/colossus/protocols/http/HttpHeadSpec.scala
```diff
@@ -91,7 +90,7 @@ class HttpHeadSpec extends WordSpec with MustMatchers {
 
   "DateHeader" must {
 
-    val formatter        = new SimpleDateFormat(DateHeader.DATE_FORMAT)
+    val formatter        = DateHeader.createFormatter()
```
dxuhuang: Test is here. In fact this is what inspired the refactoring in the first place

DanSimon: I think @benblack86 means it might make sense to have another test that compares against a hardcoded date, since these tests passed with the incorrect format as well.

PR #613, opened 2017-10-26
Filters

aliyakamercan:
Adding ability to add a list of filters to `GenRequestHandler` which can filter & modify requests & response. This will make configuring tracing and gzipping response easier. 

@benblack86 @DanSimon @jlbelmonte @amotamed @dxuhuang

### DIFF
--- a/colossus/src/main/scala/colossus/service/RequestHandler.scala
+++ b/colossus/src/main/scala/colossus/service/RequestHandler.scala
@@ -46,7 +46,7 @@ abstract class GenRequestHandler[P <: Protocol](val serverContext: ServerContext
 
   protected def handle: PartialHandler[P]
   protected def unhandledError: ErrorHandler[P]
-  protected def filters: Seq[Filter[P]]= Seq.empty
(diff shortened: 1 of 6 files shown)


colossus/src/main/scala/colossus/service/RequestHandler.scala
```diff
@@ -46,11 +46,15 @@ abstract class GenRequestHandler[P <: Protocol](val serverContext: ServerContext
...
 
   protected def onError: ErrorHandler[P] = Map()
 
-  private lazy val fullHandler: PartialHandler[P] = handle orElse {
-    case req => Callback.failed(new UnhandledRequestException(s"Unhandled Request $req"))
+  private lazy val fullHandler: PartialHandler[P] = {
+    val handler: PartialHandler[P] = handle orElse {
+      case req => Callback.failed(new UnhandledRequestException(s"Unhandled Request $req"))
+    }
+    filters.foldLeft(handler){case (h, f) => f.apply(h)}
```
benblack86: Can you name these variables

PR #625, opened 2017-11-06
Service client

benblack86:
Second attempt at service client with built in load balancing. Some of this code can be removed (hang over from previous attempt), but I wanted some opinions on whether this is a better approach. I thought harder about how it will be configured, which resulted in the load balancer being more integrated into the service client logic. I'll add comments to the important parts.

@DanSimon @amotamed @aliyakamercan @jlbelmonte @dxuhuang


colossus-docs/src/main/scala/PlayingWithServiceClientCallbacks.scala
```diff
@@ -0,0 +1,49 @@
...
+
+      val hosts = Seq(
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379)
+      )
+
+      // if we need to change the hosts, then we would create a new instance
```
benblack86: i.e. no need to add/remove clients. Instead disconnect the old redis client and create a new redis client.

colossus-tests/src/test/scala/colossus/service/ServiceClientSpec.scala
```diff
@@ -34,7 +34,7 @@ class ServiceClientSpec extends ColossusSpec with MockFactory {
     val address    = new InetSocketAddress("localhost", 12345)
     val fakeWorker = FakeIOSystem.fakeWorker
     val config = ClientConfig(
-      address,
+      Seq(address),
```
benblack86: `ClientConfig` now takes a list of addresses.

colossus/src/main/resources/reference.conf
```diff
@@ -84,13 +86,18 @@ colossus{
       }
       idle-timeout : "Inf"
       max-response-size : "1 MB"
+      request-retry-policy : {
```
benblack86: This is how to configure from config.

colossus/src/main/scala/colossus/protocols/http/HttpClient.scala
```diff
@@ -7,15 +8,13 @@ trait BaseHttpClient[M[_], B, P <: BaseHttp[B]] extends LiftedClient[P, M] {
 
   def ops: MessageOps[HttpRequestHead, B, P#Request] // = implicitly[MessageOps[HttpRequestHead, B, P#Request]]
 
-  protected lazy val hostHeader = clientConfig.map(config => {
-    HttpHeader(HttpHeaders.Host, config.address.getHostName)
-  })
+  protected lazy val hostHeader = HttpHeader(HttpHeaders.Host, address.getHostName)
```
benblack86: Had to add `address` method to `Sender` in order to make this logic work.

colossus/src/main/scala/colossus/protocols/http/streaming/StreamingHttpClient.scala
```diff
@@ -18,19 +18,22 @@ object StreamingHttpClient {
...
-          new StreamHttpClientCodec
-        ),
-        client
-      )
-      worker.worker ! WorkerCommand.Bind(handler)
-      new BasicLiftedClient(client, Some(config))(implicitly[AsyncBuilder[Callback, WorkerRef]].build(worker))
-      with StreamingHttpClient
-    }
+    override def apply(config: ClientConfig)(implicit worker: WorkerRef): StreamingHttpClient = {
+      // TODO not sure how this will work under a load balancer
```
benblack86: Presumably we need to send to the same client for streaming.... spent zero time thinking about this.

colossus/src/main/scala/colossus/service/LoadBalancer.scala
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+  }
+
+  override def disconnect(): Unit = {
+    clients.foreach(_.disconnect())
+  }
+
+  // TODO does this need to be a client
+  override def connectionStatus = ???
```
benblack86: `Client` extends `Sender` and adds `connectionStatus`. If it is only used for testing, maybe there is a better way to test it without needing this method....

colossus/src/main/scala/colossus/service/LoadBalancerSettings.scala
```diff
@@ -0,0 +1,37 @@
+package colossus.service
+
+import colossus.core.RetryPolicy
+import com.typesafe.config.Config
+
+case class LoadBalancerSettings(retryPolicy: RetryPolicy, selection: SelectionPolicy)
+
+object LoadBalancerSettings {
```
benblack86: Config loading logic updated to match the style elsewhere in codebase.

colossus/src/main/scala/colossus/service/ServiceDSL.scala
```diff
@@ -179,31 +185,37 @@ trait ClientFactory[P <: Protocol, M[_], +T <: Sender[P, M], E] {
...
+      //TODO : binding a client needs to be split up from creating the connection handler
+      // we should make a method called "create" the abstract method, and have
+      // this apply call it, then move this to a more generic parent type
+      val base    = new ServiceClient[P](address, config, worker.generateContext())
+      val handler = connectionHandler(base, codecProvider)
+      worker.worker ! WorkerCommand.Bind(handler)
+      base
+    }
+
+    new LoadBalancer[P, Client[P, Callback]](clients, config)(worker.callbackExecutor)
```
benblack86: Pass the list of created clients to the load balancer to manage.

colossus-docs/src/main/scala/PlayingWithServiceClientCallbacks.scala
benblack86: To be able to update the client/load balancer we would need to have an `update` method on the `Sender` trait, which is fine, but I suppose I was trying to avoid touching it. Your point about connection churn makes a lot of sense.

benblack86:
In response to your first point, I think we can make it updatable.

I don't think forcing every client to use a load balancer is a problem since in the basic case of one client and no retry policy, we can just call send directly without extra wrapping.

At least for our use cases at Tumblr we hardly ever create a raw client. Where we do use the raw client it is only because we don't care about failure. However, in most cases where we use 3rd party http clients (because we need to use SSL) we tend to have retry logic. Having retry logic built into colossus clients would cut down on this copy & paste retry code.

On the issue of "framework of frameworks" vs "application framework", I'm of the opinion that colossus be used in the pursuit of building (micro)services. Our use cases all point to an application framework. I don't think that is a big change from what we have. I think that means adding more functionality (http compression, SSL, request tracing) and making colossus easy to use (load balancing built in, more documentation). If we can build a framework that handles all use cases at Tumblr, I think we will have created a framework that covers the majority of other people's use cases.

I find it difficult to visualize what a "framework of frameworks" would be or how we get there. What are the use case for other developers? If we built a new application framework on top of colossus, I would be concerned that would slow down development as we would have two frameworks to understand.

DanSimon:
> I don't think forcing every client to use a load balancer is a problem since in the basic case of one client and no retry policy, we can just call send directly without extra wrapping.

I don't think we can do this if we also allow the host list to change.  What happens if you start a client with one host, no wrapping, and then want to add a second host (not too concerned about the vise versa)?  But if we can come up with a solution to that, then I have fewer qualms with the PR.

The "framework of frameworks" vs "application framework" drove a lot of the decisions in 0.9.0, but I realize I wasn't completely clear about what I meant with that.  I agree that at the end of the day, _Colossus is a framework to build services_.  As a user building a service, I want to have as little friction as possible.  The whole **FoF** vs **AF** (hopefully I don't start introducing too many abbreviations) is primarily to drive how we properly support different notions of what a "service" is and how we organize the code.

0.9.0 had a major shift where protocols now own almost all of the traits/classes that users directly interact with.  For example, users no longer use `ServiceClient` directly, but instead use a protocol-specific interface like `HttpClient` etc.  This is because every protocol works differently and needs to support a different set of features, and doing things this way helps avoid "kitchen sink" designs where we try to build everything possible into monolithic core classes.  This doesn't scale, so instead we need to think of each protocol as a user itself of a more generic FoF that provides its features in a modular and composable fashion, such that the protocol is able to build its own types and API with only what it exactly needs.

That said, I do hope we can keep things modular enough so if somebody wanted to use all or just part of Colossus as the basis for a larger framework, they could do so without having to hack around a lot of stuff.

benblack86:
My plan is to update the client to make it updatable, clean up the code and write docs. Will reopen once ready.

PR #626, opened 2017-11-07
Fix publish script for snapshots

aliyakamercan:
@amotamed @jlbelmonte @benblack86 @DanSimon @dxuhuang 

Checking `-SNAPSHOT` git tag. 

I am not sure about this. Maybe we should release snapshots with each merge to develop (<version>_<commit sha>-SNAPSHOT) ?

Any preferences?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,6 +106,27 @@ In order to accept your pull request, we need you to submit a Contributor Licens
     
 6. Merge master into develop and update version in `version.sbt` to point to the next snapshot.
 
+# Publishing Snapshots
+
+1. Update version ending with `-SNAPSHOT` in `version.sbt` 
+    ```bash
+    git add version.sbt
+    git commit -m "v{VERSION NUMBER}"
+    ```
+2. Create tag using just the version with `-SNAPSHOT`.
+    ```bash
+    git tag -a v{VERSION NUMBER}
+    ```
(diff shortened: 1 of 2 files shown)


benblack86:
What is `TRAVIS_BRANCH`?

aliyakamercan:
It is the tag in this case. For PRs there are 2 builds;  PR branch and the branch it is being merged to.

DanSimon:
Personally I don't think think auto-publishing a snapshot with merges to develop is particularly useful, so I'm fine with dropping that.  Someone asked for it a long time ago and I don't know if anyone has actually used it.  

But that said, what's the reason for making this change?  I don't think we should be publishing very many snapshots anyway.  It's easy enough to build and publish Colossus locally if you want to test a change you just made.  And we should be publishing milestone/candidate releases often enough for any pre-release testing.

aliyakamercan:
@dan Yes, I guess we can release milestones and releases candidates of off develop. Regardless
```bash
($TRAVIS_PULL_REQUEST == "false" && $TRAVIS_BRANCH == "develop" && $(cat version.sbt) =~ "-SNAPSHOT")
```
this piece of code is wrong since branch cannot be `develop` on a non-pr build. I will remove that bit altogether, we will only be able to make milestone, rc and regular releases.

aliyakamercan:
@benblack86 @DanSimon please take a look again.

PR #629, opened 2017-11-08
Worker manager refactor

### LINKED ISSUE #619: Improve logging in WorkerManager
Mainly by adding io system name to output so we know which one is which when running multiple io systems.

benblack86:
Main aim was just to improve the logging coming from the worker manager by adding the io system name to each log statement. Whilst I was there I also:

+ Increased `maxTries` for the `waitForStatus` test method as I was getting random failures on my slow computer.
+ Renamed `WorkerAgent` to the more accurate name of `WorkerRefs`.
+ Moved `WorkerFactory` out of `WorkerManager` file.
+ Generally refactored the code.

No logic changes! Fixes #619

@aliyakamercan @jlbelmonte @amotamed @dxuhuang @DanSimon

### DIFF
--- a/colossus/src/main/scala/colossus/core/WorkerManager.scala
+++ b/colossus/src/main/scala/colossus/core/WorkerManager.scala
@@ -1,84 +1,84 @@
 package colossus.core
 
-import akka.actor._
 import akka.pattern.ask
 import akka.routing.RoundRobinGroup
 import akka.util.Timeout
 import java.net.InetSocketAddress
 
+import akka.actor.{Actor, ActorRef, OneForOneStrategy, PoisonPill, Props, Stash, Terminated}
+import akka.actor.SupervisorStrategy.Restart
 import colossus.metrics.logging.ColossusLogging
 import colossus.{IOCommand, IOSystem}
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.collection.immutable.Iterable
+import scala.collection.mutable.ArrayBuffer
 import scala.util.{Failure, Success}
 
 /**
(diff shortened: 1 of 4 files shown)


colossus/src/main/scala/colossus/core/WorkerManager.scala
```diff
@@ -1,84 +1,84 @@
...
 
   /*
    * This strategy is chosen specifically for durability.  We don't want to kill the entire system if workers start
    * excepting.
    */
-  override val supervisorStrategy = {
+  override val supervisorStrategy: OneForOneStrategy = {
     OneForOneStrategy() {
-      case _: Exception => Restart
+      case e: Exception =>
```
benblack86: I don't think we need this supervisorStrategy since by default an exception was cause a restart. However, just to be extra cautious I will log first and see if this is actually happening in the real world.

benblack86:
@amotamed I believe both `WorkerFactory` and `DefaultWorkerFactory` should only be used internally, therefore kept private.

colossus/src/main/scala/colossus/core/WorkerManager.scala
```diff
@@ -1,84 +1,84 @@
...
         context.system.scheduler.scheduleOnce(IdleCheckFrequency, self, IdleCheck)
         unstashAll()
         context.become(running)
       } else {
         context.become(waitForWorkers(nowReady))
       }
-    }
-    case other => stash()
+
+    case _ =>
```
jlbelmonte: don't we want to keep a reference to this message and log it?

benblack86: Don't think so, at least we aren't at the moment. Everything will be unstashed later.

PR #635, opened 2017-11-13
Servers can set address(es) to bind to

### LINKED ISSUE #624: Adding functionality to specify which IP address will be bind to when starting the server.
Hi all,

It seems currently the IP address is hard coded([here](https://github.com/tumblr/colossus/blob/99c76e710257bcb075db8d4bb52cb8861efac49a/colossus/src/main/scala/colossus/core/server/Server.scala#L114)) when you was trying to start a server.

It will be better if I can specify which IP addresses will be binded when I have multiple IP address on my server.

Would you mind to add this functionality to colossus?

dkarivalis:
Fixes #624 

@benblack86 @aliyakamercan @amotamed @jlbelmonte @DanSimon @dxuhuang

### DIFF
--- a/colossus-metrics/src/main/scala/colossus/metrics/ConfigHelper.scala
+++ b/colossus-metrics/src/main/scala/colossus/metrics/ConfigHelper.scala
@@ -3,9 +3,9 @@ package colossus.metrics
 import java.net.InetSocketAddress
 
 import com.typesafe.config.{Config, ConfigFactory}
+
 import scala.concurrent.duration._
 import scala.util.Try
-
 import scala.collection.JavaConverters._
 
 //has to be a better way
@@ -20,6 +20,8 @@ object ConfigHelpers {
     def getLongOption(path: String): Option[Long] = getOption(path, config.getLong)
 
     def getBoolOption(path: String): Option[Boolean] = getOption(path, config.getBoolean)
+
+    def getStringListOption(path: String): Option[Seq[String]] = getOption(path, config.getStringList).map(_.asScala)
 
     private def getOption[T](path: String, f: String => T): Option[T] = {
       if (config.hasPath(path)) {
@@ -37,6 +39,8 @@ object ConfigHelpers {
     def getFiniteDurationOption(path: String): Option[FiniteDuration] = getOption(path, getFiniteDuration)
 
     def getScalaDuration(path: String): Duration = Duration(config.getString(path))
+
+    def getAddresses(path:String): Seq[String] = getStringListOption(path).getOrElse(Seq.empty)
 
     private def finiteDurationOnly(str: String, key: String) = {
       Duration(str) match {
--- a/colossus/src/main/scala/colossus/core/ServerSettings.scala
+++ b/colossus/src/main/scala/colossus/core/ServerSettings.scala
@@ -17,6 +17,7 @@ import scala.concurrent.duration._
   * `lowWatermarkPercentage`.  This can be totally disabled by just setting both
   * watermarks to 1.
   *
+  * @param addresses List of addresses to bind to. Defaults to [ "0.0.0.0" ]
   * @param port Port on which this Server will accept connections
   * @param maxConnections Max number of simultaneous live connections
   * @param lowWatermarkPercentage Percentage of live/max connections which represent a normal state
@@ -42,6 +43,7 @@ import scala.concurrent.duration._
   * completes the shutdown.
   */
 case class ServerSettings(
+    addresses: Seq[String] = Seq.empty,
     port: Int,
     maxConnections: Int = 1000,
     maxIdleTime: Duration = Duration.Inf,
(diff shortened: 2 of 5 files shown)


colossus-metrics/src/main/scala/colossus/metrics/ConfigHelper.scala
```diff
@@ -38,6 +39,13 @@ object ConfigHelpers {
 
     def getScalaDuration(path: String): Duration = Duration(config.getString(path))
 
+    def getAddresses(path:String): Seq[String] = try {
+      config.getStringList(path).asScala
+    } catch {
+      case m: Missing => Seq("localhost")
+      case wt: WrongType => throw new WrongType(config.origin(), wt.getMessage)
```
benblack86: Seems like you are catching and then throwing the same exception. You don't need to catch it.

colossus/src/main/scala/colossus/core/server/Server.scala
```diff
@@ -84,7 +84,9 @@ private[colossus] class Server(io: IOSystem, serverConfig: ServerConfig, stateRe
   }
 
   val ss: ServerSocket = ssc.socket()
-  val address          = new InetSocketAddress(settings.port)
+  val addresses: Seq[InetSocketAddress] =
+    serverConfig.settings.addresses.map(address => new InetSocketAddress(address, settings.port)) // "localhost" default
```
benblack86: Is `InetSocketAddress("localhost", settings.port)` the same as `InetSocketAddress(settings.port)`. Seemed a little more complicated under the hood?

DanSimon: I'm not sure we actually want to default to localhost.  `new InetSocketAddress(settings.port)` defaults to binding with the "wildcard" address, aka `0.0.0.0`, but usually localhost resolves to `127.0.0.1`.  So I'm pretty sure defaulting to localhost means that it'll only accept connections coming from the same machine.

Though on the other hand, defaulting to bind to localhost is a much more secure default I suppose.  But we'd have to be _extra_ clear about it in all docs, especially since it would be changing default behavior and could cause mass panic as unsuspecting devs roll out a "version upgrade" that suddenly stops accepting all incoming connections.

benblack86: I think keeping the wildcard address is probably sensible for now.

colossus/src/main/scala/colossus/core/server/Server.scala
```diff
@@ -109,7 +111,7 @@ private[colossus] class Server(io: IOSystem, serverConfig: ServerConfig, stateRe
   def start() = {
     //setup the server
     try {
-      ss.bind(address, settings.tcpBacklogSize.getOrElse(0))
+      addresses.foreach(address => ss.bind(address, settings.tcpBacklogSize.getOrElse(0)))
       info(s"name: Bound to port ${settings.port}")
```
benblack86: We should update this log statement to be a little more helpful:

```
info(s"name: Bound to ${addresses.mkStr(", ")}")
```

colossus/src/main/scala/colossus/core/server/Server.scala
dkarivalis: agreed that sticking to the wildcard address will lead to a better version-upgrade experience. 

interesting point on the security aspect of using localhost .. possibly something to consider when doing a bigger version bump that includes a migration guide?

dkarivalis:
@benblack86 @DanSimon updated!

colossus/src/main/scala/colossus/core/ServerSettings.scala
```diff
@@ -42,6 +43,7 @@ import scala.concurrent.duration._
   * completes the shutdown.
   */
 case class ServerSettings(
+    addresses: Seq[String] = Seq("0.0.0.0"),
```
benblack86: I think it would be safer to have the empty list as default, then if the list is empty use `new InetSocketAddress(settings.port)`.

colossus-metrics/src/main/scala/colossus/metrics/ConfigHelper.scala
```diff
@@ -38,6 +39,12 @@ object ConfigHelpers {
 
     def getScalaDuration(path: String): Duration = Duration(config.getString(path))
 
+    def getAddresses(path:String): Seq[String] = try {
+      config.getStringList(path).asScala
+    } catch {
+      case _: Missing => Seq("0.0.0.0")
```
amotamed: I would hold off hardcoding the string, also, please use Try rather than try.  Here we should do seq.empty and then on empty only initialize the port.  0.0.0.0 is an ipv4 specific address

colossus/src/main/scala/colossus/core/server/Server.scala
```diff
@@ -84,7 +84,9 @@ private[colossus] class Server(io: IOSystem, serverConfig: ServerConfig, stateRe
   }
 
   val ss: ServerSocket = ssc.socket()
-  val address          = new InetSocketAddress(settings.port)
+  val addresses: Seq[InetSocketAddress] =
+    serverConfig.settings.addresses.map(address => new InetSocketAddress(address, settings.port)) // "0.0.0.0" default
```
amotamed: here I'd do an if empty new InestSockeAddress w/ the port otherwise do the map

dkarivalis:
@amotamed @benblack86 @DanSimon updated!

PR #637, opened 2017-11-13
Service client

benblack86:
Same as before, but now the client can be updated. Every client will have a load balancer layer. I shall comment below. Obviously, this still needs tests and documentation.

@aliyakamercan @jlbelmonte @amotamed @dxuhuang @DanSimon

### DIFF
--- a/colossus-docs/src/main/scala/PlayingWithServiceClientCallbacks.scala
+++ b/colossus-docs/src/main/scala/PlayingWithServiceClientCallbacks.scala
@@ -1,0 +1,58 @@
+import java.net.InetSocketAddress
+
+import akka.actor.Actor.Receive
+import akka.actor.ActorSystem
+import akka.util.ByteString
+import colossus.IOSystem
+import colossus.metrics.logging.ColossusLogging
+import colossus.protocols.http.HttpMethod._
+import colossus.protocols.http.UrlParsing._
+import colossus.protocols.http.{Http, HttpServer, Initializer, RequestHandler}
+import colossus.protocols.redis.Redis
+import colossus.service.GenRequestHandler.PartialHandler
+
+import scala.util.{Failure, Success, Try}
+
+object PlayingWithServiceClientCallbacks extends App with ColossusLogging {
+
+  implicit val actorSystem = ActorSystem()
+  implicit val ioSystem    = IOSystem()
+  implicit val ec          = actorSystem.dispatcher
+
+  val serverRef = HttpServer.start("example-server", 9000) { initContext =>
+    new Initializer(initContext) {
+
+      val hosts = Seq(
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379),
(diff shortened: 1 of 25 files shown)


colossus-docs/src/main/scala/PlayingWithServiceClientCallbacks.scala
```diff
@@ -0,0 +1,58 @@
...
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379),
+        new InetSocketAddress("localhost", 6379)
+      )
+
+      val redisClient = Redis.client(hosts)
+
+      override def receive: Receive = {
+        case _ => redisClient.update(hosts)
```
benblack86: Create the client and then update at will

colossus/src/main/scala/colossus/service/Async.scala
```diff
@@ -14,9 +16,12 @@ trait Sender[P <: Protocol, M[_]] {
   def send(input: P#Request): M[P#Response]
 
   def disconnect()
-  
+
   def addInterceptor(interceptor: Interceptor[P]): Unit
 
+  def address(): InetSocketAddress
```
benblack86: Every sender now has an address and update method.

colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -447,4 +456,7 @@ class ServiceClient[P <: Protocol](
   def disconnect() {
     connection.disconnect()
   }
+
+  override def update(addresses: Seq[InetSocketAddress]): Unit =
+    throw new NotImplementedError("Can not directly update service client")
```
benblack86: `ServiceClient` cannot be updated directly. Updates are handled by the load balancing layer.

colossus/src/main/scala/colossus/service/Async.scala
DanSimon: I don't think it has to be solved in this PR, but it's definitely weird that `address()` returns a single address while `update` takes in sequence.  I mean it makes perfect sense in the context of what we're doing with them, but purely as methods on a trait it feels like we're trying to make `Sender` do too many things now.  We shouldn't need to have implementations that just immediately throw a "you can't do this with this subtype" exception.

colossus/src/main/scala/colossus/service/LoadBalancer.scala
```diff
@@ -0,0 +1,70 @@
...
+        go(retryList.head, retryList.tail, 1)
+      }
+    }
+  }
+
+  override def disconnect(): Unit = {
+    clients.foreach(_.disconnect())
+  }
+
+  // TODO is this just used for tests?
```
DanSimon: >is this just used for tests?

Possibly, though I think it would be nice to know if a load balancer is still maintaining active connections or has been shutdown.  This is probably not the best way to do that, though.

colossus/src/main/scala/colossus/service/Async.scala
benblack86: Totally agree. Going to leave that for later.

benblack86:
Going to re-open as a new PR... again :)

PR #639, opened 2017-11-27
Service client

benblack86:
Service client with load balancer and retries built in! Includes tests and doc changes.

Overall I feel like this is moving in the correct direction, but I think it can be cleaned up in future PRs. My problem is that we are squishing different functionality into the client trait. I think what we currently call clients, should be renamed connections. A client would then contain one-to-many connections, which it can manage. You can conceive that if you want to skip the client and just talk to the connection, that should be possible. Anyways, that can wait for another day.

@aliyakamercan @jlbelmonte @amotamed @DanSimon


amotamed:
LGTM 👍

PR #641, opened 2017-11-28
Colossus root project name

benblack86:
Change the root level project to `colossus-root` so it doesn't just show up as `root` in IntelliJ menu.

@aliyakamercan @jlbelmonte @amotamed @DanSimon


dkarivalis:
Shows up as `colossus-root` in the project tool window:
screen shot 2017-11-29 at 4 18 44 pm

👍

PR #642, opened 2017-11-28
Delete scatter gather code

benblack86:
A question in the form of a PR. What is this scatter gather logic? Seems like something left over from rediscover? Seems unrelated to Redis protocol so I'm deleting it.

@DanSimon @jlbelmonte @amotamed @aliyakamercan


DanSimon:
Yeah it can be removed.  It is completely generic but I don't think serves any real purpose anymore.

PR #643, opened 2017-11-29
Simplify redis API

### LINKED ISSUE #632: Inconsistency between redis and memcache interfaces
More of a question for @DanSimon 

If sending a get request to redis you should generally use `mapTry`:

```
redisClient.get(ByteString(name)).mapTry {
  case Success(value) =>
    // data found
  case Failure(_) =>
    // no data found
}
```

However, memcache returns an option so `map` is sufficient unless you want to handle the case where a incorrect response is sent by memcached.

```
memcacheClient.get(ByteString(name)).map {
  case Some(Value(_, data, _)) =>
    // data found
  case None =>
    // no data found
}
```

Any particular reason for this? Should we make it consistent? Looks like http follows the same style as redis. Which style is the best?

benblack86:
This PR simplifies the Redis API by merging calls that return an option with their counterparts that don't. In other words, returning an option is the default behavior. See https://github.com/tumblr/colossus/issues/632 for more details. Fixes #632 

@DanSimon @aliyakamercan @jlbelmonte @amotamed

### DIFF
--- a/colossus-tests/src/it/scala/colossus/RedisHashITSpec.scala
+++ b/colossus-tests/src/it/scala/colossus/RedisHashITSpec.scala
@@ -2,7 +2,7 @@ package colossus
 
 import akka.util.ByteString
 
-class RedisHashITSpec extends BaseRedisITSpec{
+class RedisHashITSpec extends BaseRedisITSpec {
 
   val keyPrefix = "colossusITHash"
 
@@ -11,9 +11,9 @@ class RedisHashITSpec extends BaseRedisITSpec{
     val field1 = ByteString("field")
     val field2 = ByteString("field2")
     val field3 = ByteString("field3")
-    val val1 = ByteString("value1")
-    val val2 = ByteString("value2")
-    val val3 = ByteString("value3")
+    val val1   = ByteString("value1")
+    val val2   = ByteString("value2")
+    val val3   = ByteString("value3")
 
     "hdel" in {
       val delKey = getKey()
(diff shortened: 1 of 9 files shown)


benblack86:
Looks like I need to fix some integration tests.

colossus-tests/src/it/scala/colossus/RedisHashITSpec.scala
```diff
@@ -23,10 +23,10 @@ class RedisHashITSpec extends BaseRedisITSpec{
         _ <- client.hset(delKey, field3, val3)
         x <- client.hdel(delKey, field1, field2)
         y <- client.hget(delKey, field3) //should be unaffected
-        z <- client.hgetOption(delKey, field1) //snould be none, since it was deleted
-      } yield { (x,y,z) }
+        z <- client.hget(delKey, field1) //snould be none, since it was deleted
```
amotamed: let's get that should fixed :)

colossus-tests/src/it/scala/colossus/RedisSetITSpec.scala
```diff
@@ -22,7 +22,7 @@ class RedisSetITSpec extends BaseRedisITSpec {
         (x, y)
       }
 
-      res.futureValue must be ((3, Set(val1, val2, val3)))
+      res.futureValue must be((3, Set(val1, val2, val3)))
```
amotamed: I prefer the space since we're doing that with must, thoughts?

benblack86: This is scalafmt. Since there are brackets involved, I'm guessing it removes the space. Personally, I prefer to use assert(...) instead of must/must be/etc.

PR #645, opened 2017-11-30
Jlbelmonte/gzip

jlbelmonte:
Using  colossus service filters I added a couple of filters one for Http protocol and another for StreamingHttp to add support for Gzip and Deflate.

I have extracted the Compressor and Zlib compressor out from colossus.protocol.memcache package to colossus.util.

I also added a Gzip compressor.

@benblack86 @aliyakamercan @amotamed @DanSimon


jlbelmonte:
going to close this PR until I fix the test

colossus-docs/src/main/scala/StreamingHttpExample.scala
```diff
@@ -24,6 +27,12 @@ object StreamServiceExample {
           )
         }
       }
+
+      case _ => Callback.successful(StreamingHttpResponse(
```
benblack86: What's this about?

colossus-docs/src/main/scala/StreamingHttpExample.scala
```diff
@@ -1,19 +1,22 @@
...
-import service._
+import colossus.core._
+import colossus.protocols.http._
+import colossus.protocols.http.filters.HttpStreamCustomFilters
+import colossus.protocols.http.streaming._
+import colossus.service._
 import colossus.streaming._
 
-object StreamServiceExample {
+object StreamServiceExample extends App {
```
benblack86: I suppose this is normally run from `Main`.

colossus/src/main/scala/colossus/util/Compressor.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import akka.util.{ByteString, ByteStringBuilder}
+
+trait Compressor {
+  def compress(bytes: ByteString): ByteString
+
+  def decompress(bytes: ByteString): ByteString
+}
+
+object NoCompressor extends Compressor {
```
benblack86: Can you split these up into different files?

colossus/src/main/scala/colossus/protocols/http/filters/HttpCustomFilters.scala
```diff
@@ -0,0 +1,39 @@
...
+import colossus.service.GenRequestHandler.PartialHandler
+import colossus.util.{GzipCompressor, ZCompressor}
+
+
+object HttpCustomFilters {
+
+  class CompressionFilter(bufferedKb: Int = 10) extends Filter[Http] {
+
+    override def apply(next: PartialHandler[Http]): PartialHandler[Http] = {
+      case request =>
```
benblack86: Any way we can get rid of the duplicate logic between this and the streaming version?

colossus-docs/src/main/scala/StreamingHttpExample.scala
jlbelmonte: Other examples are runnable too like HttpServiceExample, HelloWorld, Filter example. I think examples that can run in a standalone manner are helpful

colossus-docs/src/main/scala/StreamingHttpExample.scala
jlbelmonte: When the example run, if only one route is defined, and other is hit nothing happens, this way we can return a 404 and give feedback when interacting with the example

benblack86: Sounds like a bug? Can you create a ticket if so.

colossus-docs/src/main/paradox/streaming.md
```diff
@@ -136,3 +136,11 @@ body of the request/response is a stream of message chunks.
 
 @@snip [StreamingHttpExample.scala]($examples$/StreamingHttpExample.scala) { #streaming_http }
 
+## Enabling compression for the streaming services
```
benblack86: Shorter title would be better since it shows up in the side bar `## Compression`

PR #647, opened 2017-12-06
Add content-type header to responses

dkarivalis:
This fixes a bug where a content-type set in an `HttpBodyEncoder` is never written in the response header. According to [RFC 2616](https://tools.ietf.org/html/rfc2616#section-7.2.1) we _should_ include the content-type header, also adhering to [techempower's](https://www.techempower.com/benchmarks/#section=code&hw=ph&test=json) json serialization requirement 6

Note that when benchmarking our [example json route](https://github.com/tumblr/colossus/blob/master/colossus-examples/src/main/scala/colossus.examples/BenchmarkExample.scala#L28) this causes Requests/sec to decrease because we do more work (writing the header), and Transfer/sec to increase because we're transferring more data (additional bytes for the extra header)

@benblack86 @aliyakamercan @amotamed @DanSimon @jlbelmonte


benblack86:
I suppose this PR broke it? https://github.com/tumblr/colossus/pull/569/files

dkarivalis:
> I suppose this PR broke it? https://github.com/tumblr/colossus/pull/569/files

Yeah, I don't think it was caught in tests because we have several ways to add a content-type header

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -411,10 +411,9 @@ object Connection {
 }
 
 object ContentType {
-
+  val HeaderKey       = "Content-Type"
```
benblack86: This value is already defined in `HttpHeaders`

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -135,7 +135,14 @@ trait HttpResponseBuilding {
   def initialVersion: HttpVersion
 
   def respond[T: HttpBodyEncoder](code: HttpCode, data: T, headers: HttpHeaders = HttpHeaders.Empty) = {
-    HttpResponse(HttpResponseHead(initialVersion, code, headers), HttpBody(data))
+    HttpResponse(HttpResponseHead(initialVersion, code, headersWithContentType[T](headers)), HttpBody(data))
+  }
+
+  def headersWithContentType[T: HttpBodyEncoder](headers: HttpHeaders = HttpHeaders.Empty): HttpHeaders = {
+    implicitly[HttpBodyEncoder[T]].contentType match {
```
benblack86: Not sure this is going to work. What happens if you test:

```
HttpResponse.ok("hello").withContentType("other")
```

dkarivalis: IMO this is a separate issue / outside the scope of this PR, as this is fixing the current architecture

you can also do
```
HttpResponse.ok("Hello", HttpHeaders(HttpHeader("Content-Type", "first"))).withContentType("second")
```
to end up with multiple content-type's in the response header

colossus/src/main/scala/colossus/protocols/http/Header.scala
dkarivalis: i saw [this comment](https://github.com/dkarivalis/colossus/blob/7c1bceda598665523806e4f39ecbac4582b66c91/colossus/src/main/scala/colossus/protocols/http/Header.scala#L286) and figured it may be cleaner to have two (one for lower case, another for mixed case)

dkarivalis: ah just read in https://tools.ietf.org/html/rfc2616#section-4.2 that `[Header] Field names are case-insensitive`. going to switch to the already defined variable ^_^

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
DanSimon: I agree this is kind of part of a larger issue, since you can have the same problem with other headers we generate like `Date` and `Content-Length`.  

Hopefully it should be clear that by requiring the body type to have a `HttpBodyEncoder` the response will automatically have the right content-type, but we should definitely make sure though to have clear documentation about this.  Maybe we should add scaladoc to `respond` noting this.

dkarivalis: opened an issue to move the conversation out of this PR: https://github.com/tumblr/colossus/issues/649

colossus/src/main/scala/colossus/protocols/http/HttpBody.scala
```diff
@@ -7,6 +7,8 @@ import scala.util.Try
 
 class HttpBody(private val body: Array[Byte]) {
 
+  protected[http] var contentType: String = ""
```
benblack86: Lets rename this variable to indicate what is being used for, maybe something like `suggestedContentTypeFromEncoder`

colossus/src/main/scala/colossus/protocols/http/HttpResponse.scala
```diff
@@ -85,6 +85,9 @@ case class HttpResponse(head: HttpResponseHead, body: HttpBody) extends Encoder
 
   def encode(buffer: DataOutBuffer, extraHeaders: HttpHeaders) {
     head.encode(buffer)
+    if (head.headers.firstValue(HttpHeaders.ContentType).isEmpty && !body.contentType.isEmpty) {
```
benblack86: Comment why this is happening would be good.

Issue #646, opened 2017-12-01
Benchmark latest colossus vs techempower version


benblack86:
See if the latest version is more performant. Also check to see if using the same JSON library as blaze produces better results.

benblack86:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Scala/colossus/src/main/scala/example/Main.scala

benblack86:
Example updated. @dkarivalis will update techempower repo.

PR #648, opened 2017-12-06
Swap benchmark example json lib

dkarivalis:
Addresses https://github.com/tumblr/colossus/issues/646. 
With `wrk -c 50 -t 2 -d 10 --latency -s pipeline.lua http://localhost:9007/json -- 16` on local environment, saw a small performance increase (~5%) in request/sec and transfer/sec for our json benchmarking example

At the least, I think it makes sense to be uniform with [s-server](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Scala/s-server/src/main/scala/woshilaiceshide/sserver/tfb/Server.scala#L9)'s and [blaze](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Scala/blaze/src/main/scala/Main.scala#L6)'s json library for the techempower framework benchmarks

@benblack86 @aliyakamercan @amotamed @DanSimon @jlbelmonte


benblack86:
Do you have a PR to change it in techempower repo?

dkarivalis:
> Do you have a PR to change it in techempower repo?

Thinking we should merge this first so the techempower PR can reference colossus's latest

build.sbt
```diff
@@ -89,9 +89,10 @@ lazy val testkitDependencies = libraryDependencies ++= Seq(
 
 lazy val ExamplesSettings = Seq(
   libraryDependencies ++= Seq(
-    "org.json4s"     %% "json4s-jackson"  % "3.5.3",
-    "ch.qos.logback" %  "logback-classic" % "1.2.2",
-    "org.mockito"    %  "mockito-all"     % "1.10.19"
+    "org.json4s"                   %% "json4s-jackson"       % "3.5.3",
```
benblack86: Can we remove this?

colossus-examples/src/main/scala/colossus.examples/BenchmarkExample.scala
```diff
@@ -3,30 +3,24 @@ package colossus.examples
...
-    def encode(json: JValue) = {
-      HttpBody(compact(render(json)))
-    }
+  implicit object JsonBody extends HttpBodyEncoder[Array[Byte]] {
+    val contentType: String = ContentType.ApplicationJson
+    def encode(json: Array[Byte]): HttpBody = new HttpBody(json)
   }
 
-  val json: JValue = ("message" -> "Hello, World!")
+  def json: Array[Byte] = mapper.writeValueAsBytes(Map("message" -> "Hello, World!"))
```
benblack86: Is there no JObject we can create first? The `writeValueAsBytes` logic would make more sense in the encode method.

colossus-docs/src/test/scala/MyHttpHandlerSpec.scala
```diff
@@ -33,14 +32,16 @@ class MyHttpHandlerSpec extends HttpServiceSpec {
 
     "return 200 and body that satisfies predicate" in {
       expectCodeAndBodyPredicate(HttpRequest.get("ping/1"), HttpCodes.OK) { body =>
```
benblack86: We should leave this code as before since it is cleaner using json4s.

colossus-examples/src/main/scala/colossus.examples/BenchmarkExample.scala
```diff
@@ -3,30 +3,25 @@ package colossus.examples
...
 
-object BenchmarkService {
-
-  implicit object JsonBody extends HttpBodyEncoder[JValue] {
-    val contentType = ContentType.ApplicationJson
-
-    def encode(json: JValue) = {
-      HttpBody(compact(render(json)))
-    }
+  implicit object JsonBody extends HttpBodyEncoder[HelloWorld] {
```
benblack86: I would generalize this a little since this isn't what you would do in the real world. Change this to `AnyRef`.

colossus-examples/src/main/scala/colossus.examples/BenchmarkExample.scala
```diff
@@ -3,30 +3,25 @@ package colossus.examples
...
+    private val mapper: ObjectMapper = new ObjectMapper().registerModule(DefaultScalaModule)
+    val contentType: String = ContentType.ApplicationJson
+    def encode(helloWorld: HelloWorld): HttpBody = new HttpBody(mapper.writeValueAsBytes(helloWorld))
   }
 
-  val json: JValue = ("message" -> "Hello, World!")
+  case class HelloWorld(message: String)
 
-  def start(port: Int)(implicit io: IOSystem) {
+  def helloWorld: HelloWorld = HelloWorld("Hello, World!")
```
benblack86: And then have `def helloWorld: AnyRef = HelloWorld("Hello, World!")`

build.sbt
dkarivalis: need it for `MyHttpHandlerSpec` ^_^

PR #654, opened 2017-12-15
Better exception message format handling

### LINKED ISSUE #594: Create a base trait for some built-in exceptions
There are a bunch of exceptions that are mainly used to indicate an error state has occurred.  `DroppedReplyException`,`RequestTimeoutException`, and `ConnectionLostException` are some examples.  In some (maybe all?) of these cases, the stack trace is completely useless since the exceptions are always generated in the same spot and have no direct relevance to any user code.

We should figure out exactly which exceptions fall into this category and group them under some base type, something like `ColossusRuntimeException`.

benblack86:
Created a `ColossusRuntimeException` trait that is extended by `ServiceClientException`, `ServiceClientExceptions` and `UnhandledRequestException`. I then updated the `RequestFormatter` so it is no longer configurable, but instead will log the name of any `ColossusRuntimeException` and suppress the stack trace. If users want different behavior, then they can write their own `RequestFormatter`. Hopefully this makes the logic a little simpler. Fixes #594

@dkarivalis @DanSimon @aliyakamercan @jlbelmonte @amotamed

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/package.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/package.scala
@@ -43,7 +43,7 @@ package object http {
       def errorResponse(error: ProcessingFailure[HttpRequest]): HttpResponse = error match {
         case RecoverableError(request, reason) =>
           reason match {
-            case _: UnhandledRequestException => request.notFound(s"Not found")
+            case _: UnhandledRequestException => request.notFound("Not found")
             case _                            => request.error(reason.toString)
           }
 
--- a/colossus/src/main/scala/colossus/protocols/http/package.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/package.scala
@@ -40,16 +40,16 @@ package object http {
       ServiceClientFactory.basic("http", () => new HttpClientCodec)
 
     class ServerDefaults {
-      def errorResponse(error: ProcessingFailure[HttpRequest]) = error match {
+      def errorResponse(error: ProcessingFailure[HttpRequest]): HttpResponse = error match {
         case RecoverableError(request, reason) =>
           reason match {
-            case c: UnhandledRequestException => request.notFound(s"Not found")
-            case other                        => request.error(reason.toString)
+            case _: UnhandledRequestException => request.notFound(s"Not found")
+            case _                            => request.error(reason.toString)
           }
-        case IrrecoverableError(reason) => {
+
+        case IrrecoverableError(_) =>
           HttpResponse(HttpResponseHead(HttpVersion.`1.1`, HttpCodes.BAD_REQUEST, HttpHeaders.Empty),
                        HttpBody("Bad Request"))
-        }
       }
     }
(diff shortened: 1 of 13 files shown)


colossus/src/main/scala/colossus/service/RequestHandler.scala
```diff
@@ -58,13 +59,14 @@ abstract class GenRequestHandler[P <: Protocol](val serverContext: ServerContext
...
+
+  private lazy val errorHandler: ErrorHandler[P] = onError orElse unhandledError
 
   def handleFailure(error: ProcessingFailure[Request]): Response = errorHandler(error)
 
   def tagDecorator: TagDecorator[P] = TagDecorator.default[P]
-  def requestLogFormat: Option[RequestFormatter[Request]] =
-    Some(new ConfigurableRequestFormatter[Request](config.errorConfig))
+
+  def requestLogFormat: RequestFormatter[Request] = new StandardRequestFormatter[Request]
```
dkarivalis: may be out of scope of this PR .. `RequestFormatter` is a bit ambiguous - possibly refactor the name to `RequestExceptionLogger`?

colossus/src/main/scala/colossus/protocols/http/package.scala
```diff
@@ -40,16 +40,16 @@ package object http {
       ServiceClientFactory.basic("http", () => new StaticHttpClientCodec)
 
     class ServerDefaults {
-      def errorResponse(error: ProcessingFailure[HttpRequest]) = error match {
+      def errorResponse(error: ProcessingFailure[HttpRequest]): HttpResponse = error match {
         case RecoverableError(request, reason) =>
           reason match {
-            case c: UnhandledRequestException => request.notFound(s"Not found")
-            case other                        => request.error(reason.toString)
+            case _: UnhandledRequestException => request.notFound(s"Not found")
```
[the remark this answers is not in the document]
benblack86: Who needs a linter, when we have @dkarivalis

colossus/src/main/scala/colossus/service/RequestHandler.scala
benblack86: `RequestExceptionFormatter` sounds good

colossus/src/main/scala/colossus/service/ServiceClient.scala
```diff
@@ -92,7 +93,7 @@ object ClientConfig {
   }
 }
 
-class ServiceClientException(message: String) extends Exception(message)
+class ServiceClientException(message: String) extends Exception(message) with ColossusRuntimeException
```
amotamed: this is where we're squashing the stracktrace, we should pass it along here

benblack86: Can we get an example of the error that is happening so we can see if the stack trace is helpful, because for all these `ColossusRuntimeException` the stack trace shouldn't be helpful.

PR #666, opened 2018-01-21
Replace nscala-time with java.time in Cookie class

### LINKED ISSUE #662: Remove nscala-time
We only seem to use it for parsing cookie expire time, which should be possible using standard java libraries. I think this would be good to do since it means one less dependency, one less potential conflict.

pscheid92:
Fixes #662 

Before:
  Used nscala-time library to parse expiration date in Cookie class/object.

After:
  Use java.time utilities to parse expiration date.


pscheid92:
Wasn't sure if other developers depend on it in their feature branches.
I will remove it as soon as I arrive at home.

PR #672, opened 2018-08-24
fix: contentType in StringEncoder should specify charset=utf-8

lxohi:
in order to avoid encoding detection issues in web browsers

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/http/Header.scala
+++ b/colossus/src/main/scala/colossus/protocols/http/Header.scala
@@ -420,7 +420,7 @@ object Connection {
 
 object ContentType {
   val TextPlain       = "text/plain"
-  val TextPlainUtf8   = "text/plain; charset=utf-8"
+  val TextPlainUtf8   = "text/plain; charset=UTF-8"
   val ApplicationJson = "application/json"
   val OctetStream     = "application/octet-stream"
 }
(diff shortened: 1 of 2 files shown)


yahoocla:
Thank you for submitting this pull request, however I do not see a valid CLA on file for you. Before we can merge this request please visit https://yahoocla.herokuapp.com/ and agree to the terms. Thanks! :smile:

andwxh:
close for trigger the re-check of yahoocla

colossus/src/main/scala/colossus/protocols/http/Header.scala
```diff
@@ -420,6 +420,7 @@ object Connection {
 
 object ContentType {
   val TextPlain       = "text/plain"
+  val TextPlainUtf8   = "text/plain; charset=utf-8"
```
dkarivalis: hey @lxohi! would you mind updating `utf-8` -> `UTF-8` to be better aligned with the value definition [here](https://tools.ietf.org/html/rfc3629#section-8)?

dkarivalis:
possibly try to re-sign the CLA? CLA web hook is reporting `No CLA found for this user`

lxohi:
Yes, I've changed charset to uppercase & re-signed the CLA.

dkarivalis (approved):
👍

PR #673, opened 2018-10-23
memcache: implement CAS and GETS

dkarivalis:
what
---
implements the `CAS` and `GETS` commands in our memcache client; more on the commands [here](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L153).

reran [these benchmark tests](https://github.com/tumblr/colossus/blob/master/colossus-tests/src/it/scala/colossus/MemcacheITSpec.scala#L194) with 10 cycles @ 1000 iterations each to confirm that the changes to the parser & writer do not negatively impact the performance of the existing get/set commands. on a mbp with an unstable environment, we have similar results: [benchmark results](https://gist.github.com/dkarivalis/7b07ccd4dea3f0f0535529879bcb493e)

@byxorna @vrrs
@DanSimon @benblack86 @aliyakamercan @amotamed @jlbelmonte @dxuhuang

### DIFF
--- a/colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
+++ b/colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
@@ -25,9 +25,11 @@ import colossus.util.compress.{Compressor, NoCompressor}
 object UnifiedProtocol {
   val ADD     = ByteString("add")
   val APPEND  = ByteString("append")
+  val CAS     = ByteString("cas")
   val DECR    = ByteString("decr")
   val DELETE  = ByteString("delete")
   val GET     = ByteString("get")
+  val GETS    = ByteString("gets")
   val INCR    = ByteString("incr")
   val PREPEND = ByteString("prepend")
   val REPLACE = ByteString("replace")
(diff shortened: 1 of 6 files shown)


benblack86:
Hmmm not sure hehe

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -189,28 +204,36 @@ sealed trait MemcacheWriteCommand extends MemcacheCommand {
...
-
-    val padding = 8
+       8 or 9, depending if <CAS UNIQUE> is present*/
 
     val flagsStr    = ByteString(flags.toString)
     val ttlStr      = ByteString(ttl.toString)
     val dataSizeStr = ByteString(value.size.toString)
+    val casUniqueStr = ByteString(casUniqueMaybe.getOrElse("").toString)
+
+    val padding = 8 + (if (casUniqueStr.nonEmpty) 1 else 0)
```
vrrs: Please create a function that return the padding and perhaps mention where this magic 8 comes from.

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -226,12 +249,17 @@ sealed trait MemcacheWriteCommand extends MemcacheCommand {
     b.append(SP)
 
     b.append(dataSizeStr)
+
+    if (casUniqueStr.nonEmpty) {
+      b.append(SP)
```
vrrs: SP should be named MemcachedMessageDelimeter or something along the line. SP is short for space ?? Not very helpful

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -291,11 +320,18 @@ object MemcacheReplyParser {
...
 
-  def reply = delimitedString(' ', '\r') <~ byte |> { pieces =>
+  def reply = delimitedString(' ', '\r') <~ byte |> { pieces: Seq[String] =>
     pieces.head match {
-      case "VALUE"                   => value(Vector.empty, pieces(1), pieces(2).toInt, pieces(3).toInt)
+      case "VALUE"                   =>
+        if (pieces.size < numberPiecesInCasValueReply) {
+          value(Vector.empty, pieces(1), pieces(2).toInt, pieces(3).toInt)
+        } else {
+          casValue(pieces(1), pieces(2).toInt, pieces(3).toInt, pieces(4).toInt)
```
vrrs: Why in regular value the key is set to Vector.empty but in casValue is set to pieces(1) ?

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -41,73 +43,86 @@ object UnifiedProtocol {
...
 
 //TODO: 'Flags' doesn't fully support the memcached protocol:  ie: using an Int doesn't allow us to utilize the full 32 unsigned int space
-//TODO: implement CAS
-//TODO: incrs/decrs are not utilizing full 64 bit unsigned int space since we are using singed Longs
+//TODO: incrs/decrs are not utilizing full 64 bit unsigned int space since we are using signed Longs
 object MemcacheCommand {
 
   import UnifiedProtocol._
 
+  private def getBytes(commandName: ByteString, compressor: Compressor = NoCompressor)
```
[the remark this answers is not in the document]
byxorna: 👍 are we running linters on this code? Surprised a linter didnt see this...

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -189,28 +204,36 @@ sealed trait MemcacheWriteCommand extends MemcacheCommand {
...
      ----
-       8 */
-
-    val padding = 8
+       8 or 9, depending if <CAS UNIQUE> is present*/
 
     val flagsStr    = ByteString(flags.toString)
     val ttlStr      = ByteString(ttl.toString)
     val dataSizeStr = ByteString(value.size.toString)
+    val casUniqueStr = ByteString(casUniqueMaybe.getOrElse("").toString)
```
byxorna: This seems to be an `Option[Int]` that has `getOrElse[String]` being called. Is that chill?

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
```diff
@@ -189,28 +204,36 @@ sealed trait MemcacheWriteCommand extends MemcacheCommand {
...
+
+    val padding = 8 + (if (casUniqueStr.nonEmpty) 1 else 0)
 
-    val sizeHint = commandName.length + flagsStr.length + ttlStr.length + dataSizeStr.length + value.size + padding
+    val sizeHint = commandName.length +
+                      flagsStr.length +
+                        ttlStr.length +
+                   dataSizeStr.length +
+                  casUniqueStr.length +
```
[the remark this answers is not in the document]
dkarivalis: IMO `casUnique` makes more sense, as it is consistent with the protocol https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L178

colossus/src/main/scala/colossus/protocols/memcache/MemcacheOps.scala
```diff
@@ -65,6 +64,14 @@ trait MemcacheClient[M[_]] extends LiftedClient[Memcache, M] {
   def append(key: ByteString, value: ByteString): M[Boolean] =
     storageAsBooleanCommand(Append(key, value), key)
 
+  def cas(key: ByteString, value: ByteString, ttl: Int = 0, flags: Int = 0, casUnique: Int): M[Boolean] = {
+    executeCommand(Cas(key, value, ttl, flags, Some(casUnique)), key) {
+      case Stored   => success(true)
```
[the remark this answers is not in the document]
dkarivalis: i believe the error/client_error/server_error return codes are handled here: https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/protocols/memcache/MemcacheOps.scala#L33

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
dkarivalis: the padding is explained 7 lines up (https://github.com/tumblr/colossus/pull/673/files#diff-102c569d28658e91e688d75f0650d38fR223)

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
dkarivalis: i believe this is because we support a [`MemcacheOps#getAll(...)`](https://github.com/tumblr/colossus/blob/master/colossus/src/main/scala/colossus/protocols/memcache/MemcacheOps.scala#L85) which may return multiple values

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
dkarivalis: oh hmm, i tested this in `sbt console` and it appears it is only chill because of the `#toString`:

screen shot 2018-11-01 at 4 32 06 pm

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
vrrs: Regardless, you should do extract method. This method is already too long. Then you can extract the comment and put it where this new method is.

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
vrrs: Uhm i dont think that explain why casValue is not value(Vector.empty, pieces(1), pieces(2).toInt, pieces(3).toInt, pieces(4).toInt) I thought casValue simply adds a piece at the end

colossus/src/main/scala/colossus/protocols/memcache/MemcacheOps.scala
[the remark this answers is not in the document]
vrrs: Ohh i see, its using partial funcs. Disregard

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
dkarivalis: to support the `MemcacheOps#getAll(...)` functionality, `value(...)` is a recursive method. to avoid intertwining the CAS logic with the unneeded recursion in our case, `casValue(...)` is its own method

colossus/src/main/scala/colossus/protocols/memcache/Memcache.scala
vrrs: Yeah thats a tricky statement. casUniqueMaybe.getOrElse("") gives you any type. This is a more readable version:
``` 
val casUniqueAsStr = casUniqueMaybe match { 
  case Some(_) => casUniqueMaybe.get.toString
  case None => ""
 }
```

dkarivalis: interesting! personally i prefer the condensed version in the case 🐱

vrrs: The fact that we are talking about this suggests that the condensed version is not the one you should be using.

dkarivalis: ah okay, i did not know you felt so strongly about this! will update ^_^

benblack86: Alternative: `anIntMaybe.map(_.toString).getOrElse("")`

dxuhuang: Or: `antIntMaybe.fold("") { _.toString }`

PR #681, opened 2019-06-22
Adding Scala 2.13 support

plokhotnyuk:


simpadjo:
For those who want to run colossus on scala-2.13, I finalized PR by @plokhotnyuk in my repo:
https://github.com/simpadjo/colossus .
I also published the packages (snapshots for now) but apparently they are not usable from build tools w/o a token due to github quirks.

